PR #12, opened 2018-11-23
Monitor state of a running container to cleanly shutdown

IRCody:
Monitor the state of a running container to enable clean shutdown of the
container and vm.

*Description of changes:*

This adds a continuous monitor goroutine that calls agent.State() once per second to see when the container stops and initiate shutdown and publish the taskExit event. The goal is to enable `ctr run` to successfully run the container and exit. To that goal a few other minor changes are here:
* remove a .init.pid file if it exists -- runc leaves this behind
* add a call to runc.Cleanup() before we shutdown the agent
* add empty string values for container stdio
* made calls to runtime.Kill() to also stop the vm

Doing it this way is a little hacky but is (I think) the simplest approach.  

### DIFF
--- a/agent/service.go
+++ b/agent/service.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"io/ioutil"
 	"path/filepath"
+	"os"
 
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/namespaces"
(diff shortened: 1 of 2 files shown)


IRCody:
I did not close this. Suspect it had something to do with it being set from private -> public.

PR #21, opened 2018-11-27
Monitor state of a running container to cleanly shutdown

IRCody:
It wasn't giving me the option to re-open #12 so I am just resubmitting it.  Please refer to #12 for discussion history. I copied over the two outstanding comments.

Monitor the state of a running container to enable clean shutdown of the
container and vm. Does this by adding a continuous monitor
goroutine that calls agent.State() once per second to see when the
container stops and initiate shutdown and publish the taskExit event.

### DIFF
--- a/agent/service.go
+++ b/agent/service.go
@@ -16,6 +16,7 @@ package main
 import (
 	"context"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 
 	"github.com/containerd/containerd/log"
@@ -47,23 +48,30 @@ func NewTaskService(runc shim.Shim) shim.Shim {
 func (ts *TaskService) Create(ctx context.Context, req *shimapi.CreateTaskRequest) (*shimapi.CreateTaskResponse, error) {
 	log.G(ctx).WithFields(logrus.Fields{"id": req.ID, "bundle": req.Bundle}).Info("create")
 
-	// Use mount path instead of bundle path inside the VM
-	req.Bundle = bundleMountPath
-
-	// Do not pass any mounts to runc, everything is already mounted for us
-	req.Rootfs = nil
-
 	// Passthrough runcOptions
 	opts, err := unpackBundle(filepath.Join(bundleMountPath, "config.json"), req.Options)
 	if err != nil {
 		return nil, err
 	}
 	req.Options = opts
+	// Use mount path instead of bundle path inside the VM
+	req.Bundle = bundleMountPath
+
+	// Do not pass any mounts to runc, everything is already mounted for us
+	req.Rootfs = nil
+	// TODO: handle stdio see: https://github.com/firecracker-microvm/firecracker-containerd/issues/17
+	req.Stdin = ""
+	req.Stdout = ""
+	req.Stderr = ""
 
 	ctx = namespaces.WithNamespace(ctx, defaultNamespace)
+	// before create call ensure we remove any existing .init.pid file
+	// We can ignore errors since it's valid for the file to not be present
+	os.Remove(".init.pid")
 	resp, err := ts.runc.Create(ctx, req)
+
 	if err != nil {
-		log.G(ctx).WithError(err).Error("create failed")
+		log.G(ctx).WithError(err).Error("error creating container")
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


agent/service.go
```diff
@@ -327,7 +335,7 @@ func (ts *TaskService) Connect(ctx context.Context, req *shimapi.ConnectRequest)
 
 func (ts *TaskService) Shutdown(ctx context.Context, req *shimapi.ShutdownRequest) (*types.Empty, error) {
 	log.G(ctx).WithFields(logrus.Fields{"id": req.ID, "now": req.Now}).Debug("shutdown")
-
+	ts.runc.Cleanup(ctx)
```
IRCody: Copied from closed PR from @mxpv:
>runc has NamespaceRequired check. It makes sense to move this line below namespaces.WithNamespace(ctx, defaultNamespace). Also would be nice to log warning if cleanup fails.

runtime/service.go
```diff
@@ -320,10 +366,7 @@ func (s *service) Connect(ctx context.Context, req *taskAPI.ConnectRequest) (*ta
 func (s *service) Shutdown(ctx context.Context, req *taskAPI.ShutdownRequest) (*ptypes.Empty, error) {
 	log.G(ctx).WithFields(logrus.Fields{"id": req.ID, "now": req.Now}).Debug("shutdown")
 
-	if _, err := s.agentClient.Shutdown(ctx, req); err != nil {
-		log.G(ctx).WithError(err).Error("failed to shutdown agent")
-	}
-
+	s.agentClient.Shutdown(ctx, req)
```
IRCody: copied from closed pr [comment](https://github.com/firecracker-microvm/firecracker-containerd/pull/12#discussion_r236523314) from @mxpv: 
>Just curious: is it by mistake or logging was removed for some reason?

IRCody: Shutdown was not returning successfully. When looking at why it wasbecause the runc impl always returns an error because it just calls os.Exit() prior to returning [here](https://github.com/containerd/containerd/blob/master/runtime/v2/runc/service.go#L585).

Looking at it now though I think it makes more sense to ignore that result in the agent since the agent is doing more than just that runc call during shutdown. Will update it to have this logging back and ignore the always-error from calling runc.Shutdown inside the agent instead.

agent/service.go
```diff
@@ -47,23 +48,30 @@ func NewTaskService(runc shim.Shim) shim.Shim {
...
 	if err != nil {
 		return nil, err
 	}
 	req.Options = opts
+	// Use mount path instead of bundle path inside the VM
+	req.Bundle = bundleMountPath
+
+	// Do not pass any mounts to runc, everything is already mounted for us
+	req.Rootfs = nil
+	// TODO: handle stdio
```
nmeyerhans: We've got a bunch of these TODO comments in our code. Can we start including links to associated GitHub issues along side them? We lose too much state otherwise.

runtime/service.go
```diff
@@ -192,10 +194,47 @@ func (s *service) Start(ctx context.Context, req *taskAPI.StartRequest) (*taskAP
...
+				ExecID: exec_id,
+			}
+			resp, err := s.agentClient.State(ctx, req)
+			if err != nil {
+				log.G(ctx).WithError(err).Error("error monitoring state")
+				continue
+			}
+			// if ending state, stop vm and break
+			if resp.Status == task.StatusStopped {
+				s.publish.Publish(ctx, runtime.TaskExitEventTopic, &eventstypes.TaskExit{
```
nmeyerhans: I wonder if this is actually the right approach to take here. Polling isn't unreasonable, but shouldn't the even originate from within the VM? Furthermore, shouldn't the VM shut itself down when the task exits?

runtime/service.go
```diff
@@ -265,11 +304,18 @@ func (s *service) Resume(ctx context.Context, req *taskAPI.ResumeRequest) (*ptyp
 // Kill a process with the provided signal
 func (s *service) Kill(ctx context.Context, req *taskAPI.KillRequest) (*ptypes.Empty, error) {
 	log.G(ctx).WithFields(logrus.Fields{"id": req.ID, "exec_id": req.ExecID}).Debug("kill")
+	// right now we want to kill vm always when kill is called
+	// may not be true in multi-container vm
+	defer func() {
+		log.G(ctx).Debug("Stopping VM during kill")
+		if err := s.stopVM(); err != nil {
```
nmeyerhans: `stopVM()` simply terminates the firecracker process. We need to take care to shut down more cleanly, or else we risk corrupting the VMs filesystem and/or losing buffered writes.

We probably want to run `/sbin/reboot` inside the VM during agent's handling of `Shutdown` and `Kill`, then wait a short period of time before terminating the VM more forcefully (if it still exists).

agent/service.go
IRCody: We could I guess. I don't think all TODO comments warrant a github issue. 

I added a link to the relevant issue for this TODO though.

runtime/service.go
IRCody: I think this makes sense.  I think there are some other shutdown related issues that should be dealt with together. I created #25 to track that. Do you feel strongly about it being done in this PR?

agent/service.go
nmeyerhans: Thanks. Maybe they don't always need it, but in general, if it's something that we know needs to be done, we should document it. Issues are the right place for that.

runtime/service.go
nmeyerhans: OK

runtime/service.go
nmeyerhans: OK.

PR #23, opened 2018-11-27
Retry vsock dials

mxpv:
*Description of changes:*
This PR adds vsock dial retries.

### DIFF
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -16,6 +16,7 @@ package main
 import (
 	"context"
 	"io/ioutil"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


runtime/service.go
```diff
@@ -415,6 +416,33 @@ func newServer() (*ttrpc.Server, error) {
...
+
+	var lastErr error
+	var currentDelay = initialDelay
+	for i := 1; i <= retryCount; i++ {
+		conn, err := vsock.Dial(contextID, port)
+		if err == nil {
+			return conn, nil
+		}
+
+		log.G(ctx).WithError(err).Warnf("vsock dial failed (attempt %d of %d), will retry in %s", i, retryCount, currentDelay)
```
nmeyerhans: I think this warning will end up firing more than we'd like to see in normal operations. Let's leave it in place for now, but we will probably want to change the severity and/or wording once we've seen this in operation a while. My hunch is that the 400ms attempt is the one that'll hit most frequently.

PR #26, opened 2018-11-29
Add Travis and linter files

mxpv:
*Description of changes:*
Add travis and linter configuration.
Travis will build all plugins and run bunch of linters.

~Note 1: Don't merge this please, want to make sure it works as expected with Travis.~
Note 2: There is no tests/coverage stuff, out of scope of this PR.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,21 @@
+linters:
+  disable-all: true
+  enable:
+    - govet
+    - staticcheck
+    - unused
+    - gosimple
+    - structcheck
+    - varcheck
+    - ineffassign
+    - deadcode
+    - typecheck
+    - golint
+    - gosec
+    - unconvert
+    - goconst
+    - goimports
+    - misspell
+    - prealloc
+    - scopelint
+    - gocritic
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: go
+
+go:
+  - 1.11.x
+
+env:
+  - GO111MODULE=on
+
+install:
+  - make deps
+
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -0,0 +1,14 @@
+language: go
+
+go:
+  - 1.11.x
+
+env:
+  - GO111MODULE=on
+
+install:
+  - curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $GOPATH/bin v1.12.3
```
samuelkarp: It'd be good to move this in the Makefile, like `make get-deps` or something like that.

.travis.yml
```diff
@@ -0,0 +1,14 @@
...
+
+script:
+  - make
+  - golangci-lint run
```
samuelkarp: It'd be good to put this in the Makefile so you can just run `make lint`.

.golangci.yml
```diff
@@ -0,0 +1,22 @@
+linters:
+  disable-all: true
+  enable:
+    - govet
+    - errcheck
```
samuelkarp: I'm not sure about this one.  I think I'd rather have it off than have the extra `_ =` assignments and `//nolint:errcheck` comments.

.travis.yml
mxpv: Done

.travis.yml
mxpv: Done

.golangci.yml
mxpv: Done

PR #27, opened 2018-11-29
Add stdio proxying between container, agent, runtime

### LINKED ISSUE #17: Proxy stdio streams
A way to take the containers stdio streams and proxy those to the fifo's that containerd provides outside the vm over vsock.

IRCody:
Add stdio proxying from the container fifos to the agent, from the agent over
vsock to the runtime, and from the runtime to the fifo's specified by
containerd.

*Issue #, if available:*
closes #17 
*Description of changes:*
Proxies the stdout of the container over vsock to the runtime where it is re-written to the fifo's that containerd expects the output to be written.

Tested by running alpine /bin/sh and was able to use it interactively.
```
% sudo ./ctr run --runtime aws.firecracker --snapshotter firecracker-snapshotter docker.io/library/alpine:latest 2123 /bin/sh
ps
PID   USER     TIME  COMMAND
    1 root      0:00 /bin/sh
    6 root      0:00 ps
ls -al
total 16
drwxr-xr-x   18 root     root          1024 Nov 29 00:31 .
drwxr-xr-x   18 root     root          1024 Nov 29 00:31 ..
drwxr-xr-x    2 root     root          3072 Nov 29 00:31 bin
drwxr-xr-x    5 root     root           320 Nov 29 00:32 dev
drwxr-xr-x   15 root     root          1024 Nov 29 00:31 etc
drwxr-xr-x    2 root     root          1024 Sep 11 20:23 home
drwxr-xr-x    5 root     root          1024 Nov 29 00:31 lib
drwxr-xr-x    5 root     root          1024 Nov 29 00:31 media
drwxr-xr-x    2 root     root          1024 Sep 11 20:23 mnt
dr-xr-xr-x   64 root     root             0 Nov 29 00:32 proc
drwx------    2 root     root          1024 Sep 11 20:23 root
drwxr-xr-x    2 root     root            40 Nov 29 00:32 run
drwxr-xr-x    2 root     root          1024 Nov 29 00:31 sbin
drwxr-xr-x    2 root     root          1024 Sep 11 20:23 srv
dr-xr-xr-x   11 root     root             0 Nov 29 00:32 sys
drwxrwxrwt    2 root     root          1024 Sep 11 20:23 tmp
drwxr-xr-x    7 root     root          1024 Nov 29 00:31 usr
drwxr-xr-x   11 root     root          1024 Nov 29 00:31 var
```
There is still an error related to how we shut down the container/agent/runtime: 
`ctr: rpc error: code = Unknown desc = ttrpc: client shutting down: read vsock:host(2):1222: transport endpoint is not connected: unknown`

### DIFF
--- a/agent/service.go
+++ b/agent/service.go
@@ -15,23 +15,31 @@ package main
 
 import (
 	"context"
+	"io"
 	"io/ioutil"
+	"net"
 	"os"
 	"path/filepath"
+	"syscall"
 
+	"github.com/containerd/containerd/cio"
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/namespaces"
 	"github.com/containerd/containerd/runtime/v2/shim"
 	shimapi "github.com/containerd/containerd/runtime/v2/task"
+	"github.com/containerd/fifo"
 	"github.com/gogo/protobuf/types"
+	"github.com/mdlayher/vsock"
 	"github.com/sirupsen/logrus"
 
+	"github.com/firecracker-microvm/firecracker-containerd/internal"
 	"github.com/firecracker-microvm/firecracker-containerd/proto"
 )
 
 const (
 	defaultNamespace = "default"
 	bundleMountPath  = "/container"
+	defaultStdioPath = "/container/fifo"
 )
 
 // TaskService represents inner shim wrapper over runc in order to:
@@ -39,14 +47,15 @@ const (
 // - Add debug logging to simplify debugging
 // - Make place for future extensions as needed
 type TaskService struct {
-	runc   shim.Shim
-	cancel context.CancelFunc
+	runc    shim.Shim
+	cancels []context.CancelFunc
+	io      *cio.FIFOSet
 }
 
 func NewTaskService(runc shim.Shim, cancel context.CancelFunc) shimapi.TaskService {
 	return &TaskService{
-		runc:   runc,
-		cancel: cancel,
+		runc:    runc,
+		cancels: []context.CancelFunc{cancel},
 	}
 }
 
@@ -64,15 +73,23 @@ func (ts *TaskService) Create(ctx context.Context, req *shimapi.CreateTaskReques
 
 	// Do not pass any mounts to runc, everything is already mounted for us
 	req.Rootfs = nil
-	// TODO: handle stdio see: https://github.com/firecracker-microvm/firecracker-containerd/issues/17
-	req.Stdin = ""
-	req.Stdout = ""
-	req.Stderr = ""
-
+	// handle STDIO
+	ts.io, err = cio.NewFIFOSetInDir(defaultStdioPath, req.ID, req.Terminal)
+	if err != nil {
+		log.G(ctx).WithError(err).Error("error proxying io")
+		return nil, err
+	}
+	req.Stdin = ts.io.Stdin
+	req.Stderr = ts.io.Stderr
+	req.Stdout = ts.io.Stdout
+	ioctx, cancel := context.WithCancel(ctx)
+	ts.cancels = append(ts.cancels, cancel)
+	ts.proxyStdio(ioctx, req.Stdin, req.Stdout, req.Stderr)
 	ctx = namespaces.WithNamespace(ctx, defaultNamespace)
 	// before create call ensure we remove any existing .init.pid file
 	// We can ignore errors since it's valid for the file to not be present
 	os.Remove(".init.pid")
+	log.G(ctx).Debug("calling runc create")
 	resp, err := ts.runc.Create(ctx, req)
 
 	if err != nil {
@@ -82,6 +99,58 @@ func (ts *TaskService) Create(ctx context.Context, req *shimapi.CreateTaskReques
 
 	log.G(ctx).WithField("pid", resp.Pid).Debugf("create succeeded")
 	return resp, nil
+}
+
+func (ts *TaskService) proxyStdio(ctx context.Context, stdin, stdout, stderr string) {
+	go proxyIO(ctx, stdin, internal.StdinPort, true)
+	go proxyIO(ctx, stdout, internal.StdoutPort, false)
+	go proxyIO(ctx, stderr, internal.StderrPort, false)
+}
+
+func proxyIO(ctx context.Context, path string, port uint32, in bool) {
+	if path == "" {
+		return
+	}
+	log.G(ctx).Debug("setting up IO for " + path)
+	f, err := fifo.OpenFifo(ctx, path, syscall.O_RDWR|syscall.O_NONBLOCK|syscall.O_CREAT, 0700)
+	if err != nil {
+		log.G(ctx).WithError(err).Error("error opening fifo")
+		return
+	}
+	listener, err := vsock.Listen(port)
+	if err != nil {
+		log.G(ctx).WithError(err).Error("unable to listen on vsock")
+		f.Close()
+		return
+	}
+
+	var conn net.Conn
+	for {
+		// accept is non-blocking so try to accept until we get
+		// a connection
+		// TODO: investigate if there is a way to distinguish
+		// transient errors from permanent ones.
+		conn, err = listener.Accept()
+		if err != nil {
+			continue
+		}
+		break
+	}
+	go func() {
+		<-ctx.Done()
+		conn.Close()
+		f.Close()
+	}()
+	log.G(ctx).Debug("begin copying io")
+	buf := make([]byte, internal.DefaultBufferSize)
+	if in {
+		_, err = io.CopyBuffer(f, conn, buf)
+	} else {
+		_, err = io.CopyBuffer(conn, f, buf)
+	}
+	if err != nil {
+		log.G(ctx).WithError(err).Error("error with stdio")
+	}
 }
 
 func unpackBundle(path string, bundle *types.Any) (*types.Any, error) {
@@ -331,9 +400,17 @@ func (ts *TaskService) Shutdown(ctx context.Context, req *shimapi.ShutdownReques
 	}
 
 	// We don't want to call runc.Shutdown here as it just os.Exits behind.
-	// Invoking cancel here for graceful shutdown instead and call runc Shutdown at end.
-	ts.cancel()
+	// calling all cancels here for graceful shutdown instead and call runc Shutdown at end.
+	ts.cancelAll()
 
 	log.G(ctx).Debug("going to gracefully shutdown agent")
 	return &types.Empty{}, nil
+}
+
+func (ts *TaskService) cancelAll() {
+	// cancel LIFO order
+	for i := len(ts.cancels); i >= 0; i-- {
+		log.G(context.Background()).Debug("Cancelling ", i)
+		ts.cancels[i]()
+	}
 }
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/containerd/console v0.0.0-20181022165439-0650fd9eeb50 // indirect
 	github.com/containerd/containerd v1.2.0
 	github.com/containerd/continuity v0.0.0-20181027224239-bea7585dbfac
-	github.com/containerd/fifo v0.0.0-20180307165137-3d5202aec260 // indirect
+	github.com/containerd/fifo v0.0.0-20180307165137-3d5202aec260
 	github.com/containerd/go-runc v0.0.0-20180907222934-5a6d9f37cfa3 // indirect
 	github.com/containerd/ttrpc v0.0.0-20181001154009-f51df4475b76
 	github.com/containerd/typeurl v0.0.0-20181015155603-461401dc8f19 // indirect
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"context"
+	"io"
 	"io/ioutil"
 	"net"
 	"os"
@@ -34,6 +35,7 @@ import (
 	"github.com/containerd/containerd/runtime"
 	"github.com/containerd/containerd/runtime/v2/shim"
 	taskAPI "github.com/containerd/containerd/runtime/v2/task"
+	"github.com/containerd/fifo"
 	"github.com/containerd/ttrpc"
 	"github.com/firecracker-microvm/firecracker-containerd/internal"
 	"github.com/firecracker-microvm/firecracker-containerd/proto"
@@ -61,6 +63,8 @@ type service struct {
 	agentClient  taskAPI.TaskService
 	config       *Config
 	machine      *firecracker.Machine
+	ctx          context.Context
+	cancel       context.CancelFunc
 }
 
 var _ = (taskAPI.TaskService)(&service{})
@@ -185,6 +189,8 @@ func (s *service) Create(ctx context.Context, request *taskAPI.CreateTaskRequest
 		log.G(ctx).WithError(err).Error("create failed")
 		return nil, err
 	}
+	s.ctx, s.cancel = context.WithCancel(ctx)
+	go s.proxyStdio(s.ctx, request.Stdin, request.Stdout, request.Stderr, defaultCID)
 	log.G(ctx).Infof("successfully created task with pid %d", resp.Pid)
 	return resp, nil
 }
@@ -196,7 +202,7 @@ func (s *service) Start(ctx context.Context, req *taskAPI.StartRequest) (*taskAP
 		return nil, err
 	}
 	// TODO: Do we need to cancel this at some point?
-	go s.monitorState(ctx, req.ID, req.ExecID, resp.Pid)
+	go s.monitorState(s.ctx, req.ID, req.ExecID, resp.Pid)
 
 	return resp, nil
 }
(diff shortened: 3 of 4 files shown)


agent/service.go
```diff
@@ -79,6 +112,73 @@ func (ts *TaskService) Create(ctx context.Context, req *shimapi.CreateTaskReques
...
+		log.G(ctx).Debug("Accepted connection")
+		if err != nil {
+			log.G(ctx).WithError(err).Error("unable to accept vsock connection")
+			return
+		}
+
+		buf := make([]byte, internal.DefaultBufferSize)
+		for {
+			read, err := reader.Read(buf)
+			_, werr := conn.Write(buf[:read])
```
samuelkarp: It looks like you're ignoring both `err` and `werr`, so the loop here is infinite.  Do we want the loop to be infinite or is there a reasonable exit condition?

agent/service.go
```diff
@@ -79,6 +112,73 @@ func (ts *TaskService) Create(ctx context.Context, req *shimapi.CreateTaskReques
...
+		log.G(ctx).Debug("Accepted connection")
+		if err != nil {
+			log.G(ctx).WithError(err).Error("unable to accept vsock connection")
+			return
+		}
+
+		buf := make([]byte, internal.DefaultBufferSize)
+		for {
+			n, err := conn.Read(buf)
+			_, werr := writer.Write(buf[:n])
```
samuelkarp: Same comment as above about infinite loops

runtime/service.go
```diff
@@ -236,6 +240,72 @@ func (s *service) monitorState(ctx context.Context, id, exec_id string, pid uint
...
+	if path != "" {
+		reader, err := fifo.OpenFifo(ctx, path, syscall.O_RDONLY|syscall.O_NONBLOCK, 0700)
+		if err != nil {
+			log.G(ctx).WithError(err).Error("error opening fifo")
+			return
+		}
+		conn, err := vsock.Dial(CID, port)
+		if err != nil {
+			log.G(ctx).WithError(err).Error("unable to dial agent vsock")
+		}
```
mxpv: Missing return in case of error?

runtime/service.go
```diff
@@ -236,6 +240,72 @@ func (s *service) monitorState(ctx context.Context, id, exec_id string, pid uint
...
+		buf := make([]byte, internal.DefaultBufferSize)
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				n, err := reader.Read(buf)
+				n, werr := conn.Write(buf[:n])
```
mxpv: Why not to use `io.CopyBuffer(conn, reader, buf) / CopyN / LimitReader`  instead?
This will simplify error handling as well.

runtime/service.go
```diff
@@ -236,6 +240,72 @@ func (s *service) monitorState(ctx context.Context, id, exec_id string, pid uint
...
+				}
+				if werr != nil {
+					log.G(ctx).WithError(err).Error("error writing to vsock")
+				}
+			}
+		}
+	}
+}
+
+func proxyOut(ctx context.Context, path string, CID, port uint32) {
```
mxpv: `proxyIn` and `proxyOut` are almost identical except a few lines of code? May be have just one function with `direction` parameter?

runtime/service.go
```diff
@@ -317,6 +387,7 @@ func (s *service) Kill(ctx context.Context, req *taskAPI.KillRequest) (*ptypes.E
 	if err != nil {
 		return nil, err
 	}
+	s.cancel()
```
mxpv: I think we need proper resource cleanup on both sides:
- Wait for goroutines to finish reading/writing
- Close vsock connections if no longer needed
- Close fifos

agent/service.go
```diff
@@ -59,15 +67,40 @@ func (ts *TaskService) Create(ctx context.Context, req *shimapi.CreateTaskReques
...
+		if stdout, err = fifo.OpenFifo(ctx, ts.io.Stdout, syscall.O_RDONLY|syscall.O_CREAT|syscall.O_NONBLOCK, 0700); err != nil {
+			return nil, err
+		}
+	}
+	var stderr io.ReadWriteCloser
+	if ts.io.Stderr != "" {
+		if stderr, err = fifo.OpenFifo(ctx, ts.io.Stderr, syscall.O_RDONLY|syscall.O_CREAT|syscall.O_NONBLOCK, 0700); err != nil {
+			return nil, err
+		}
+	}
```
mxpv: - Move this to `createFifos() (*cio.FIFOSet, error)` ?
- Also can we add some context to err returns, so we can distinguish which one exactly failed?

agent/service.go
```diff
@@ -59,15 +67,40 @@ func (ts *TaskService) Create(ctx context.Context, req *shimapi.CreateTaskReques
...
 
+	ts.proxyStdio(ctx, stdin, stdout, stderr)
 	ctx = namespaces.WithNamespace(ctx, defaultNamespace)
 	// before create call ensure we remove any existing .init.pid file
 	// We can ignore errors since it's valid for the file to not be present
 	os.Remove(".init.pid")
+	log.G(ctx).Debug("Calling runc create")
```
mxpv: nit: lowercase everywhere?

agent/service.go
IRCody: Right now the exit conditions is "the container exits, causing the process to exit". There could probably be a better way to signal that we are good to exit though and wind this down gracefully.

It might make sense to wait on these for a tiny bit of time before calling os.Exit() in shutdown?

runtime/service.go
IRCody: idk if I like the direction parameter but I bet there is a way to extract the shared code part to combine them.

go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/godbus/dbus v0.0.0-20181025153459-66d97aec3384 // indirect
 	github.com/gogo/protobuf v1.1.1
 	github.com/google/go-cmp v0.2.0 // indirect
-	github.com/mdlayher/vsock v0.0.0-20181130155850-676f733b747c
+	github.com/mdlayher/vsock v0.0.0-20180806202057-2b18bec03226
```
IRCody: Updating this dependency broke this PR, so I reverted it temporarily until we can figure out why. 

One time I got this in the agent logs: 
```
ime="2018-12-05T07:23:22Z" level=info msg="creating runc shim" id=
time="2018-12-05T07:23:22Z" level=info msg="listening to vsock" port=10789
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 779.41µs" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 153.551µs" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 2.145821ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 3.010051ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 15.113937ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 5.16732ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 16.969758ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 83.776148ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=info msg=create bundle=/run/containerd/io.containerd.runtime.v2.task/default/2151 id=2151
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 117.216µs" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 279.449µs" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=debug msg="Created listener"
time="2018-12-05T07:23:22Z" level=debug msg="Created listener"
time="2018-12-05T07:23:22Z" level=debug msg="Created listener"
time="2018-12-05T07:23:22Z" level=debug msg="Calling runc create"
time="2018-12-05T07:23:22Z" level=debug msg="Accepted connection"
time="2018-12-05T07:23:22Z" level=error msg="unable to accept vsock connection" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=debug msg="Accepted connection"
time="2018-12-05T07:23:22Z" level=error msg="unable to accept vsock connection" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=debug msg="Accepted connection"
time="2018-12-05T07:23:22Z" level=error msg="unable to accept vsock connection" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 398.084µs" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 7.669287ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 4.491574ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 22.458836ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 43.445515ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=debug msg="create succeeded" pid=206
time="2018-12-05T07:23:22Z" level=debug msg=state exec_id= id=2151
time="2018-12-05T07:23:22Z" level=debug msg="state succeeded" bundle=/container id=2151 pid=206 status=CREATED
time="2018-12-05T07:23:22Z" level=debug msg=start exec_id= id=2151
time="2018-12-05T07:23:22Z" level=debug msg=wait exec_id= id=2151
time="2018-12-05T07:23:22Z" level=debug msg="start succeeded" pid=206
time="2018-12-05T07:23:22Z" level=debug msg=state exec_id= id=2151
time="2018-12-05T07:23:22Z" level=debug msg="state succeeded" bundle=/container id=2151 pid=206 status=RUNNING
time="2018-12-05T07:23:22Z" level=debug msg=resize_pty exec_id= id=2151
time="2018-12-05T07:23:22Z" level=debug msg="resize_pty succeeded"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 62.182873ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 80.644968ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 478.584091ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:22Z" level=error msg="ttrpc: failed accept; backoff 539.11079ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:23Z" level=debug msg=state exec_id= id=2151
time="2018-12-05T07:23:23Z" level=debug msg="state succeeded" bundle=/container id=2151 pid=206 status=RUNNING
time="2018-12-05T07:23:23Z" level=error msg="ttrpc: failed accept; backoff 853.353331ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:24Z" level=debug msg=state exec_id= id=2151
time="2018-12-05T07:23:24Z" level=debug msg="state succeeded" bundle=/container id=2151 pid=206 status=RUNNING
time="2018-12-05T07:23:24Z" level=error msg="ttrpc: failed accept; backoff 324.778273ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:24Z" level=error msg="ttrpc: failed accept; backoff 138.149956ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:24Z" level=error msg="ttrpc: failed accept; backoff 693.774911ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:25Z" level=debug msg=state exec_id= id=2151
time="2018-12-05T07:23:25Z" level=debug msg="state succeeded" bundle=/container id=2151 pid=206 status=RUNNING
time="2018-12-05T07:23:25Z" level=error msg="ttrpc: failed accept; backoff 183.515637ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:25Z" level=error msg="ttrpc: failed accept; backoff 821.866378ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:26Z" level=debug msg=state exec_id= id=2151
time="2018-12-05T07:23:26Z" level=debug msg="state succeeded" bundle=/container id=2151 pid=206 status=RUNNING
time="2018-12-05T07:23:26Z" level=error msg="ttrpc: failed accept; backoff 342.231109ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:26Z" level=error msg="ttrpc: failed accept; backoff 797.652072ms" error="resource temporarily unavailable"
time="2018-12-05T07:23:27Z" level=debug msg=state exec_id= id=2151
...
```
but all other times it fails at startup with:
`
ctr: ttrpc: client shutting down: read vsock:host(2):1415: transport endpoint is not connected: unknown
`

The error seems to be coming from ttrpc [here](https://github.com/containerd/ttrpc/blob/2a805f71863501300ae1976d29f0454ae003e85a/server.go#L106) cause by errors from calling Accep (resource temporarily unavailable). I think the only change between the two revisions is this PR https://github.com/mdlayher/vsock/pull/20 . I'm not sure exactly how that change causes these errors. It'd be interesting if someone else could verify the behavior in case it's something local to my box. 


The steps to test the updated dependency:
1. modify go.mod to add `github.com/mdlayher/vsock v0.0.0-20181130155850-676f733b747c` for the vsock version instead of `github.com/mdlayher/vsock v0.0.0-20180806202057-2b18bec03226`
1. `go mod tidy`
1. make clean all
1. copy over runtime to path and agent to the image
1. ctr run an image like : `% sudo ./ctr run --runtime aws.firecracker --snapshotter firecracker-snapshotter --tty docker.io/library/alpine:latest 2165 /bin/sh`. I tried both with/without --tty and it didn't seem to make a difference.

mxpv: I'll take a closer look on this, feel free to revert this commit if it blocks your PR.

IRCody: I'd rather figure out the core issue than revert if possible.

nmeyerhans: I can confirm this problem when building with `github.com/mdlayher/vsock v0.0.0-20180806202057-2b18bec03226`

The problem is not reproducible with the previous version.

mxpv: @nmeyerhans can you clarify? you see error messages or it just doesn't work?
This messages are caused by non-blocking behavior I recently [added](https://github.com/mdlayher/vsock/commit/fda437ecbf04836c30bfb8c397ca7383c5d01a7e) to `vsock` package, so `Accept` no longer blocks, but just returns temporary error https://github.com/containerd/ttrpc/blob/master/server.go#L95
Except these annoying error messages, `agent` should work as expected.

agent/service.go
```diff
@@ -84,6 +98,50 @@ func (ts *TaskService) Create(ctx context.Context, req *shimapi.CreateTaskReques
...
+		conn, err := listener.Accept()
+		if err != nil {
+			listener.Close()
+			f.Close()
+			log.G(ctx).WithError(err).Error("error accepting vsock connection")
+		}
+		go func() {
+			<-ctx.Done()
+			conn.Close()
+			f.Close()
```
mxpv: How do you cancel this `ctx` ?

IRCody: It's not cancelled atm. I'll update it to cancel before it calls TaskService.Cancel() in Shutdown. Does that seem like the right place?

IRCody: This has already been rebased on top of the commit and has that cancel. I don't think it helps though since that cancel is passed in and not generated/controlled by the TaskService itself.

Would it make sense to, instead of having NewTaskService() take a cancelFunc, to return one to the caller so that the TaskService maintains control of what the cancel does?

mxpv: We need same `context` to be passed to `server.Serve(ctx, listener)` and `server.Shutdown(ctx)` in order to stop properly everything.

How about something like this:
```go
taskService, shutdownSignal := NewTaskService(runcTaskService) <- no cancel here

group.Go(func() error {
     return server.Serve(ctx, listener)
})

...

group.Go(func() error {
     <- shutdownSignal
     cancel() <--  call cancel here instead of calling it from `Shutdown`
})
```

agent/service.go
```diff
@@ -84,6 +98,50 @@ func (ts *TaskService) Create(ctx context.Context, req *shimapi.CreateTaskReques
...
 }
 
+func (ts *TaskService) proxyStdio(ctx context.Context, stdin, stdout, stderr string) {
+	go proxyIO(ctx, stdin, internal.StdinPort, true)
+	go proxyIO(ctx, stdout, internal.StdoutPort, false)
+	go proxyIO(ctx, stderr, internal.StderrPort, false)
+}
+
+func proxyIO(ctx context.Context, path string, port uint32, in bool) {
+	if path != "" {
```
samuelkarp: Can you reverse the conditional and out-dent the main body?

```go
func proxyIO(ctx context.Context, path string, port uint32, in bool) {
	if path == "" {
		return
	}
	...
}
```

runtime/service.go
```diff
@@ -236,6 +241,44 @@ func (s *service) monitorState(ctx context.Context, id, exec_id string, pid uint
...
 }
 
+func (s *service) proxyStdio(ctx context.Context, stdin, stdout, stderr string, CID uint32) {
+	go proxyIO(ctx, stdin, CID, internal.StdinPort, true)
+	go proxyIO(ctx, stdout, CID, internal.StdoutPort, false)
+	go proxyIO(ctx, stderr, CID, internal.StderrPort, false)
+}
+
+func proxyIO(ctx context.Context, path string, CID, port uint32, in bool) {
+	if path != "" {
```
samuelkarp: Can you reverse the conditional and out-dent the main body?

```go
func proxyIO(ctx context.Context, path string, port uint32, in bool) {
	if path == "" {
		return
	}
	...
}
```

agent/service.go
IRCody: I did something a little different by holding a slice of []context.CancelFuncs and calling them all during shutdown, adding the one that is passed in by NewTaskService() to the list. Does that seem like it will work?

agent/service.go
```diff
@@ -84,6 +101,58 @@ func (ts *TaskService) Create(ctx context.Context, req *shimapi.CreateTaskReques
...
+	if path == "" {
+		return
+	}
+	log.G(ctx).Debug("setting up IO for " + path)
+	f, err := fifo.OpenFifo(ctx, path, syscall.O_RDWR|syscall.O_NONBLOCK|syscall.O_CREAT, 0700)
+	if err != nil {
+		log.G(ctx).WithError(err).Error("error opening fifo")
+		return
+	}
+	listener, err := vsock.Listen(port)
```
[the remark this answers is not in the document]
IRCody: I think this makes sense. We also do need to add tests. I will try to take a pass at both on the next revision.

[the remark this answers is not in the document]
nmeyerhans: I dislike mocks with great passion. I dislike restructuring code to facilitate mocks even more.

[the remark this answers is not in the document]
xibz: @nmeyerhans - I disagree. We need to have tests, and we need some way of doing that and right now we have to mock since there isn't any other options without doing a full blown functional test. The benefit of this, also, is it also allows for extendability.

We shouldn't use the excuse of not liking something to avoid writing tests.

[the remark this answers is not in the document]
mxpv: I don't think we need separate bridge here. vsock already implements `net.Listener` / `net.Conn`, they are already generic interfaces for both vsock and net tcp/udp things.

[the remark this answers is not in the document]
xibz: Yea, except you can't swap it in the `net.Listener` easily. That's the point of the bridge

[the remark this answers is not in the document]
mxpv: Can be same trick as with ioctl ?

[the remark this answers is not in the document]
xibz: @mxpv - Talked with Cody, and he stated he could refactor things later to allow for the `net.Listener` to be passed in which would solve the issue. So, I am okay with adding the test later when we refactor this code.

PR #29, opened 2018-11-29
Makefile improvements

nmeyerhans:
Add dependencies for the non-PHONY targets

Reduce some redundancy

Restructure a little bit for readability

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,26 +11,17 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-all: agent runtime snapshotter
+SUBDIRS:=agent runtime snapshotter
 
-.PHONY: agent
-agent:
-	$(MAKE) -C agent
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -11,26 +11,14 @@
...
+$(SUBDIRS):
+	$(MAKE) -C $@
 
-.PHONY: snapshotter
-snapshotter:
-	$(MAKE) -C snapshotter
-
-.PHONY: proto
-proto:
-	proto/generate.sh
```
mxpv: are we throwing away proto stuff? I was thinking about moving it to `Makefile` from `generate.sh`

xibz: +1 on this. Would rather have this in the make file rather than a `generate.sh` file.

nmeyerhans: woops, that was a mistake! proto should still be there

PR #31, opened 2018-11-30
Refactor snapshotter entry point

mxpv:

*Description of changes:*
This refactors snapshotter to support more than one implementation and renames current implementation to naive.

### DIFF
--- a/snapshotter/Makefile
+++ b/snapshotter/Makefile
@@ -11,12 +11,12 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-all: snapshotter
+all: naive
 
-snapshotter: *.go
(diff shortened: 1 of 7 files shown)


snapshotter/Makefile
```diff
@@ -11,12 +11,12 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-all: snapshotter
+all: naive
 
-snapshotter: *.go
-	go build -o snapshotter ./cmd/snapshotter
+naive: *.go
+	go build -o naive_snapshotter ./cmd/naive
```
IRCody: We should update the readme to point to the binary built here instead of snapshotter.

mxpv: Done

Issue #35, opened 2018-12-03
Figure out why cri plugin must be disabled


IRCody:
If the cri plugin is enabled, our runtime causes errors upon exiting:

```
INFO[2018-12-03T11:02:09.030879139-08:00] TaskExit event &TaskExit{ContainerID:,ID:,Pid:224,ExitStatus:0,ExitedAt:2018-12-03 19:01:50.016792753 +0000 UTC,}
ERRO[2018-12-03T11:02:09.030966305-08:00] Failed to handle backOff event &TaskExit{ContainerID:,ID:,Pid:224,ExitStatus:0,ExitedAt:2018-12-03 19:01:50.016792753 +0000 UTC,} for   error="can't find container for TaskExit event: Prefix can't be empty"
```

The error "Prefix can't be empty" appears to be coming from a [pkg](https://github.com/moby/moby/blob/master/pkg/truncindex/truncindex.go#L17) in moby that is used by the [cri plugin](https://github.com/containerd/cri/blob/e312635396a13bc67cc5c9f3e5f873dc92cfc891/pkg/store/sandbox/sandbox.go#L64).

It's unclear why because AFAIK we are not invoking any cri functionality but having the plugin enabled seems to change the code path for exiting in some way.

mikebrow:
What containerd namespace did you use for the containers?

mikebrow:
Please turn the log to debug and let's see what the output looks like :-)

mikebrow:
See.. https://github.com/containerd/containerd/blob/master/api/events/task.pb.go#L75-L81

if TaskExit.ID is empty the code throws the empty error

mikebrow:
The issue is the containerid may be a dup.. so.. we use ID as a unique identifier to pick between the dups.

We generate it with this code:
// GenerateID generates a random unique id.
func GenerateID() string {
	b := make([]byte, 32)
	rand.Read(b)
	return hex.EncodeToString(b)
}

mikebrow:
I suppose we could ignore any task exit with missing ID.

IRCody:
So there is a containerId and ID. If you run via something like:
`./ctr run --runtime aws.firecracker --snapshotter firecracker-snapshotter --tty docker.io/library/alpine:latest 1234 /bin/sh`

The containerId should be 1234. Where is the ID for the TaskExitEvent expected to come form in this case?

[...]

PR #37, opened 2018-12-05
verify that id is set

mikebrow:
Signed-off-by: Mike Brown <brownwm@us.ibm.com>

Address Issue #35

Ensure that id is set, and note that it is a required argument.  

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"os"
 	"os/signal"
@@ -40,7 +41,7 @@ func main() {
 		debug bool
 	)
 
-	flag.StringVar(&id, "id", "", "Shim task id")
+	flag.StringVar(&id, "id", "", "ContainerID (required)")
 	flag.IntVar(&port, "port", defaultPort, "Vsock port to listen to")
 	flag.BoolVar(&debug, "debug", false, "Turn on debug mode")
 	flag.Parse()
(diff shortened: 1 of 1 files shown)


agent/main.go
```diff
@@ -40,7 +41,7 @@ func main() {
 		debug bool
 	)
 
-	flag.StringVar(&id, "id", "", "Shim task id")
+	flag.StringVar(&id, "id", "", "Shim task id (required)")
```
IRCody: The thought here is that this id is the ContainerId?

mikebrow: yes, i'll change the text

mikebrow:
looks like a go's new module requirements tooling has retry issues...

mikebrow:
pushed again as the build is flaky with go module updates timing out...

agent/main.go
IRCody: This doesn't completely address #35 because the id isn't known currently at the time the agent starts up. That could be changed in the future though so this change still makes sense.

mikebrow: interesting.. 

I figured you knew what the containerid was at this point, is this more an anonymous agent in the vm that will act as a proxy for new container(s) at some later time? We generally know the containerid before we create the task&shim to run the container in other flows.

nmeyerhans: At the moment, the agent starts upon VM startup, with no knowledge of the container. However, this will clearly need to change as we move toward a model in which multiple containers (e.g. a pod) shares a VM. So it's not unreasonable to assume that the ID will be known ahead of time by the agent.

IRCody:
@mikebrow: Thanks! Sorry for the delay in merging.

Issue #40, opened 2018-12-08
runtime: Implement CID management


nmeyerhans:
At the moment, we're hardcoding a vsock CID in the shim. This limits us to running a single instance per host. We should instead ensure that each VM has a unique CID.

PR #42, opened 2018-12-13
Detect available vsock context ID dynamically

mxpv:

*Issue #, if available:*  #40

*Description of changes:*
Detect available vsock CID using `VHOST_VSOCK_SET_GUEST_CID` ioctl

### DIFF
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"io"
 	"io/ioutil"
+	"math"
 	"net"
 	"os"
 	"os/exec"
@@ -24,6 +25,7 @@ import (
 	"strings"
 	"syscall"
 	"time"
+	"unsafe"
 
 	eventstypes "github.com/containerd/containerd/api/events"
 	"github.com/containerd/containerd/api/types"
@@ -37,20 +39,19 @@ import (
 	taskAPI "github.com/containerd/containerd/runtime/v2/task"
 	"github.com/containerd/fifo"
 	"github.com/containerd/ttrpc"
-	"github.com/firecracker-microvm/firecracker-containerd/internal"
-	"github.com/firecracker-microvm/firecracker-containerd/proto"
-	firecracker "github.com/firecracker-microvm/firecracker-go-sdk"
+	"github.com/firecracker-microvm/firecracker-go-sdk"
 	ptypes "github.com/gogo/protobuf/types"
 	"github.com/mdlayher/vsock"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
+
+	"github.com/firecracker-microvm/firecracker-containerd/internal"
+	"github.com/firecracker-microvm/firecracker-containerd/proto"
 )
 
 const (
 	defaultVsockPort = 10789
-	// TODO: This will need to be changed once we are managing CID's
-	defaultCID = uint32(3)
 )
 
 // implements shimapi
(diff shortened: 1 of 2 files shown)


runtime/service.go
```diff
@@ -489,6 +490,47 @@ func dialVsock(ctx context.Context, contextID uint32, port uint32) (net.Conn, er
 	return nil, lastErr
 }
 
+// findNextAvailableVsockCID finds first available vsock context ID.
+// It uses VHOST_VSOCK_SET_GUEST_CID ioctl which allows some CID ranges to be statically reserved in advance.
+// The ioctl fails with EADDRINUSE if cid is already taken and with EINVAL if the CID is invalid.
+// Taken from https://bugzilla.redhat.com/show_bug.cgi?id=1291851
+func findNextAvailableVsockCID() (uint32, error) {
```
xibz: Can we test this function somehow?

runtime/service.go
```diff
@@ -489,6 +490,47 @@ func dialVsock(ctx context.Context, contextID uint32, port uint32) (net.Conn, er
...
+	)
+
+	file, err := os.OpenFile("/dev/vhost-vsock", syscall.O_RDWR, 0666)
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to open vsock device")
+	}
+
+	defer file.Close()
+
+	for contextID := startCID; contextID < maxCID; contextID++ {
```
xibz: We may want to provide a context in this function since this could iterate for a very long time

runtime/service.go
```diff
@@ -489,6 +490,54 @@ func dialVsock(ctx context.Context, contextID uint32, port uint32) (net.Conn, er
...
+		select {
+		case <-ctx.Done():
+			return 0, ctx.Err()
+		default:
+			cid := contextID
+			_, _, err = unix.Syscall(
```
xibz: If we allow mocking of the `Syscall`, we can test this. How do you feel about adding a small abstraction that will allow for you to call like `var unixSyscall = unix.Syscall` then in the tests, set the `unixSyscall` to a mock function.

```go
func TestMock(t *testing.T) {
    unixSyscall = func(/*parameters*/) (uintptr, uintptr, syscall.Errno) {
        return 0, 0, syscall.Errno(0)
    }
}
}
```

runtime/service_test.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"context"
+	"syscall"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFindNextAvailableVsockCID(t *testing.T) {
+	sysCall = func(trap, a1, a2, a3 uintptr) (r1, r2 uintptr, err syscall.Errno) {
```
xibz: I would add the `unix.EADDRINUSE` case and default error case.

`unix.EADDRINUSE` could be squashed into this test
```go
count := 0
sysCall = func(trap, a1, a2, a3 uintptr) (r1, r2 uintptr, err syscall.Errno) {
    if count == 0 {
         return r1, r2, unix.EADDRINUSE
    }
    count++
    return 0, 0, 0
}
```

mxpv: Current implementation is fine. As of now I want to have our code running basic scenarios, not shiny and 100% covered with tests

runtime/service.go
```diff
@@ -489,6 +493,54 @@ func dialVsock(ctx context.Context, contextID uint32, port uint32) (net.Conn, er
...
+func findNextAvailableVsockCID(ctx context.Context) (uint32, error) {
+	const (
+		// Corresponds to VHOST_VSOCK_SET_GUEST_CID in vhost.h
+		ioctlVsockSetGuestCID = uintptr(0x4008AF60)
+		// 0, 1 and 2 are reserved CIDs, see http://man7.org/linux/man-pages/man7/vsock.7.html
+		startCID = 3
+		maxCID   = math.MaxUint32
+	)
+
+	file, err := os.OpenFile("/dev/vhost-vsock", syscall.O_RDWR, 0666)
```
IRCody: Can we move /dev/vhost-vsock to a constant.

mxpv: Done

runtime/service.go
```diff
@@ -537,25 +542,62 @@ func dialVsock(ctx context.Context, contextID uint32, port uint32) (net.Conn, er
...
+
+	defer file.Close()
 
+	for contextID := startCID; contextID < maxCID; contextID++ {
+		select {
+		case <-ctx.Done():
+			return 0, ctx.Err()
+		default:
+			cid := contextID
+			_, _, err = sysCall(
```
samuelkarp: Why not just use `syscall.Syscall` here?

runtime/service.go
```diff
@@ -537,25 +542,62 @@ func dialVsock(ctx context.Context, contextID uint32, port uint32) (net.Conn, er
...
-				specified vsock port.
-				TODO: We will need some sort of vsock CID accounting mechanism
-				to know what CID to use, defaults to 3 for now
-			- Return this client or error
-	*/
+// findNextAvailableVsockCID finds first available vsock context ID.
+// It uses VHOST_VSOCK_SET_GUEST_CID ioctl which allows some CID ranges to be statically reserved in advance.
+// The ioctl fails with EADDRINUSE if cid is already taken and with EINVAL if the CID is invalid.
+// Taken from https://bugzilla.redhat.com/show_bug.cgi?id=1291851
+func findNextAvailableVsockCID(ctx context.Context) (uint32, error) {
```
samuelkarp: I'd lean toward taking this out to either a separate file or a separate package.  It's logically distinct from pretty much everything else in this file.

runtime/service.go
```diff
@@ -537,25 +542,62 @@ func dialVsock(ctx context.Context, contextID uint32, port uint32) (net.Conn, er
...
+	const (
```
samuelkarp: Once we move this out of this current file, I think it would make sense to put the `const` block at the package level instead of the function level.

runtime/service.go
mxpv: To make it testable. See thread with Ben above.

runtime/service.go
mxpv: Will address in subsequent PRs as this one is already merged.

runtime/service.go
samuelkarp: Got it, thanks.

PR #49, opened 2018-12-21
Update for compatibility with Go Firecracker SDK interface changes

nmeyerhans:
Configuration and execution of the firecracker process and VM changed a bit recently. This change updates the runtime shim to support the newer abstractions.

Manual testing:
`make clean all && sudo make install`

runtime changes tested with `ctr run` with the following images:
```
docker.io/library/busybox:latest (including `--tty`)
docker.io/library/debian:stretch-slim (including `--tty`)
docker.io/library/hello-world:latest
```

### DIFF
--- a/snapshotter/.gitignore
+++ b/snapshotter/.gitignore
@@ -1,1 +1,2 @@
-snapshotter
+naive_snapshotter
+devmapper_snapshotter


runtime/service.go
```diff
@@ -598,44 +600,65 @@ func (s *service) startVM(ctx context.Context, request *taskAPI.CreateTaskReques
...
-		cfg.AdditionalDrives = append(cfg.AdditionalDrives, firecracker.BlockDevice{HostPath: mnt.Source, Mode: "rw"})
-	}
-
-	s.machine = firecracker.NewMachine(cfg, firecracker.WithLogger(log.G(ctx)))
-	s.machineCID = cid
-
-	log.G(ctx).Println("initializing machine")
-	if _, err := s.machine.Init(ctx); err != nil {
-		log.G(ctx).WithError(err).Error("machine init failed")
+		idx := strconv.Itoa(i + 2)
```
mxpv: I get confused every time I see these offsets :( Can we have `driveCount++` by any chance?

runtime/service.go
```diff
@@ -598,44 +600,65 @@ func (s *service) startVM(ctx context.Context, request *taskAPI.CreateTaskReques
...
+		root := false
+		ro := false
+		cfg.Drives = append(cfg.Drives,
+			models.Drive{
+				DriveID:      &idx,
+				PathOnHost:   firecracker.String(mnt.Source),
+				IsRootDevice: &root,
+				IsReadOnly:   &ro,
```
mxpv: Can these two be just inlined? 
```go
IsRootDrive: firecracker.Bool(false)
IsReadOnly:  firecracker.Bool(false)
```

runtime/service.go
nmeyerhans: In previous versions of the SDK, drive indexes were hidden within the SDK. I don't like SDK users to have to think about drive indexes at all...

mxpv: It depends. Potentially we might want to allow user to update state? https://github.com/firecracker-microvm/firecracker/blob/master/api_server/swagger/firecracker.yaml#L120

runtime/service.go
nmeyerhans: Yes, they should be inlined.

Beyond that, though, the client shouldn't need to care. These should also be handled within the SDK...

runtime/service.go
nmeyerhans: Yes, we will want to support that, but I don't think we need to use firecracker's internal indexing system to do so. Our slice of Drives in firecracker.Config struct should allow us to construct any necessary API call with information we already have (again, entirely within the SDK)

mxpv: As I understand it's not strictly should be a number. So user might want to assign something more meaningful (for example https://github.com/kata-containers/runtime/blob/master/virtcontainers/fc.go#L300).
I also can imagine use case when we might want to assign image layer's key/hash/... instead of having slice index.

nmeyerhans: Yep, that's fair. Either way, we can maintain a mapping of whatever the user wants to call the drive to whatever firecracker uses as its index internally. I don't see any reason to force the caller to think about how Firecracker itself organizes drives.

xibz: Personally, we should have the SDK be fundementally flexible. As in users should be given the power to set things as needed. While this isnt the most user friendly, we can fix that by adding helper functions like buildDrives which takes a list of parameters that builds each drive and assigns it an ID. This allows for both a powerful, flexbile, and easy to use SDK.

runtime/service.go
```diff
@@ -598,44 +600,63 @@ func (s *service) startVM(ctx context.Context, request *taskAPI.CreateTaskReques
...
+			CPUTemplate: models.CPUTemplate(s.config.CPUTemplate),
+			MemSizeMib:  256,
+		},
+		LogFifo:     s.config.LogFifo,
+		LogLevel:    s.config.LogLevel,
+		MetricsFifo: s.config.MetricsFifo,
+		Debug:       s.config.Debug,
+	}
+
+	idx := strconv.Itoa(1)
```
xibz: Out of curiosity why isn't this just `idx := "1"`? Im guessing `"0"` is reserved for something else?

runtime/service.go
samuelkarp: @xibz I think that makes sense, but if we're missing the higher-level features that actually make the SDK usable we're providing little value over the generated Swagger client.  We should prioritize these usability features and helper functions, so that a user of the SDK doesn't need to understand this level of detail about the Firecracker API but *can* if they need to.

runtime/service.go
```diff
@@ -598,44 +600,63 @@ func (s *service) startVM(ctx context.Context, request *taskAPI.CreateTaskReques
...
-		BinPath:         s.config.FirecrackerBinaryPath,
 		SocketPath:      s.config.SocketPath,
 		VsockDevices:    []firecracker.VsockDevice{{Path: "root", CID: cid}},
 		KernelImagePath: s.config.KernelImagePath,
 		KernelArgs:      s.config.KernelArgs,
-		RootDrive:       firecracker.BlockDevice{HostPath: s.config.RootDrive, Mode: "rw"},
-		CPUCount:        int64(s.config.CPUCount),
-		CPUTemplate:     firecracker.CPUTemplate(s.config.CPUTemplate),
-		MemInMiB:        256,
-		Console:         s.config.Console,
```
IRCody: Since Console was removed here, can it safely be removed from runtime/config.go Config struct?

runtime/service.go
nmeyerhans: Well that's certainly interesting. I can't find where that came from. There's _no way_ I'd have made up anything as insane as a 1-indexed array, so it must have been somewhere. But afaict the drive IDs have never even needed to be integers, let alone array indexes.

Let's fix that in a future PR.

runtime/service.go
nmeyerhans: As indicated below, drive ID doesn't even need to be an integer, let alone an index. So that can go. The drives builder from #43 should take care of this, anyway, so let's switch to that in another PR.

runtime/service.go
nmeyerhans: Removed.

Issue #45, opened 2018-12-20
Replace devicemapper with dmsetup


mxpv:
We need to replace `devicemapper` package from `moby` and invoke `dmsetup` command line tool instead.

PR #51, opened 2018-12-26
Replace devmapper package with dmsetup wrapper

mxpv:
*Issue #, if available:* https://github.com/firecracker-microvm/firecracker-containerd/issues/45

*Description of changes:*
This removes moby's `devmapper` package in favor of `dmsetup` wrapper


README.md
```diff
@@ -70,9 +70,8 @@ channel on the [Firecracker Slack](https://tinyurl.com/firecracker-microvm)
 Each of the components requires Go 1.11 and utilizes Go modules.  You must have
 a properly set up Go toolchain capable of building the components.
 
-The devicemapper snapshotter requires the C-language libdevmapper headers to be
-installed and available on your computer.  On Ubuntu, these headers can be
-installed from the `libdevmapper-dev` package.
+The devicemapper snapshotter requires `dmsetup` command line tool to be
+installed and available on your computer.
```
samuelkarp: How can you install the tool? (Pick a package name on a popular Linux distribution like Ubuntu, Debian, Fedora, or Amazon Linux 2.)

go.mod
```diff
@@ -20,7 +20,6 @@ require (
 	github.com/google/go-cmp v0.2.0 // indirect
 	github.com/hashicorp/go-multierror v1.0.0
 	github.com/mdlayher/vsock v0.0.0-20181130155850-676f733b747c
-	github.com/moby/moby v0.7.3-0.20181205005855-1895e082b613
```
samuelkarp: Yay!

snapshotter/devmapper/devmapper.go
```diff
@@ -200,10 +201,6 @@ func (dm *Snapshotter) Close() error {
 		result = multierror.Append(result, err)
 	}
 
-	if err := dm.pool.Close(context.Background(), false, false); err != nil {
```
samuelkarp: No equivalent is necessary?

snapshotter/devmapper/pool_device.go
```diff
@@ -47,38 +41,18 @@ type PoolDevice struct {
...
-	metaFile, err := os.Open(metaVolume)
+	version, err := dmsetup.Version()
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to open meta volume")
+		log.G(ctx).Errorf("dmsetup not available")
+		return nil, err
 	}
 
-	defer metaFile.Close()
+	log.G(ctx).Debugf("using dmsetup: %s", version)
```
samuelkarp: I'd actually want this at Info level.

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+	"regexp"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+)
+
+const (
```
samuelkarp: Do these constants need to be exported?

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+	DevMapperDir = "/dev/mapper/"
```
samuelkarp: This should probably be the default, but I'm not sure if we can safely assume that `/dev/mapper` is the right path across Linux distributions.

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+	SectorSize   = 512
+)
+
+// DeviceInfo represents device info returned by "dmsetup info"
```
samuelkarp: "dmsetup(8) provides more information on each of these fields."

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+// DeleteDevice sends "delete <deviceID>" message to the given thin-pool
+func DeleteDevice(poolName string, deviceID int) error {
+	_, err := dmsetup("message", poolName, "0", fmt.Sprintf("delete %d", deviceID))
+	return err
+}
+
+// RemoveDevice removes a device (see "dmsetup remove")
+func RemoveDevice(deviceName string, force, retry, deferred bool) error {
```
samuelkarp: I'm wondering if this might be slightly more friendly (even if it is more complicated):

```go
type RemoveDeviceOpt struct {
	Force    bool
	Retry    bool
	Deferred bool
}

type RemoveDeviceOptFn func(*RemoveDeviceOpt) error

func RemoveWithForce(opt *RemoveDeviceOpt) error {
	opt.Force = true
}

// ...

func RemoveDevice(deviceName string, opts ...RemoveDeviceOptFn) error {
	options := RemoveDeviceOpt{}
	for _, opt := range opts {
		if err := opt(options); err != nil {
			return err
		}
	}
	// ...
}
```

The reason it might be more friendly is that you could write `dmsetup.RemoveDevice("myDevice", RemoveWithForce, RemoveWithDeferred)` instead of `dmsetup.RemoveDevice("myDevice", true, false, true)`.

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+			&info.Minor,
+			&info.OpenCount,
+			&info.TargetCount,
+			&info.EventNumber)
+
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to parse line %q", line)
+		}
+
+		// Parse attributes (see "man dmsetup" for details)
```
samuelkarp: "man 8 dmsetup"

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+		return nil, err
+	}
+
+	var (
+		lines   = strings.Split(output, "\n")
+		devices = make([]*DeviceInfo, len(lines))
+	)
+
+	for i, line := range lines {
+		var (
```
samuelkarp: Any particular reason not to do this?

```go
attr := ""
info := &DeviceInfo{}
```

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+		"name,blkdevname,attr,major,minor,open,segments,events",
+		"--separator",
+		" ",
+		deviceName)
+
+	if err != nil {
+		return nil, err
+	}
+
+	var (
```
samuelkarp: Any particular reason not to do this?

```go
lines := strings.Split(output, "\n")
devices := make([]*DeviceInfo, len(lines))
```

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+// tryGetUnixError tries to find Linux error code from dmsetup output
+func tryGetUnixError(output string) (unix.Errno, bool) {
+	// It's useful to have Linux error codes like EBUSY, EPERM, ..., instead of just text.
+	// Unfortunately there is no better way than extracting/comparing error text.
+	text := parseDmsetupError(output)
+	if text == "" {
+		return 0, false
+	}
+
+	errTableInit.Do(func() {
```
samuelkarp: Can you move this out to a separate function?

```go
func initErrTable() {
	errTableInit.Do(func() {
		//...
	})
}
```

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+	err, ok := errTable[text]
+	return err, ok
+}
+
+// dmsetup returns error messages in format:
+// 	device-mapper: message ioctl on <name> failed: File exists\n
+// 	Command failed\n
+// parseDmsetupError extracts text between "failed: " and "\n"
+func parseDmsetupError(output string) string {
+	re := regexp.MustCompilePOSIX("failed: [^:]+$")
```
samuelkarp: Not super wild about regex on arbitrary strings.  Can we just do a search for "failed: " (`strings.Contains`) and then split at ":" (`strings.SplitN`)?

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+	str := re.FindString(output)
+
+	// Strip "failed: " prefix
+	str = strings.TrimPrefix(str, "failed: ")
+
+	str = strings.ToLower(str)
+	return str
+}
+
+// makeThinPoolMapping makes thin-pool table entry
```
samuelkarp: Can you move this up, right below `ReloadPool`?

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+	// low_water_mark - the low water mark, expressed in blocks of size data_block_size
+	// feature_args - the number of feature arguments
+	// args
+	lengthSectors := dataDeviceSizeBytes / SectorSize
+	target := fmt.Sprintf("0 %d thin-pool %s %s %d 32768 1 skip_block_zeroing", lengthSectors, metaFile, dataFile, blockSizeSectors)
+
+	return target, nil
+}
+
+// makeThinMapping makes thin target table entry
```
samuelkarp: Can you move this up, right below `ActivateDevice`?

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+	// start - starting block in virtual device
+	// length - length of this segment
+	// metadata_dev - the metadata device
+	// data_dev - the data device
+	// data_block_size - the data block size in sectors
+	// low_water_mark - the low water mark, expressed in blocks of size data_block_size
+	// feature_args - the number of feature arguments
+	// args
+	lengthSectors := dataDeviceSizeBytes / SectorSize
+	target := fmt.Sprintf("0 %d thin-pool %s %s %d 32768 1 skip_block_zeroing", lengthSectors, metaFile, dataFile, blockSizeSectors)
```
samuelkarp: I'd recommend doing a bit more here:

(top of the file, or directly above the function)
```go
const (
	lowWaterMark = 32768 // Picked as low water mark because [reason]
	skipZeroing  = "skip_block_zeroing" // Skipping zeroing to reduce latency for device creation
)
```

```go
fmt.Sprintf("0 %d thin-pool %s %s %d %d 1 %s",
	lengthSectors,
	metaFile,
	dataFile,
	blockSizeSectors
	lowWaterMark,
	skipZeroing)
```

Moving these things to named constants lets you leave comments on why we picked them, to accompany your existing comment about the command structure.

snapshotter/devmapper/devmapper.go
mxpv: No, we want pool remain exist after we shutdown snapshotter (so we don't loose data)

snapshotter/pkg/dmsetup/dmsetup.go
mxpv: Yes, I think so. For example sector size is needed in Config validation.

snapshotter/devmapper/devmapper.go
samuelkarp: Got it, was this an error in the previous implementation then?

mxpv: It actually did nothing because both `removeDevices` and  `removePool` were set to false.

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+	// Unfortunately there is no better way than extracting/comparing error text.
+	text := parseDmsetupError(output)
+	if text == "" {
+		return 0, false
+	}
+
+	errTableInit.Do(func() {
+		// Precompute map of <text>=<errno> for optimal lookup
+		errTable = make(map[string]unix.Errno)
+		for errno := unix.EPERM; errno <= unix.EHWPOISON; errno++ {
```
xibz: Why not have this in an `init` function instead? Would get rid of the `sync.Once`

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+	if err != nil {
+		return 0, errors.Wrapf(err, output)
+	}
+
+	output = strings.TrimSuffix(output, "\n")
+	return strconv.ParseUint(output, 10, 64)
+}
+
+func dmsetup(args ...string) (string, error) {
+	data, err := exec.Command("dmsetup", args...).CombinedOutput()
```
xibz: Do we really want both `stdout` and `stderr`? Looks like `Table` is the only thing that uses the output, so I am curious what would be the behavior if the output was not of the expected format, like some `stderr` stuff in there?

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,310 @@
...
+	if strings.HasPrefix(deviceName, DevMapperDir) {
+		return deviceName
+	}
+
+	return DevMapperDir + deviceName
+}
+
+// BlockDeviceSize returns size of block device in bytes
+func BlockDeviceSize(devicePath string) (uint64, error) {
+	data, err := exec.Command("blockdev", "--getsize64", "-q", devicePath).CombinedOutput()
```
xibz: Same as below with wanting `stdout` and `stderr`

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,338 @@
...
+	output = strings.TrimSpace(output)
+
+	return output, nil
+}
+
+func initErrTable() {
+	errTableInit.Do(func() {
+		// Precompute map of <text>=<errno> for optimal lookup
+		errTable = make(map[string]unix.Errno)
+		for errno := unix.EPERM; errno <= unix.EHWPOISON; errno++ {
```
xibz: This for loop scares me a little as the error numbers is dispersed.

snapshotter/pkg/dmsetup/dmsetup.go
mxpv: In case of error `dmsetup` and/or `blockdev` will return non zero error code, so `exec.Command` will return an error. `CombinedOutput` is needed because I want to capture all error output and return it in error message, so we have error details

snapshotter/pkg/dmsetup/dmsetup.go
mxpv: What do you mean? Range of error codes goes from 1 to 133, see [here](https://github.com/golang/sys/blob/master/unix/zerrors_linux_amd64.go#L2568)

snapshotter/pkg/dmsetup/dmsetup.go
mxpv: See comment above

snapshotter/pkg/dmsetup/dmsetup.go
xibz: oh good! I was looking at the `const` list and saw that it was pretty random. Ignore me :D

snapshotter/pkg/dmsetup/dmsetup.go
xibz: Is the output used for logging? Or how is the string to be used?

mxpv: Output is used to add error details, see [here](https://github.com/firecracker-microvm/firecracker-containerd/pull/51/files#diff-dc540783dc366cf76abc1b48546be56eR279) and eventually will be logged by snapshotter or containerd

xibz: Do we need to return it then? I am just wondering since most calls to `dmsetup` ignore the the first returned parameter.

mxpv:
Done

PR #53, opened 2018-12-27
Initial README for the devmapper snapshotter

samuelkarp:
*Description of changes:*
* Moved the existing README to the naive folder
* Created a new README for devmapper


snapshotter/cmd/devmapper/README.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+The name of the plugin in this example is "firecracker-dm-snapshotter".  The
+`address` entry points to a socket file exposed by the snapshotter, which is
+determined when you run it.
+
+## Usage
+
+```
+./devmapper_snapshotter -address UNIX-DOMAIN-SOCKET -path ROOT -debug
```
mxpv: We don't use `-path` for devmapper (we have optional `-config` instead)

samuelkarp: Oops, missed that...

Issue #57, opened 2018-12-27
Document the runtime configuration file and its required fields


samuelkarp:
We provide an [example](https://github.com/firecracker-microvm/firecracker-containerd/blob/bd30957b56796f8d9765d5c4a96197a30906dd4c/runtime/config.json.example), but no documentation on what the fields mean or how they should be set.

PR #58, opened 2018-12-27
runtime: document the config file and its fields

samuelkarp:
*Issue #, if available:*
https://github.com/firecracker-microvm/firecracker-containerd/issues/57

*Description of changes:*
Documentation

### DIFF
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -26,6 +26,41 @@ shim:
 	* If the runtime is invoked as aws.firecracker
 	* Then the binary name needs to be containerd-shim-aws-firecracker
 
+## Configuration
+
+The runtime expects a JSON-formatted configuration file to be located either in
+`/etc/containerd/firecracker-runtime.json` or in a location defined by the
+`FIRECRACKER_CONTAINERD_RUNTIME_CONFIG_PATH` environment variable.  The
+configuration file has the following fields:
+
+* `firecracker_binary_path` (optional) - A path to locate the `firecracker`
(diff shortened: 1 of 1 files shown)


runtime/README.md
```diff
@@ -26,6 +26,41 @@ shim:
...
+  communicating with the Firecracker API.  A relative path like
+  `./firecracker.sock` is recommended so that the socket is created in the
+  temporary working directory allocated by containerd.
+* `kernel_image_path` (required) - A path where the kernel image file is
+  located.  A fully-qualified path is recommended.
+* `kernel_args` (required) - Arguments for the kernel command line.
+* `root_drive` (required) - A path where the root drive image file is located. A
+  fully-qualified path is recommended.
+* `cpu_count` (required) - The number of vCPUs to make available to a microVM.
+* `cpu_template` (required) - The Firecracker CPU emulation template.  Supported
```
IRCody: Is the difference between C3 and T2 cpu template documented anywhere?

samuelkarp: Very minimal documentation [here](https://github.com/firecracker-microvm/firecracker/blob/v0.12.0/api_server/swagger/firecracker.yaml#L318-L325).

The actual difference is the difference between these two files: [C3](https://github.com/firecracker-microvm/firecracker/blob/v0.12.0/cpuid/src/c3_template.rs) and [T2](https://github.com/firecracker-microvm/firecracker/blob/v0.12.0/cpuid/src/t2_template.rs).

PR #62, opened 2019-01-08
Rework transaction handling

mxpv:

*Description of changes:*
This PR replaces `complete` with `withTransaction` func so we don't have to return `complete(ctx, trans, err)` on every error.
Also there is no way transaction can be forgotten to Commit/Rollback.
`withTransaction` also reports transaction errors in addition to snapshotter errors.

### DIFF
--- a/snapshotter/devmapper/devmapper.go
+++ b/snapshotter/devmapper/devmapper.go
@@ -89,35 +89,29 @@ func NewSnapshotter(ctx context.Context, configPath string) (*Snapshotter, error
 func (dm *Snapshotter) Stat(ctx context.Context, key string) (snapshots.Info, error) {
 	log.G(ctx).WithField("key", key).Debug("stat")
 
-	ctx, trans, err := dm.store.TransactionContext(ctx, false)
-	if err != nil {
-		return snapshots.Info{}, err
-	}
+	var (
+		info snapshots.Info
+		err  error
+	)
 
-	defer trans.Rollback()
+	err = dm.withTransaction(ctx, false, func(ctx context.Context) error {
+		_, info, _, err = storage.GetInfo(ctx, key)
+		return err
+	})
 
-	_, info, _, err := storage.GetInfo(ctx, key)
-	if err != nil {
-		return snapshots.Info{}, err
-	}
-
-	return info, nil
+	return info, err
 }
 
 func (dm *Snapshotter) Update(ctx context.Context, info snapshots.Info, fieldpaths ...string) (snapshots.Info, error) {
 	log.G(ctx).Debugf("update: %s", strings.Join(fieldpaths, ", "))
 
-	ctx, trans, err := dm.store.TransactionContext(ctx, true)
-	if err != nil {
-		return snapshots.Info{}, err
-	}
+	var err error
+	err = dm.withTransaction(ctx, true, func(ctx context.Context) error {
+		info, err = storage.UpdateInfo(ctx, info, fieldpaths...)
+		return err
+	})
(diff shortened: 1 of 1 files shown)


snapshotter/devmapper/devmapper.go
```diff
@@ -312,18 +310,44 @@ func (dm *Snapshotter) buildMounts(snap storage.Snapshot) []mount.Mount {
 	return mounts
 }
 
-func complete(ctx context.Context, trans storage.Transactor, err error) error {
-	if err == nil {
+// withTransaction wraps fn callback with containerd's meta store transaction.
+// If callback returns an error or transaction is not writable, database transaction will be discarded.
+func (dm *Snapshotter) withTransaction(ctx context.Context, writable bool, fn func(ctx context.Context) error) error {
+	ctx, trans, err := dm.store.TransactionContext(ctx, writable)
```
xibz: It seems `writable` parameter isn't needed here. Why not add a new helper function instead?
instead of
```go
	return dm.withTransaction(ctx, false, func(ctx context.Context) error {
		return storage.WalkInfo(ctx, fn)
	})
```
how about
```go
// probably needs a better name
func threadSafeCall(ctx context.Context, fn func(ctx context.Context) error) error  {
	m := sync.Mutex
	m.Lock()
	defer m.Unlock()
	return fn(ctx)
}
```
```go
	return threadSafeCall(ctx, func(ctx context.Context) error {
		return storage.WalkInfo(ctx, fn)
	})
```

snapshotter/devmapper/devmapper.go
```diff
@@ -312,18 +310,44 @@ func (dm *Snapshotter) buildMounts(snap storage.Snapshot) []mount.Mount {
 	return mounts
 }
 
-func complete(ctx context.Context, trans storage.Transactor, err error) error {
-	if err == nil {
+// withTransaction wraps fn callback with containerd's meta store transaction.
+// If callback returns an error or transaction is not writable, database transaction will be discarded.
+func (dm *Snapshotter) withTransaction(ctx context.Context, writable bool, fn func(ctx context.Context) error) error {
```
xibz: would be nice to ensure the behavior of this is tested. As in when an error occurs, the data isn't commited.

snapshotter/devmapper/devmapper.go
mxpv: Transactions are required by containerd's `storage` package, even if they are readonly.  Also see comment [here](https://github.com/containerd/containerd/blob/adfaa697a8b778eb319a49c02406c84bc0f1bdf9/snapshots/storage/metastore.go#L41-L44).

snapshotter/devmapper/devmapper.go
mxpv: containerd's snapshotter suite tests cover error scenarios as well (for example [here](https://github.com/containerd/containerd/blob/adfaa697a8b778eb319a49c02406c84bc0f1bdf9/snapshots/testsuite/testsuite.go#L720) and [here](https://github.com/containerd/containerd/blob/adfaa697a8b778eb319a49c02406c84bc0f1bdf9/snapshots/testsuite/testsuite.go#L173)). Not sure it makes a lot of sense to cover same code for our own unit tests.

xibz: Ah, cool. Wasnt aware.

snapshotter/devmapper/devmapper.go
xibz: Interesting. Alright, lgtm

PR #63, opened 2019-01-09
agent: build a statically-linked binary

samuelkarp:
*Description of changes:*

The agent runs inside an arbitrary rootfs and may not have glibc available.  Build a statically-linked binary so it can be run in a rootfs that lacks glibc.

### DIFF
--- a/agent/Makefile
+++ b/agent/Makefile
@@ -14,7 +14,7 @@
 all: agent
 
 agent: *.go
-	go build -o agent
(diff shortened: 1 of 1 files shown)


nmeyerhans:
Take a look at #64 for an alternative, which lets us choose the linker settings at build-time. It defaults to dynamic linking, with static selectable by setting a variable.

samuelkarp:
@nmeyerhans Yep, that's better.  Thanks!

PR #67, opened 2019-01-15
Validate commit messages and file headers

mxpv:

This will lint commit messages:
- DCO - require `signed off`
- dangling-whitespace - prevent dangling whitespaces on line endings
- short-subject - required <90 

Note: I've added `-range` flag since there are already a lot of invalid commit messages in master branch.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,7 +32,7 @@ To send us a pull request, please:
 1. Fork the repository.
 2. Modify the source; please focus on the specific change you are contributing. If you also reformat all the code, it will be hard for us to focus on your change.
 3. Ensure local tests pass.
-4. Commit to your fork using clear commit messages. Make sure your commits are signed (just add a line to every git commit message: `Signed-off-by: John Doe <john.doe@mail.com>` or use `git commit -s`)
+4. Commit to your fork using clear commit messages. Make sure your commits are signed off (just add a line to every git commit message: `Signed-off-by: John Doe <john.doe@mail.com>` or use `git commit -s`)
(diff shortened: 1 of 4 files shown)


mxpv:
Yep. We didn't mention that. I've updated contribution guide.

CONTRIBUTING.md
```diff
@@ -32,7 +32,7 @@ To send us a pull request, please:
 1. Fork the repository.
 2. Modify the source; please focus on the specific change you are contributing. If you also reformat all the code, it will be hard for us to focus on your change.
 3. Ensure local tests pass.
-4. Commit to your fork using clear commit messages.
+4. Commit to your fork using clear commit messages. Make sure your commits are signed (just add a line to every git commit message: `Signed-off-by: John Doe <john.doe@mail.com>` or use `git commit -s`)
```
IRCody: I think you mean "signoff" instead of signed here.

Makefile
```diff
@@ -29,8 +29,12 @@ clean:
 deps:
 	test -n "$(GOPATH)" # Make sure GOPATH defined
 	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b ${GOPATH}/bin v1.12.3
+	go get -u github.com/vbatts/git-validation
```
samuelkarp: These lines cause my `go.mod` and `go.sum` to get modified, because running `go get` when modules are enabled is how you add dependencies to a module.  I think we don't want to do that.

samuelkarp: Thanks!

PR #69, opened 2019-01-17
docs: Updating getting started guide to work with latest

xibz:
This makes changes to the getting started guide which reflected an older version of the package.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -254,7 +254,7 @@ configuration file has the following fields:
 Start the containerd snapshotter
 
 ```bash
-$ ./snapshotter /var/run/firecracker-snapshotter.sock /tmp/fc-snapshot
+$ ./naive-snapshotter -address /var/run/firecracker-containerd/naive-snapshotter.sock -path /tmp/fc-snapshot
 ```
 
 In another terminal, start containerd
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -191,7 +191,7 @@ Add the snapshotter plugin to your `/etc/containerd/config.toml`
 [proxy_plugins]
   [proxy_plugins.firecracker-naive]
     type = "snapshot"
-    address = "/var/run/firecracker-containerd/naive-snapshotter.sock"
+    address = "/var/run/firecracker-snapshotter.sock"
```
samuelkarp: Can you change this back?

docs/getting-started.md
```diff
@@ -254,7 +254,7 @@ configuration file has the following fields:
 Start the containerd snapshotter
 
 ```bash
-$ ./snapshotter /var/run/firecracker-snapshotter.sock /tmp/fc-snapshot
+$ ./naive-snapshotter -address /var/run/firecracker-snapshotter.sock -path /tmp/fc-snapshot
```
samuelkarp: Instead of changing the path above, can you change it here?

docs/getting-started.md
xibz: Done

Issue #61, opened 2019-01-04
Don't hold transaction lock while activating/deactivating thin-device


mxpv:
Device activation/deactivation can potentially take too long and we don't want to keep an open write transaction for this.

samuelkarp:
In general, I don't really like the idea of doing disk operations while holding an active lock on the database (holding open a transaction).  Instead, we can record that an activation/deactivation is taking place and then update the record after the activity completes.

PR #71, opened 2019-01-17
Manage device states, rework device removal

mxpv:
*Issue #, if available:*
https://github.com/firecracker-microvm/firecracker-containerd/issues/61

*Description of changes:*
This PR adds device states instead of transactions and reworks device removal. 

### DIFF
--- a/snapshotter/devmapper/pool_device.go
+++ b/snapshotter/devmapper/pool_device.go
@@ -210,13 +210,13 @@ func (p *PoolDevice) CreateSnapshotDevice(ctx context.Context, deviceName string
 
 // DeactivateDevice deactivates thin device
 func (p *PoolDevice) DeactivateDevice(ctx context.Context, deviceName string, deferred bool) error {
-	info, err := p.metadata.GetDevice(ctx, deviceName)
-	if err != nil {
-		return errors.Wrapf(err, "failed to query device %q for deactivation", deviceName)
-	}
+	devicePath := dmsetup.GetFullDevicePath(deviceName)
+	if _, err := os.Stat(devicePath); err != nil {
+		if os.IsNotExist(err) {
+			return ErrNotFound
+		}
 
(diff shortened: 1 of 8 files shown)


snapshotter/devmapper/pool_device.go
```diff
@@ -73,87 +73,190 @@ func NewPoolDevice(ctx context.Context, config *Config) (*PoolDevice, error) {
 	}, nil
 }
 
+// transition invokes 'fn' callback to perform devmapper operation and reflects device state changes/errors in meta store.
+// 'tryingState' will be set before invoking callback. If callback succeeded 'successState' will be set, otherwise
+// error details will be recorded in meta store.
+func (p *PoolDevice) transition(ctx context.Context, deviceName string, tryingState DeviceState, successState DeviceState, fn func() error) error {
```
xibz: nit: I would probably make this a graph of states. Something like
```go
// or have this defined in the PoolDevice
var stateGraph = map[DeviceState]DeviceState{
    Resuming: Resumed,
    // and so on.
}

func (p *PoolDevice) transition(ctx context.Context, deviceName, state DeviceState, fn func() error) error {
}
```

xibz: Can we also have `fn` be named something a little more explicit since it is an ambiguous closure? Something like `devMapperFn func() error`

mxpv: I'm not sure about this idea, this adds one more place to manage device states.
So if you forget to update keys, you'll get runtime error, which is not good.
Also I think it's better to explicitly define from/to state we're transitioning.

mxpv: would `transitionFn` work for you ?

mxpv: I tend to leave it as is.

Issue #15, opened 2018-11-26
Snapshotter improvements


IRCody:
The existing snapshotter is, by design, simple and broadly compatible but not very efficient. We should implement an alternative snapshotter following a model similar to Docker's device-mapper storage driver.

mxpv:
https://github.com/containerd/containerd/pull/3022

PR #73, opened 2019-01-23
Support container-storage-setup tool 

mxpv:
*Description of changes:*
This makes `devmapper_snapshotter` compatible with `container-storage-setup` tool and Docker's `dm.*` flags. See `README` for more details.

relates to #15 


snapshotter/cmd/devmapper/flag.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+// parseKeyValueOpt splits a string to key=value
+func parseKeyValueOpt(opt string) (string, string, error) {
+	pair := strings.SplitN(opt, "=", 2)
```
xibz: `strings.SplitN("=", "=", 20)` will return `[ "", "" ]`. Is that correct behavior?
`strings.SplitN("=bar")` will return `["", "bar"]`
`strings.SplitN("bar=")` will return `["bar", ""]`.

snapshotter/cmd/devmapper/flag.go
```diff
@@ -0,0 +1,59 @@
...
+func visitKeyValueOpts(optName string, optFn func(key, value string) error) error {
+	// Nothing to visit
+	if len(os.Args) < 3 {
+		return nil
+	}
+
+	// Skip exec path
+	args := os.Args[1:]
+
+	for i := 0; i < len(args)-1; i++ {
```
xibz: I am suspecting the last element is a file or something?

snapshotter/cmd/devmapper/main.go
```diff
@@ -30,10 +34,55 @@ const (
...
+	if rootPath != "" {
+		config.RootPath = rootPath
+	}
+
+	// Append and/or overwrite file configuration with --storage-opt dm.XXX=YYY command line flags
+	if err := visitKeyValueOpts("--storage-opt", func(key, value string) error {
+		return applyStorageOpt(ctx, key, value, config)
+	}); err != nil {
+		log.G(ctx).WithError(err).Fatal("failed to apply storage options")
+		return
```
xibz: shouldn't we `os.Exit(1)` on error cases?

snapshotter/cmd/devmapper/main.go
```diff
@@ -30,10 +34,55 @@ const (
...
+	}
+
+	if err := config.Validate(); err != nil {
+		log.G(ctx).WithError(err).Fatal("invalid configuration")
+		return
```
xibz: same here

snapshotter/cmd/devmapper/main.go
xibz: Oh, sorry, the `return` made me thing it wasn't calling `Fatal`.

snapshotter/cmd/devmapper/flag.go
xibz: Ah, cool. Remember seeing that but didn't connect it above.

snapshotter/cmd/devmapper/flag.go
mxpv: Technically empty key would break anything, it just wouldn't much any key we expect. But I've added an extra error check.

snapshotter/cmd/devmapper/main.go
xibz: How do you feel about getting rid of the `return` statements since they don't really do anything.

mxpv: They don't, but I've added them to explicitly state that program is not going to continue execution from this block, because `log.G(ctx) ... ` something might not be clear enough

xibz: I would much prefer a comment to signify things. I think adding code to signify something when it does nothing will lead to confusion. I was confused at least ><

mxpv: I don't want to write a comment instead of every return, there are log error messages for that.
Also I don't understand how `return` may confuse someone :)

snapshotter/cmd/devmapper/flag.go
```diff
@@ -0,0 +1,59 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
mxpv: That will break linter as it expects header to be exactly same as in `.headers/go.txt` (https://github.com/firecracker-microvm/firecracker-containerd/blob/master/.headers/go.txt).
Let's address this in separate PR.

snapshotter/cmd/devmapper/main.go
```diff
@@ -30,10 +34,52 @@ const (
...
+		config     = &devmapper.Config{}
+		configPath = ""
+		rootPath   = ""
+	)
+
 	flag.StringVar(&configPath, "config", "", "Path to devmapper configuration file")
+	flag.StringVar(&rootPath, "path", "", "Path to snapshotter data")
+
+	// These flags are needed for compatibility with container-storage-setup tool as there is no way to
+	// ignore unknown flags in `flag` package. Storage options are parsed using visitKeyValueOpts func.
```
samuelkarp: You can use a `flag.FlagSet` and change the `flag.ErrorHandling` to use `flag.ContinueOnError`.

snapshotter/cmd/devmapper/main.go
```diff
@@ -30,10 +34,52 @@ const (
...
+		ctx        = context.Background()
+		config     = &devmapper.Config{}
+		configPath = ""
+		rootPath   = ""
+	)
+
 	flag.StringVar(&configPath, "config", "", "Path to devmapper configuration file")
+	flag.StringVar(&rootPath, "path", "", "Path to snapshotter data")
+
+	// These flags are needed for compatibility with container-storage-setup tool as there is no way to
```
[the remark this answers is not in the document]
mxpv: Yes, it does. `flags` handles double dashes https://github.com/golang/go/blob/master/src/flag/flag.go#L898

snapshotter/cmd/devmapper/main.go
mxpv: These lines are needed to prevent errors like `flag provided but not defined: -storage-opt`.
With `FlagSet` I still need to register them, so that wouldn't help.

samuelkarp: Discussed offline; using a FlagSet requires more refactoring and we can come back to that later.

PR #81, opened 2019-02-01
devmapper: implement usage

mxpv:

*Issue #, if available:*

*Description of changes:*
This PR implements snapshotter's `Usage` method.
Also adds a few minor fixes and improvements.

### DIFF
--- a/snapshotter/devmapper/snapshotter.go
+++ b/snapshotter/devmapper/snapshotter.go
@@ -112,11 +112,32 @@ func (s *Snapshotter) Update(ctx context.Context, info snapshots.Info, fieldpath
 	return info, err
 }
 
-// Usage not yet implemented
+// Usage returns the resource usage of an active or committed snapshot excluding the usage of parent snapshots.
 func (s *Snapshotter) Usage(ctx context.Context, key string) (snapshots.Usage, error) {
 	log.G(ctx).WithField("key", key).Debug("usage")
 
-	return snapshots.Usage{}, errors.New("usage not implemented")
+	var (
+		id  string
+		err error
+	)
+
+	err = s.withTransaction(ctx, false, func(ctx context.Context) error {
+		id, _, _, err = storage.GetInfo(ctx, key)
+		return err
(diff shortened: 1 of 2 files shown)


snapshotter/devmapper/snapshotter_test.go
```diff
@@ -16,61 +16,114 @@ package devmapper
...
-
-			return err
-		}, nil
+		snap := createSnapshotter(t, root)
+		return snap, snap.Close, nil
+	})
+}
+
+func TestSnapshotterUsage(t *testing.T) {
+	tempDir, err := ioutil.TempDir("", "devmapper-snapshotter-")
```
samuelkarp: This looks like it needs `internal.RequiresRoot(t)`

snapshotter/devmapper/snapshotter.go
```diff
@@ -112,11 +112,32 @@ func (s *Snapshotter) Update(ctx context.Context, info snapshots.Info, fieldpath
...
+	}
+
+	deviceName := s.getDeviceName(id)
+	devicePath := dmsetup.GetFullDevicePath(deviceName)
+	blockDeviceSize, err := dmsetup.BlockDeviceSize(devicePath)
+	if err != nil {
+		return snapshots.Usage{}, err
+	}
+
+	return snapshots.Usage{Size: int64(blockDeviceSize)}, nil
```
xibz: Out of curiosity, what about the `Inodes` field? And out of curiosity why can't you use the `Size` that is returned by `GetInfo`?

xibz: Awesome, would be curious about those benchmarks. I think for me to approve this PR, I would either like to see this implemented in the `GetInfo` portion and have it calculate inodes, and if it turns out to be a huge hit on performance figure something out or have a comment signifying a that inodes are missing from the `Usage` call. May be a good idea to have a TODO about the inode stuff all together.

mxpv:
Will be reworked in separate PR.

PR #89, opened 2019-02-07
Add snapshotter benchmarks

mxpv:
*Description of changes:*
This PR adds snapshotter benchmarks and cloud formation template for EC2 test environment.


snapshotter/benchmark/benchmark_test.go
```diff
@@ -0,0 +1,206 @@
...
+	dmRootPath      string
+	overlayRootPath string
+	naiveRootPath   string
+)
+
+func init() {
+	flag.StringVar(&dmPoolDev, "dm.thinPoolDev", "", "Pool device to run benchmark on")
+	flag.StringVar(&dmRootPath, "dm.rootPath", "", "Root dir for devmapper snapshotter")
+	flag.StringVar(&overlayRootPath, "overlay.rootPath", "", "Root dir for overlay snapshotter")
+	flag.StringVar(&naiveRootPath, "naive.rootPath", "", "Root dir for naive snapshotter")
```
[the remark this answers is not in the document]
mxpv: The idea here is to run each snapshotter on separate EBS volume, so they have exactly same conditions.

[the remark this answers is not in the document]
mxpv: Also I keep three path variables instead of just one for convenience, so you can run all benchmarks with one command and get unified output in just one table. Theoretically we can have one variable for this, but you have to run each benchmark separately.

snapshotter/benchmark/benchmark_test.go
```diff
@@ -0,0 +1,206 @@
...
+	b.ResetTimer()
+
+	b.RunParallel(func(pb *testing.PB) {
+		ctx := context.Background()
+
+		for pb.Next() {
+			parent := ""
+
+			for l := 0; l < layerCount; l++ {
+				current := fmt.Sprintf("prepare-layer-%d", atomic.AddInt64(&layerIndex, 1))
```
xibz: What happens if this is the same value? The reason why I am asking is having atomics in a parallel test seems like there's an issue with the test. I am wondering if we can get rid of them with a hard coded value instead.

xibz:
I don't like benchmarks being in its own package. Can we just move this to the `devmapper` package and make `ResetPools` unexported?

mxpv:
> I don't like benchmarks being in its own package. Can we just move this to the `devmapper` package and make `ResetPools` unexported?

The benchmark package also contains tests for `Naive` and `Overlay` snapshotters and CF template, also we might consider other implementations, so I'm not sure whether it should be `devmapper` specific.

xibz:
> > I don't like benchmarks being in its own package. Can we just move this to the `devmapper` package and make `ResetPools` unexported?
> 
> The benchmark package also contains tests for `Naive` and `Overlay` snapshotters and CF template, also we might consider other implementations, so I'm not sure whether it should be `devmapper` specific.

Perhaps separate out the benchmarks? They don't need to be together.

mxpv:
> Perhaps separate out the benchmarks? They don't need to be together.

This is basically same benchmark (which accepts `Snapshotter` interface), the initialization logic is different for each implementation, so they are pretty related. Also where CF template should be located than?

xibz:
> > Perhaps separate out the benchmarks? They don't need to be together.
> 
> This is basically same benchmark (which accepts `Snapshotter` interface), the initialization logic is different for each implementation, so they are pretty related. Also where CF template should be located than?

I usually put test artifacts in a `testdata` folder. That's generally idiomatic Go. I think having duplicate code isn't bad in tests. Either have duplicated code or have all the benchmarks in `snapshotter` package instead. My concern of the exposure of `ResetPools` isn't too big, but if we can make it unexported that'd be awesome. However, my main concern is having a separate package for just benchmarks.

mxpv:
Would moving one level up (to `snapshotters`dir) works for you?
`benchmarkSnapshotter` and `makeApplier` are pretty large funcs, so I don't want to duplicate them (even when sometimes it's acceptable).

snapshotter/benchmark_aws.yml
```diff
@@ -0,0 +1,122 @@
...
+
+Parameters:
+  Key:
+    Type: AWS::EC2::KeyPair::KeyName
+    Description: SSH key to use
+
+  AMI:
+    Type: AWS::EC2::Image::Id
+    Description: AMI ID to use for the EC2 instance. Must be RHEL or CentOS
+    Default: "ami-3ecc8f46"
```
samuelkarp: This AMI is vended through the AWS Marketplace and would require a (free?) subscription in order to launch successfully.  It's also somewhat old ("CreationDate": "2018-06-13T15:58:14.000Z") and I wonder if there's a newer one that we can use.

snapshotter/benchmark_test.go
```diff
@@ -0,0 +1,206 @@
...
+	b.RunParallel(func(pb *testing.PB) {
+		ctx := context.Background()
+
+		for pb.Next() {
+			parent := ""
+
+			for l := 0; l < layerCount; l++ {
+				current := fmt.Sprintf("prepare-layer-%d", atomic.AddInt64(&layerIndex, 1))
+
+				mounts, err := snapshotter.Prepare(ctx, current, parent)
```
samuelkarp: Can you make sub-benchmarks for each of these steps?  The documentation for `b.RunParallel` says not to use `b.Run` inside (and there's no `pb.Run` as far as I can tell), but I'd like to get separate information for:

* how long it takes to prepare a snapshot (device activation?)
* how long it takes to write into the snapshot
* how long it takes to commit the snapshot

snapshotter/benchmark_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+				parent = fmt.Sprintf("comitted-%d", atomic.AddInt64(&layerIndex, 1))
+				err = snapshotter.Commit(ctx, parent, current)
+				require.NoError(b, err)
+			}
+		}
+	})
+}
+
+func makeApplier(layerIndex int, fileSizeBytes int64) []fstest.Applier {
```
samuelkarp: Some doc on this function would be good.  Looks to me like:

> makeApplier returns a slice of fstest.Applier where files are written randomly.  When more than one layer is specified, the returned layers will overwrite some files with the same generated names with new contents or deletions.

The only thing that I might want to see in addition to how this file functions is if you can change some of the `fstest.CreateRandomFile` calls to some applier that *modifies* a few bytes of the file without replacing it (maybe add one byte to the end or truncate a byte, or modify a byte somewhere in the middle).  That should help demonstrate the difference in performance for block-based snapshotters (like devicemapper) against file-based snapshotters (like overlay, which need to perform a copy-up of the full file any time a single bit is modified).

mxpv:
@samuelkarp done!

Issue #76, opened 2019-01-28
Allow specification of VM parameters


samuelkarp:
Many of the configuration parameters we support today should be settable at the VM level instead of using the same for all VMs.

* CPU count
* CPU template
* Memory
* Additional storage devices
* Network devices
* Kernel

aaithal:
Resolving as #109 is merged.

PR #105, opened 2019-02-11
Pass Firecracker VM opts

aaithal:
*Issue #, if available:* #76

*Description of changes:* Use CreateTaskRequest's Options field to pass firecracker config args

*Testing done:* Tested with the `taskworkflow.go` code attached in the PR. Output from execution:

```bash
2019/02/11 21:02:48.813898 Creating contained client
2019/02/11 21:02:48.814324 Created contained client
2019/02/11 21:02:48.981848 Successfully pulled docker.io/library/nginx:latest image
2019/02/11 21:02:51.202941 Successfully created task: demo for the container
2019/02/11 21:02:51.202962 Completed waiting for the container task
2019/02/11 21:02:51.231901 Successfully started the container task
2019/02/11 21:02:54.232051 Executing http GET on 172.16.0.2
2019/02/11 21:02:54.233193 Response from [172.16.0.2]:
[<!DOCTYPE html>
<html>
<head>
<style>
    body {
        width: 35em;
        margin: 0 auto;
    }
</style>
</head>
<body>
<h1>Welcome to nginx!</h1>

</body>
</html>
]
172.16.0.1 - - [11/Feb/2019:21:02:54 +0000] "GET / HTTP/1.1" 200 612 "-" "Go-http-client/1.1" "-"
```


examples/taskworkflow.go
```diff
@@ -0,0 +1,155 @@
...
+	var ip = flag.String("ip", "", "ip address assigned to the container")
+	log.SetFlags(log.Ldate | log.Ltime | log.Lmicroseconds)
+	flag.Parse()
+	if err := taskWorkflow(*ip); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func taskWorkflow(containerIP string) error {
+	log.Println("Creating contained client")
```
samuelkarp: nit: s/contained/containerd/

examples/taskworkflow.go
```diff
@@ -0,0 +1,155 @@
...
+	client, err := containerd.New("/run/containerd/containerd.sock")
+	if err != nil {
+		return errors.Wrapf(err, "creating client")
+
+	}
+	defer client.Close()
+	log.Println("Created contained client")
```
samuelkarp: nit: s/contained/containerd/

examples/taskworkflow.go
```diff
@@ -0,0 +1,155 @@
...
+	code, _, err := status.Result()
+	if err != nil {
+		return errors.Wrapf(err, "getting task's exit code")
+
+	}
+	log.Printf("task exited with status: %d\n", code)
+	return nil
+}
+
+func withFirecrackerConfig(containerIP string) containerd.NewTaskOpts {
```
samuelkarp: Should this be an exported functional option somewhere?  Maybe one that takes the MAC and HostDevName as arguments as well?

examples/taskworkflow.md
```diff
@@ -0,0 +1,87 @@
+### About
+The `taskworkfow.go` file contains code to pull and unpack a container image
+using the `naive` snaphotter, create a container and execute a task using the
+`firecracker-containerd` runtime.
+
+### Building
+After checking out the `firecracker-containerd` repo, you can build this
+example using the `go build` command. Example:
+```bash
+$ go build -o /tmp/taskworkflow examples/taskworkflow.go
```
samuelkarp: Can you add this to the Makefile as well?  I think `make examples` would be good.  Then we can include compiling the examples in our CI so they don't break.

examples/taskworkflow.md
```diff
@@ -0,0 +1,87 @@
...
+For a basic workflow, without any networking setup, just run the executable
+built with the command in the previous section as super user:
+```bash
+$ sudo /tmp/taskworkflow
+```
+
+For a workflow with networking setup for the container, create a tap device
+for the VM by following [these instructions](https://github.com/firecracker-microvm/firecracker/blob/master/docs/network-setup.md).
+
+This will create a tap device named `tap0`. Now, edit the kernel args in the
```
samuelkarp: Wow we really need to fix this problem.  https://github.com/firecracker-microvm/firecracker-containerd/issues/76 is the tracking issue.

proto/types.proto
```diff
@@ -11,3 +11,15 @@ message ExtraData {
...
+// Message to specify network config for a Firecracker VM
+message FirecrackerNetworkConfig {
+	string MAC = 1;
+	string HostDevName = 2;
+}
+
+// Message to specify the various Firecracker VM configurations for a task
+message FirecrackerConfig {
+	google.protobuf.Any RuncOptions = 1; // Can be used by the CreateTask request needs to relay any Runc options to the agent
+	FirecrackerNetworkConfig NetworkConfig = 2; // Can be used to set the networking configuration for a VM
```
samuelkarp: We might want this to be `repeated` so that more than one device can be configured.

runtime/service.go
```diff
@@ -666,6 +699,22 @@ func (s *service) stopVM() error {
...
 
+func parseCreateTaskOpts(ctx context.Context, opts *ptypes.Any) (*proto.FirecrackerConfig, error) {
+	log.G(ctx).WithField("opts", opts).Debug("unmarshaling create task create req options")
+
+	cfg, err := typeurl.UnmarshalAny(opts)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unmarshaling task create request options")
+	}
+
+	firecrackerConfig, ok := cfg.(*proto.FirecrackerConfig)
```
samuelkarp: I think we should also expect runc options here.  I'd like to avoid requiring client libraries to use the `proto.FirecrackerConfig` type if they want to use standard runc options and don't have anything Firecracker-specific to add.

examples/taskworkflow.go
aaithal: May be. But, I don't feel the do it right now since we'd need to hardcode the n/w interface name anyway, unless we accept it as a flag for the program. It's probably fine the way it is for this example. wdyt?

examples/taskworkflow.md
aaithal: It's a bit circular. If y'all are fine with this PR, it's going to turn into a very simple matter of adding more fields in the protobuf and processing unmarshalled fields :)

proto/types.proto
aaithal: Yeah, I had it that way initially. Somehow missed that here. I'll modify it to handle multiple interfaces. Good catch!

runtime/service.go
aaithal: Hmm, so essentially handle  two types here? I can work with that. I'll modify to handle that scenario.

examples/taskworkflow.go
samuelkarp: Sounds good.

examples/taskworkflow.md
samuelkarp: Yep!

examples/taskworkflow.md
```diff
@@ -0,0 +1,84 @@
...
+firecracker-containerd config to set a static ip address from the
+`172.16.0.1/24` subnet. The following example sets it to `172.16.0.2`.
+** NOTE: This example will not work if you're running more than 1 container
+on a host at the same time **
+```bash
+$ cat /etc/containerd/firecracker-runtime.json
+{
+    "firecracker_binary_path": "/home/ec2-user/firecracker-wrapper.sh",
```
nmeyerhans: Is `firecracker-wraper.sh` something that should have been included in this PR?

aaithal: Nope, should be fixed now!

proto/types.proto
```diff
@@ -11,3 +11,15 @@ message ExtraData {
 	bytes JsonSpec = 1;
 	google.protobuf.Any RuncOptions = 2;
 }
+
+// Message to specify network config for a Firecracker VM
+message FirecrackerNetworkConfig {
```
samuelkarp: I think I'd prefer that this was `FirecrackerNetworkInterface` to match the `firecracker.NetworkInterfaces` type.

proto/types.proto
```diff
@@ -11,3 +11,15 @@ message ExtraData {
...
+	string MAC = 1;
+	string HostDevName = 2;
+}
+
+// Message to specify the various Firecracker VM configurations for a task
+message FirecrackerConfig {
+	google.protobuf.Any RuncOptions = 1; // Can be used by the CreateTask request needs to relay any Runc options to the agent
+	repeated FirecrackerNetworkConfig NetworkConfig = 2; // Can be used to set the networking configuration for a VM
```
samuelkarp: I think I'd prefer that this was `NetworkInterfaces` to match the `Config.NetworkInterfaces` field.

proto/types.proto
aaithal: good suggestion. +1 for having same names everywhere.

aaithal: Should be fixed now, can you take another look?

proto/types.proto
```diff
@@ -11,3 +11,15 @@ message ExtraData {
 	bytes JsonSpec = 1;
 	google.protobuf.Any RuncOptions = 2;
 }
+
+// Message to specify network config for a Firecracker VM
+message FirecrackerNetworkInterface {
```
nmeyerhans: Should we include addition IPv4 configuration fields in this message? In your example, you provide IPv4 configuration to the kernel via a hardcoded commandline (`ip=...`). If we want to provide this information dynamically, this message seems to be the place to do it. If you'd rather save that for a later PR, though, that's fine.

aaithal: Yeah, my intention is to do that in a later PR as we'll be adding additional fields anyway as per #76. We can then get rid of that manual override in the example as well.

PR #106, opened 2019-02-18
Snapshotter improvements

mxpv:
*Description of changes:*
Address a few issues I got in `containerd` [PR](https://github.com/containerd/containerd/pull/3022) review:
- Resume base thin-device when snapshot creation failed
- Better check whether thin-device is active (all thin devices gets deactivated on reboot, so we can't rely on metadata)
- Snapshotter no longer creates or reloads thin-pool, just opens it or fails.
- Rollback thin device on error (#107)

/cc @sipsma

### DIFF
--- a/snapshotter/devmapper/pool_device.go
+++ b/snapshotter/devmapper/pool_device.go
@@ -15,7 +15,6 @@ package devmapper
 
 import (
 	"context"
-	"os"
 	"path/filepath"
 
 	"github.com/containerd/containerd/log"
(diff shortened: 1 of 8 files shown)


snapshotter/devmapper/pool_device.go
```diff
@@ -264,22 +305,34 @@ func (p *PoolDevice) DeactivateDevice(ctx context.Context, deviceName string, de
...
 
+// IsActivated returns true if thin-device is activated and not suspended
+func (p *PoolDevice) IsActivated(deviceName string) bool {
+	infos, err := dmsetup.Info(deviceName)
+	if err != nil || len(infos) == 0 {
+		// Couldn't query device info, device not active
+		return false
+	}
+
+	if devInfo := infos[0]; devInfo.Suspended {
```
xibz: Will `infos` length ever be not 1? If it is suspected to be 1, then we should return an error if `len(infos) != 1`. If the length can be any number, we should change this to be a for loop.

mxpv: Len can be either 0 (if no device) or 1, since it's not possible to have 2 devices with same name.

xibz: So if it isn't 0, which it would be since we got to this line, we need to check that `len` equals to 1. If it does not, return an error. I just don't want to make any assumptions that it will always be 0 or 1 and at least return an error for unexpected behavior.

mxpv: Fixed

snapshotter/devmapper/pool_device.go
```diff
@@ -148,27 +105,71 @@ func (p *PoolDevice) transition(ctx context.Context, deviceName string, tryingSt
...
 		Size:  virtualSizeBytes,
 		State: Unknown,
 	}
 
+	var (
+		metadataAdded bool
+		deviceCreated bool
+	)
+
+	defer func() {
```
samuelkarp: I think this might be more clear?

```go
if err := p.metadata.AddDevice(ctx, info); err != nil {
	return errors.Wrapf(err, "failed to save initial metadata for new thin device %q", deviceName)
}
defer func() {
	if retErr == nil {
		return
	}
	// rollback metadata
	retErr = multierror.Append(result, p.metadata.RemoveDevice(ctx, info.Name)
}()

if err := p.createDevice(ctx, info); err != nil {
	return err
}
defer func() {
	if retErr == nil {
		return
	}
	// rollback creation
	retErr = multierror.Append(result, p.deleteDevice(ctx, info))
}()

return p.activateDevice(ctx, info)
```

This lets the rollback case be coupled with the provisioning case (add and remove are together, create and delete are together) and lets `defer` ordering take care of executing the right code at the right time.  If `p.activateDevice` fails, `p.deleteDevice` will run first followed by `p.metadata.RemoveDevice`.  Similarly, if `p.createDevice` fails, only `p.metadata.RemoveDevice` will run.

The only gross thing here is repeated checking of `retErr` being non-nil, but I'm not sure how to avoid that.

What do you think?

snapshotter/devmapper/pool_device.go
```diff
@@ -202,52 +200,95 @@ func (p *PoolDevice) CreateSnapshotDevice(ctx context.Context, deviceName string
 		State:      Unknown,
 	}
 
+	var (
+		metadataAdded bool
+		snapCreated   bool
+	)
+
+	defer func() {
```
samuelkarp: Same pattern as above can be applied here.

snapshotter/devmapper/pool_device.go
mxpv: I couldn't decide which approach is better, so the suggested one works for me :)

mxpv:
> What changes for test setup? Do tests still use a loop device and take care of their own setup?

We still use sparse files, but added `dmsetup.Create` (see [here](https://github.com/firecracker-microvm/firecracker-containerd/pull/106/files#diff-8ac264ffe2e041eb8b322c927cf6c685R68))

samuelkarp:
Can you rebase this on master so the CI runs?  After CI passes, the code looks good to me.

PR #108, opened 2019-02-19
Better GOPATH handling in Makefile

mxpv:

*Description of changes:*
Small improvement in Makefile. `GOPATH` may not be defined as env variable (when default `~/go/` used). `go env GOPATH` handles this case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,8 @@ SUBDIRS:=agent runtime snapshotter
 export INSTALLROOT?=/usr/local
 export STATIC_AGENT
 
+GOPATH:=$(shell go env GOPATH)
+
 all: $(SUBDIRS)
 
 $(SUBDIRS):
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -15,6 +15,8 @@ SUBDIRS:=agent runtime snapshotter
 export INSTALLROOT?=/usr/local
 export STATIC_AGENT
 
+GOPATH:=$(shell go env GOPATH)
```
nmeyerhans: This should not be needed.

samuelkarp: :+1: I'd prefer to install this into a local `.gitignore`d directory instead of the user's `GOPATH`.

nmeyerhans: Never mind; I understand what's going on. We still need GOPATH to expand to `~/go` even if $GOPATH isn't explicitly set. I misunderstood your intent.

mxpv: We already install `git-validation` and `ltag` to user's `GOPATH`.
I don't think moving `golangci-lint` would help a lot.

samuelkarp: :frowning_face: I'd prefer not to install any of those into the user's `GOPATH`.

mxpv: Let's solve this separately. The intent of this PR was to fix GOPATH in one specific case.

samuelkarp: That's fine; we can table the discussion for now and move on.  But I'm going to continue to be unhappy about a Makefile target that mutates the contents of my GOPATH.

PR #109, opened 2019-02-21
More firecracker opts

aaithal:
*Issue #, if available:* #76

*Description of changes:* Added protobuf message types for all of the options listed in #76. The `taskworkflow` example has also been updated to get rid of the manual kernel args override step with these changes.

*Testing Done:* Added additional unit tests. Verified the the `taskworkflow` example executes without errors as well.

```bash
#### Verify that there's no kernel arg override for ip, gateway, netmask in `/etc/containerd/firecracker-runtime.json`
$ cat /etc/containerd/firecracker-runtime.json | grep kernel_args
  "kernel_args": "console=ttyS0 noapic reboot=k panic=1 pci=off nomodules rw",

#### Execute the example
$ sudo ./taskworkflow -ip 172.16.0.2 -gw 172.16.0.1 -mask 255.255.255.0
2019/02/21 16:42:57.390384 Creating containerd client
2019/02/21 16:42:57.391192 Created containerd client
2019/02/21 16:43:06.090317 Successfully pulled docker.io/library/nginx:latest image
2019/02/21 16:43:08.279340 Successfully created task: demo for the container
2019/02/21 16:43:08.279363 Completed waiting for the container task
2019/02/21 16:43:08.307723 Successfully started the container task
2019/02/21 16:43:11.307840 Executing http GET on 172.16.0.2
2019/02/21 16:43:11.309085 Response from [172.16.0.2]:
[<!DOCTYPE html>
<html>
<head>
<style>
    body {
        width: 35em;
        margin: 0 auto;
    }
</style>
</head>
<body>
<h1>Welcome to nginx!</h1>

</body>
</html>
]
172.16.0.1 - - [21/Feb/2019:16:43:11 +0000] "GET / HTTP/1.1" 200 612 "-" "Go-http-client/1.1" "-"
```


runtime/taskfirecrackeropts.go
```diff
@@ -0,0 +1,161 @@
...
+
+// machineConfigFromProto creates a firecracker MachineConfiguration object
+// from the protobuf FirecrackerMachineConfiguration message.
+func machineConfigFromProto(mCfg *proto.FirecrackerMachineConfiguration) (models.MachineConfiguration, error) {
+	result := models.MachineConfiguration{}
+	if mCfg.MemSizeMib > 0 && mCfg.VcpuCount == 0 {
+		// Only memory is being overridden.
+		return result, errors.New("both vm memory and vcpu count need to be overridden")
+	}
+	if mCfg.VcpuCount > 0 && mCfg.MemSizeMib == 0 {
```
mxpv: nit: Can these be just one if `MemSizeMib == 0 || VcpuCount == 0` ?

runtime/taskfirecrackeropts.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/firecracker-microvm/firecracker-containerd/proto"
+	firecracker "github.com/firecracker-microvm/firecracker-go-sdk"
+	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
+	"github.com/pkg/errors"
+)
+
+// overrideVMConfigFromTaskOpts overrides the firecracker VM config for the
+// task with values set in the protobuf message.
+func overrideVMConfigFromTaskOpts(
```
mxpv: I'm not sure if it's within scope of this PR, but I would rather throw away json configuration with giant override, and leave just protobuf stuff. So instead of `read json + build FC config + override protobuf` we would have `defaults + protobuf = FC Config`

runtime/taskfirecrackeropts.go
aaithal: So, here we're verifying that if `vcpuCount` has been overridden, `memSizeMib` is overridden as well. Technically, you can have a machine config where you override neither. `MemSizeMib == 0 || VcpuCount == 0` would not work in that scenario.

runtime/taskfirecrackeropts.go
aaithal: That's an interesting take. I'll open an issue and we can discuss that there?

mxpv: Yep :+1:

aaithal: @mxpv I took another look at the code.The JSON parsing that's happening seems to be build the default config. So, I'm not sure if that's relevant to this PR. IIUC, that's more of a question for creation of `service.config` is initialized. Was that your concern with the initial comment?

mxpv: Originally JSON was added as temp solution to simplify running. Here we have to manage two entities for configuration and keep `overrideVMConfigFromTaskOpts` to overwrite config params. This makes it more complicated than it should be and I see no reasons to keep old config. 
Anyway it's not a blocker for this PR, just concern :)

aaithal: Thanks for clarifying. Created #111 to keep track of that.

samuelkarp:
@aaithal We just added CI to this repo.  Would you be able to rebase on top of master, and then add an entry to the .buildkite/pipeline.yml similar to https://github.com/firecracker-microvm/firecracker-containerd/pull/113 to run the new tests you're adding?

examples/taskworkflow.go
```diff
@@ -80,8 +92,24 @@ func taskWorkflow(containerIP string) error {
 
 	task, err := container.NewTask(ctx,
 		cio.NewCreator(cio.WithStdio),
-		withFirecrackerConfig(containerIP, "AA:FC:00:00:00:01", "tap0"),
-	)
+		func(ctx context.Context, _ *containerd.Client, ti *containerd.TaskInfo) error {
```
samuelkarp: How come you put this inline instead of leaving it as a separate function?

proto/types.proto
```diff
@@ -12,14 +12,59 @@ message ExtraData {
 	google.protobuf.Any RuncOptions = 2;
 }
 
+// Message to specify the various Firecracker VM configurations for a task
+message FirecrackerConfig {
+	google.protobuf.Any RuncOptions = 1; // Specifies Runc options that need to relayed to the CreateTask request
+	repeated FirecrackerNetworkInterface NetworkInterfaces = 2; // Specifies the networking configuration for a VM
+	string KernelImagePath = 3; // Specifies the file path where the kernel image is located
+	string KernelArgs = 4; // Specifies the commandline arguments that should be passed to the kernel
```
samuelkarp: Should this be `repeated string`?  (I haven't thought about this in depth)

proto/types.proto
```diff
@@ -12,14 +12,59 @@ message ExtraData {
...
+	 // for a mircroVM should be large enough
+	uint32 MemSizeMib = 3;
+	uint32 VcpuCount = 4; // Specifies the number of vCPUs for the VM
+}
+
+// Message to specify the block device config for a Firecracker VM
+message FirecrackerDrive {
+	bool IsReadOnly = 1; // Specifies if the drive is read only
+	bool IsRootDevice = 2; // Specifies if the drive is the root device
+	string Partuuid = 3; // Specifies the unique id of the boot partition on this device
```
samuelkarp: `PartUUID`?

proto/types.proto
nmeyerhans: We may at some point end up wanting a more featureful kernel commandline constructor. But I don't really want to go there if we don't have to. IMO leaving it as a single opaque string is probably going to avoid the most madness.

proto/types.proto
aaithal: It's reflecting the name in https://github.com/firecracker-microvm/firecracker-go-sdk/blob/4b31e1c2d796c8a866fd5047ee689dfb03754a5f/client/models/drive.go#L46. I'm in favor of keeping it as is for the sake of consistency.

examples/taskworkflow.go
aaithal: I didn't really feel the need for a separate function for this as it'd just be passing a number of parameters that are already here everywhere. Will revert to a named method if there are more complicated use-cases to consider.

aaithal:
> @aaithal We just added CI to this repo. Would you be able to rebase on top of master, and then add an entry to the .buildkite/pipeline.yml similar to #113 to run the new tests you're adding?

Yeah, looking into doing that!

PR #113, opened 2019-02-21
runtime: add make target for tests

samuelkarp:
*Description of changes:*

New Makefile target, new buildkite step

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,4 +1,4 @@
-# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2018-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"). You may
 # not use this file except in compliance with the License. A copy of the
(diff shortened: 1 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -31,3 +31,8 @@ steps:
     commands:
       - "cd snapshotter"
       - "sudo make test EXTRAGOARGS='-v -count=1'"
+
+  - label: ':running_shirt_with_sash: runtime tests'
+    commands:
+      - "cd runtime"
```
mxpv: Why not to run all tests from root folder? So we don't need to update this if/when we add any other packages with tests (for example `agent` is going to have tests eventually)

.buildkite/pipeline.yml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2018-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mxpv: It looks like this is not forced by linter :(

.buildkite/pipeline.yml
samuelkarp: Summarizing our offline discussion: separate steps in Buildkite run in parallel, helping to increase utilization of our build fleet and decrease the test execution time.  Separate steps also gives granular status on the Buildkite status page, so it's easier to dive in and find where a test is failing.

.buildkite/pipeline.yml
samuelkarp: Yeah :slightly_frowning_face: I don't know how that linter works.

Issue #114, opened 2019-02-21
Run `example/taskworkflow.go` as a test


aaithal:
This example app provides us a way to run end-to-end tests. It'd be awesome if that was something that the CI can execute.

sipsma:
Closed via #227

PR #116, opened 2019-02-22
Set the agent shim as a subreaper to ensure containers become its children

### LINKED ISSUE #87: killing task doesn't work properly
the kill task doesn't work properly and causes shim zombie issue 
pls let me know if you need any extra info
```
ctr c create --snapshotter firecracker-dm-snapshotter --runtime aws.firecracker docker.io/library/redis:latest myredis1
ctr tasks start myredis1
```
```
1:C 06 Feb 2019 19:11:45.423 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
1:C 06 Feb 2019 19:11:45.423 # Redis version=5.0.3, bits=64, commit=00000000, modified=0, pid=1, just started
1:C 06 Feb 2019 19:11:45.423 # Warning: no config file specified, using the default config. In order to specify a config file use redis-server /path/to/redis.conf
1:M 06 Feb 2019 19:11:45.425 # You requested maxclients of 10000 requiring at least 10032 max file descriptors.
1:M 06 Feb 2019 19:11:45.426 # Server can't set maximum open files to 10032 because of OS error: Operation not permitted.
1:M 06 Feb 2019 19:11:45.426 # Current maximum open files is 1024. maxclients has been reduced to 992 to compensate for low ulimit. If you need higher maxclients increase 'ulimit -n'.
1:M 06 Feb 2019 19:11:45.428 * Running mode=standalone, port=6379.
1:M 06 Feb 2019 19:11:45.428 # WARNING: The TCP backlog setting of 511 cannot be enforced because /proc/sys/net/core/somaxconn is set to the lower value of 128.
1:M 06 Feb 2019 19:11:45.428 # Server initialized
1:M 06 Feb 2019 19:11:45.428 # WARNING overcommit_memory is set to 0! Background save may fail under low memory condition. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
1:M 06 Feb 2019 19:11:45.429 * Ready to accept connections
```
```
ctr tasks list
TASK        PID    STATUS    
myredis1    996    RUNNING
ctr tasks kill myredis1
```
```
1:signal-handler (1549480354) Received SIGTERM scheduling shutdown...
ctr: rpc error: code = Unknown desc = ttrpc: client shutting down: read vsock:host(2):1028: connection reset by peer: unknown
```
```
ps axu | grep shim
root      4016  0.4  0.0 784440 13444 pts/0    Sl   19:11   0:01 /usr/local/bin/containerd-shim-aws-firecracker -namespace default -address /run/containerd/containerd.sock -publish-binary /usr/local/bin/containerd
```
```
ctr tasks ls
ERRO[2019-02-06T19:17:25.961348016Z] converting task to protobuf                   error="ttrpc: client shutting down: read vsock:host(2):1028: connection reset by peer: unknown" id=myredis1
TASK    PID    STATUS    
```

sipsma:
*Description of changes:*
Without this, the final container spawned by RunC will be reparented to init, not the shim, and the shim won't receive SIGCHLD upon container death.

#87 is not fixed by this commit, but it's a first step for doing so by ensuring the shim is actually aware of child process deaths.

*Testing done:*
Ran local `go test` and the cool new [CI tests passed](https://buildkite.com/firecracker-microvm/sipsma-firecracker-containerd/builds/1#ebca8ef9-4bc5-4813-9937-076069713561)

Also manually verified from a shell in the VM that previously child process containers were not the children of agent and that now they are (the `agent---sh---pstree` branch):
```
/ # pstree
openrc-init-+-agent---sh---pstree
            |-sh---sleep
            `-supervise-daemo---agetty
```

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -21,10 +21,12 @@ import (
 	"os/signal"
 	"syscall"
 
+	"github.com/containerd/containerd/events"
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/runtime/v2/runc"
 	"github.com/containerd/containerd/runtime/v2/shim"
 	shimapi "github.com/containerd/containerd/runtime/v2/task"
+	"github.com/containerd/containerd/sys"
 	"github.com/containerd/ttrpc"
 	"github.com/mdlayher/vsock"
 	"github.com/sirupsen/logrus"
@@ -32,7 +34,23 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-const defaultPort = 10789
+const (
+	defaultPort = 10789
+
+	// per prctl(2), we must provide a non-zero arg when calling prctl with
+	// PR_SET_CHILD_SUBREAPER in order to enable subreaping (0 disables it)
+	enableSubreaper = 1
+)
+
+// A fake publisher that we use as a bandaid to prevent nil pointer exceptions when the
+// runc shim tries to publish events.
+// TODO replace this with a publisher that forwards events over vsock to the outer shim
+// on the host.
+type noopPublisher struct{}
+
+func (p *noopPublisher) Publish(ctx context.Context, topic string, event events.Event) error {
+	return nil
+}
 
 func main() {
 	var (
@@ -63,13 +81,19 @@ func main() {
 		log.G(ctx).WithError(errors.New("invalid argument")).Fatal("id not set")
 	}
 
+	// Ensure this process is a subreaper or else containers created via runc will
+	// not be its children.
+	if err := sys.SetSubreaper(enableSubreaper); err != nil {
+		log.G(ctx).WithError(err).Fatal("failed to set shim as subreaper")
+	}
+
 	// Create a runc task service that can be used via GRPC.
 	// This can be wrapped to add missing functionality (like
 	// running multiple containers inside one Firecracker VM)
 
 	log.G(ctx).WithField("id", id).Info("creating runc shim")
 
-	runcTaskService, err := runc.New(ctx, id, nil)
+	runcTaskService, err := runc.New(ctx, id, &noopPublisher{})
 	if err != nil {
 		log.G(ctx).WithError(err).Fatal("failed to create runc shim")
 	}


samuelkarp:
> Ran local `go test` and the cool new [CI tests passed](https://buildkite.com/firecracker-microvm/sipsma-firecracker-containerd/builds/1#ebca8ef9-4bc5-4813-9937-076069713561)

Awesome!!

> Without this, the final container spawned by RunC will be reparented to init, not the shim, and the shim won't receive SIGCHLD upon container death.

How does this play with PID namespaces?

From pid_namespaces(7):

> A  child process that is orphaned within the namespace will be reparented to this process rather than init(1) (unless  one  of  the ancestors  of the child in the same PID namespace employed the prctl(2) PR_SET_CHILD_SUBREAPER command to mark itself as the reaper of orphaned descendant processes).

I don't feel like I understand the cases where the container's PID 1 is parented to init; is that because the runc process exits and the agent doesn't inherit the child except if subreaper is set?  Similarly, is this different for additional processes executed in the PID namespace (`ctr exec`) or for containers without a PID namespace?

sipsma:
**Short answer:**

Yes, the issue is that when runc exits, its child process is the container and will be reparented to init unless agent has set itself as a subreaper. The issue and solution shouldn't be affected by PID namespacing being enabled/disabled and also shouldn't be affected by whether the process is the init of the container or a later exec.

**Longer answer:**

The problem occurs because agent calls runc as a child process and runc spawns the actual container as a child process of its own. Thus, the view from the host's root PID namespace when runc hasn't quite exited yet is `init---agent---runc---container`. 

Then, when container initialization is finished, runc exits and `container` needs to be reparented in the actual pid namespace. The question, as I understand it, is how to interpret the pid_namespaces doc you mentioned above in light of the fact that `container` might actually be the init of its own pid namespace. It can't be reparented to itself obviously.

This is one of those rare times where looking at kernel code has actually been pretty clarifying; it turns out [the reparenting of a process occurs in the context of the exiting parent](https://github.com/torvalds/linux/blob/12491ed354d23c0ecbe02459bf4be58b8c772bc8/kernel/exit.c#L597). So, when `runc` exits, the kernel is actually determining who to reparent `container` to in the context of `runc`, not the context of `container`. 

`runc` is not in the pid ns of `container`, so the procedure for determining who to reparent to (either the most recent subreaper or pid 1 of the pid ns, whoever is first) works as described but in the context of the *host's root pid ns*. Thus, when `runc` exits, `container` will by default be reparented to `init` unless `agent` has marked itself as a subreaper, in which case it will adopt `container`.

If PID namespaces are not enabled, `runc` will still just be in the host's root pid ns and thus the end effect will be the same as if `container` had been in its own pid ns.

For a process that was exec'd into a container with a pid ns, I believe it will actually follow the exact same procedure as for the init process. Then, according to pid_namespaces(7):
> Likewise, the direct children of a process that uses setns(2) to cause its children to join a PID namespace are in a different PID namespace from the caller of setns(2).  Calls to getppid(2) for such processes return 0.

So inside the container, they see their parent as 0, but outside the container they again will be reparented to the host pid ns init or `agent` if it explicitly sets itself as a subreaper.

sipsma:
@nmeyerhans noticed that, with this change in isolation, when you exit a container shell inside the vm the terminal is not released back to the original shell outside the vm on the host, which I was also able to reproduce.

I believe this is because even though the change here fixes one issue, it actually also enables another bug that had been lying dormant. What happens now is:
1. The child dies, [agent actually notices now via SIGCHLD and calls the reaper](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/agent/main.go#L110)
2. The RunC shim we're wrapping around gets notified and [tries to publish the task exited event](https://github.com/containerd/containerd/blob/85aa8ad3614ed183f6db8564140cdc876d2facee/runtime/v2/runc/service.go#L792)
3. However, it gets a nil pointer exception when doing so because [the agent provides nil for the "publisher" provided to runc](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/agent/main.go#L72)

I made a local change that just provides a "noop publisher" instead of nil that just doesn't do anything when called, that brought the behavior back to baseline. It still does not alleviate #87 but allows us to move towards fixing #87 while not regressing.

I'd propose that for now we make the change here plus the noop publisher (I'll update the PR) and then address the rest of #87 in subsequent commits. I think the best path to resolving it will be to replace the noop publisher with one that forwards events over vsock and then update the outer shim to just wait for task exited events before shutting the VM and itself down (thus allowing everything to shutdown gracefully).

xibz:
Out of curiosity, can we add a test for this to ensure the child processes are getting cleaned up?

sipsma:
> Out of curiosity, can we add a test for this to ensure the child processes are getting cleaned up?

For sure, I'll take a look at what's needed to enable that

agent/main.go
```diff
@@ -34,6 +36,15 @@ import (
 
 const defaultPort = 10789
 
+// A fake publisher that we use as a bandaid to prevent nil pointer exceptions when the
+// runc shim tries to publish events.
+// TODO replace this with a publisher that forwards events over vsock to the outer shim
```
mxpv: Just wanted to add some context here: [originally](https://github.com/containerd/containerd/blob/26ab393e7ddf23d245244043ae4629a2a57dc453/runtime/v2/runc/service.go#L663) (before Jan 25, 2019) there was just one event published by `runc` - `TaskExit` event, so we decided to go with [pull model](https://github.com/firecracker-microvm/firecracker-containerd/blob/7f37fafbadb4789decd6bf34ae74283e189f0d62/runtime/service.go#L226) as it seemed like an overkill to implement a fully-featured publisher for an exit event.

There is a [commit](https://github.com/containerd/containerd/commit/85aa8ad3614ed183f6db8564140cdc876d2facee#diff-6238adea42c707e286831f2a8566d37d) which adds a lot more events to `runc`, that was added about one month ago. I think this is when we started getting `nil` errors.

agent/main.go
```diff
@@ -63,13 +74,19 @@ func main() {
 		log.G(ctx).WithError(errors.New("invalid argument")).Fatal("id not set")
 	}
 
+	// Ensure this process is a subreaper or else containers created via runc will
+	// not be its children.
+	if err := sys.SetSubreaper(1); err != nil {
```
mxpv: Can you use constant for `1` to make it clearer what exactly is it.

agent/main.go
sipsma: Ah interesting, good to know!

agent/main.go
sipsma: Yeah, will do

samuelkarp:
>> Likewise, the direct children of a process that uses setns(2) to cause its children to join a PID namespace are in a different PID namespace from the caller of setns(2). Calls to getppid(2) for such processes return 0.
>
> So inside the container, they see their parent as 0, but outside the container they again will be reparented to the host pid ns init or agent if it explicitly sets itself as a subreaper.

This feels weird to me, but it does seem to be backed up by the implementation in the kernel:

> When we die, we re-parent all our children, and try to:
> 1. give them to another thread in our thread group, if such a member exists
> 2. give it to the first ancestor process which prctl'd itself as a
>    child_subreaper for its children (like a service manager)
> 3. give it to the init process (PID 1) in our pid namespace

> I'd propose that for now we make the change here plus the noop publisher (I'll update the PR) and then address the rest of #87 in subsequent commits.

This plan sounds good.

Before I approve the PR, I'd like the failing Travis build to be addressed:

> golangci-lint run
> agent/main.go:43: File is not `goimports`-ed (goimports)
> type noopPublisher struct {}
> make: *** [lint] Error 1

sipsma:
Just a note, after looking into covering this with a test and talking with others, the best path will be for the taskworkflow example to be upgraded to a full blown test case (#114). I also think it will make the most sense for the test to be added as part of resolving #87, as right now the test would have to account for some things working and some things sort of not.

Issue #24, opened 2018-11-28
Implement a publisher to publish events from agent -> runtime


IRCody:
When the runc shim is created [here](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/agent/main.go#L66) we are passing a nil publisher. Instead we should have an implementation of the publisher that safely publishes events outside the vm back to containerd.

sipsma:
Closed via #118

PR #118, opened 2019-02-28
Forward agent containerd events to shim and monitor task exits via event subscriptions

### LINKED ISSUE #87: killing task doesn't work properly

sipsma:
Fixes #87 

More details in full commit messages, but overall goal of the change is to forward all containerd events from the agent to the shim on the host and then use that to monitor task exits, at which time we can gracefully shut everything down.

Sending this out to get feedback on the implementation right away, but I won't merge until I've also written an integration test (which is likely going to entail addressing #114) and added unit tests for the new EventExchange code. That being said, the testing done so far has been manual. I verified that both killing a task (via `ctr kill`) and the task exiting on its own result in a task exit event being forwarded to and logged by containerd. The shim and VM exit as expected in both cases too.


sipsma:
I forgot to run the linter locally before pushing, will fix that before merging

exchangeBridge/exchange.go
```diff
@@ -0,0 +1,161 @@
...
+			} else {
+				return nil
+			}
+		default:
+		}
+
+		envelope, err := exchangeBridge.GetEvent(ctx)
+		if err != nil {
+			return err
+		}
```
mxpv: What happens if there are no new events published? Will it fail here with timeout and will stop pulling events?

exchangeBridge/exchange.go
```diff
@@ -0,0 +1,161 @@
...
+
+	select {
+	case receivedEnvelope := <-s.eventChan:
+		envelope = &eventapi.Envelope{
+			Timestamp: receivedEnvelope.Timestamp,
+			Namespace: receivedEnvelope.Namespace,
+			Topic:     receivedEnvelope.Topic,
+			Event:     receivedEnvelope.Event,
+		}
+	case err = <-s.errChan:
```
mxpv: should we return here?

exchangeBridge/exchange.go
```diff
@@ -0,0 +1,161 @@
...
+func AttachExchangeBridge(ctx context.Context, exchangeBridge ExchangeBridge, forwarder events.Forwarder) error {
+	for {
+		select {
+		case <-ctx.Done():
+			if ctxErr := ctx.Err(); ctxErr != context.Canceled {
+				return ctxErr
+			} else {
+				return nil
+			}
+		default:
```
mxpv: Do we need empty default here? Without `default` "select" statement just blocks until ctx.Done can proceed

exchangeBridge/exchange.go
sipsma: We don't need to as we will return the err below. The implementation here works as is but I agree it's kind of a different approach than what I did elsewhere, so I can update this to just return in each select case for the sake of consistency.

exchangeBridge/exchange.go
sipsma: We want the empty default so we don't get blocked on the call to ctx.Done(). The select here is just a check to see if the context was canceled already, at which point we can return from this function. If the context wasn't canceled, we want to continue on with the loop.

exchangeBridge/exchange.go
sipsma: Looking through the TTRPC codebase I wasn't able to find any implementation of timeouts, so as far as I can tell the request will just block indefinitely until events are available or the connection is closed by either side. I also manually let a container sit around for 10 minutes and verified the task exit event was still propagated as expected afterwards (admittedly not the most robust test, but not sure how else to verify at this moment).

Let me know if I'm missing something.

exchangeBridge/exchange.go
mxpv: Ah. I get it the idea. I thought this is a cycle :( I would rather suggest something like this instead:
```
if _, ok := <- ctx.Done(); ok {
			
}
```

exchangeBridge/exchange.go
mxpv: Yep. Timeouts are not yet implemented https://github.com/containerd/ttrpc/issues/3

exchangeBridge/exchange.go
mxpv: I prefer to return as early as possible. This makes reading code a bit easier.

exchangeBridge/exchange.go
sipsma: I completely forgot you could do that to implement a non-blocking read from a channel. Agree it's more readable, will update the PR

exchangeBridge/exchange.go
sipsma: Thanks, that's good to know! Hopefully they add support for timeouts in a backwards-compatible way, but I'll assume they won't and see what I can do to cover this when adding tests (maybe just have a container sit around for 30ish seconds before existing or something like that).

exchangeBridge/exchange.go
sipsma: Apparently that syntax actually is still blocking: https://play.golang.org/p/ccsYmAn4BTf

So I'll leave it as is but add a command above the select clarifying what it's doing

exchangeBridge/exchange.go
```diff
@@ -0,0 +1,161 @@
...
+	ExchangeBridgeServiceName = "aws.firecracker.containerd.exchangeBridge"
+	GetEventMethodName = "GetEvent"
+)
+
+// The ExchangeBridge interface provides an API for retrieving containerd events from a remote event exchange. It exists
+// separately from containerd's existing EventClient and EventServer interfaces in order to support retrieving events
+// via long-polling as opposed to a streaming model. This allows us to use TTRPC for the implementation, which currently
+// does not support streaming.
+type ExchangeBridge interface {
+	GetEvent(ctx context.Context) (*eventapi.Envelope, error)
```
samuelkarp: Can you add godoc for `GetEvent`?

exchangeBridge/exchange.go
```diff
@@ -0,0 +1,161 @@
...
+// in the subscription's channel of published events.
+func NewExchangeBridgeService(ctx context.Context, eventSource events.Subscriber) ExchangeBridge {
+	eventChan, errChan := eventSource.Subscribe(ctx)
+	return &exchangeBridgeService{
+		eventChan: eventChan,
+		errChan:   errChan,
+	}
+}
+
+func RegisterExchangeBridgeService(srv *ttrpc.Server, svc ExchangeBridge) {
```
samuelkarp: This needs some godoc as well.

exchangeBridge/exchange.go
samuelkarp: I'd encourage returning at the site of an error.  It's easy to miss that the error is supposed to be returned when we come back and edit this code later if it's only returned at the bottom of the function.  I'd also encourage returning `nil, err` so that we explicitly have no `eventapi.Envelope` instead of one that may be invalid.

exchangeBridge/exchange.go
```diff
@@ -0,0 +1,161 @@
+package exchangeBridge
```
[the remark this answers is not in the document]
xibz: And further, the folder name shouldn't be camel cased either. I think calling the folder `exchange` and the package `exchange` would be my preference.

exchangeBridge/exchange.go
```diff
@@ -0,0 +1,161 @@
...
+	rpcClient *ttrpc.Client
+}
+
+func NewExchangeBridgeClient(rpcClient *ttrpc.Client) ExchangeBridge {
+	return &exchangeBridgeClient{
+		rpcClient: rpcClient,
+	}
+}
+
+func (c *exchangeBridgeClient) GetEvent(ctx context.Context) (*eventapi.Envelope, error) {
```
xibz: Can we get a description on how this method work in getting an event?

exchangeBridge/exchange.go
sipsma: This one I felt at the time was self-explanatory, but in retrospect it doesn't hurt to explain that it must be called to add the method to the provided server, will do.

exchangeBridge/exchange.go
sipsma: Yeah that makes total sense, will update

exchangeBridge/exchange.go
```diff
@@ -0,0 +1,161 @@
...
+const (
+	ExchangeBridgeServiceName = "aws.firecracker.containerd.exchangeBridge"
+	GetEventMethodName = "GetEvent"
+)
+
+// The ExchangeBridge interface provides an API for retrieving containerd events from a remote event exchange. It exists
+// separately from containerd's existing EventClient and EventServer interfaces in order to support retrieving events
+// via long-polling as opposed to a streaming model. This allows us to use TTRPC for the implementation, which currently
+// does not support streaming.
+type ExchangeBridge interface {
```
xibz: I am curious if it makes some sense to expose `ExchangeBridgeService` and pass in a `EventListener` during creation. Need a better name I think for the `EventListener.

```go
type EventListener interface {
    GetEvent(ctx context.Context) (*eventapi.Envelope, error)
}

type TTRPCEventListener struct {
    // stuff
}

func NewExchangeBridgeService(ctx context.Context, eventSource events.Subscriber, e EventListener) {
}
```

Then you can have a `TTRPCEventListener`. This allows reuse of the of some of the code and allows switching out of how the events are retrieved. What are your thoughts on that?

exchangeBridge/exchange.go
xibz: And can you move this down after the method. I tend to group methods with the structure it is tied to.

sipsma: Sure, will do

exchangeBridge/exchange.go
sipsma: Sure, will do

sipsma:
> Can we also get some tests to ensure this works properly?

Yes, mentioned this in the PR description but I don't plan on merging this until I've gotten an integration test up and running, which may take a few days. Just getting any feedback on the implementation out of the way first.

agent/main.go
```diff
@@ -71,7 +65,7 @@ func main() {
 	signals := make(chan os.Signal, 32)
 	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM, unix.SIGCHLD)
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(namespaces.WithNamespace(context.Background(), defaultNamespace))
```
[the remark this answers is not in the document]
sipsma: Created an issue for that here: https://github.com/firecracker-microvm/firecracker-containerd/issues/119

Let me know if the labeling is correctly following our conventions

exchangeBridge/exchange.go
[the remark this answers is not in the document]
sipsma: I ended up doing something else entirely. The package is "eventbridge" and I changed the service name to just "Getter" (reflecting that all it does is GetEvent over the eventbridge). I think this is the best compromise.

eventbridge/eventbridge.go
```diff
@@ -0,0 +1,173 @@
...
+			Topic:     receivedEnvelope.Topic,
+			Event:     receivedEnvelope.Event,
+		}, nil
+	case err := <-s.errChan:
+		return nil, err
+	case <-ctx.Done():
+		if ctxErr := ctx.Err(); ctxErr != context.Canceled {
+			return nil, ctxErr
+		}
+		return nil, nil
```
mxpv: I would rather return an error and make client responsible whether he wants to ignore `context.Canceled`.
When `GetEvent` returns `nil` error I expect that I can use envelope safely. Or at least this should be mentioned in godoc.

sipsma: Yeah I see your point, I went back and forth on this a bit. My original reasoning was that because the caller is providing context, they can be aware if it's canceled just as easily as the callee and thus shouldn't be forced to always check `if err != context.Canceled`. Err should just indicate that something went wrong with the callee's logic specifically.

However, you make a good a point that its pretty standard to assume that if err is nil you can go ahead and use the primary return value(s). Given that's a much more common assumption and doesn't change anything functionally, I'll switch to what you're suggesting.

eventbridge/eventbridge.go
```diff
@@ -0,0 +1,173 @@
...
+		select {
+		case <-ctx.Done():
+			if ctxErr := ctx.Err(); ctxErr != context.Canceled {
+				return ctxErr
+			}
+			return nil
+		default:
+		}
+
+		envelope, err := source.GetEvent(ctx)
```
xibz: I have a concern with this loop just spinning since it makes the assumption that `GetEvent` will handle not spinning this loop. I think another method of `HasEvent` or something of that nature needs to be added and should behave similarly to how `ctx.Done` works by returning a channel.

sipsma: The Godoc for getterClient.GetEvent mentions that it's blocking, I can add an additional short comment above the line your commenting on re-iterating it.

xibz: I think this needs to be added to the interface's method signature of `GetEvent`. There is no mention of that in the interface.

sipsma: Yeah I just noticed that too, I am adding a comment to the interface method too in addition to the implementations.

mxpv:
Ref https://github.com/firecracker-microvm/firecracker-containerd/issues/24

sipsma:
Updated w/ unit tests. I also have an e2e test written in my fork of the repo, will update this PR with it once I have merged the prerequisite #134

sipsma:
I updated w/ the e2e test. It's a pretty sizable change on top of the original so it's a separate commit.

The travis build is exiting with non-zero for reasons unknown to me at this time. When I run "make lint" on my own from the same commit it exits with 0.

sipsma:
The travis build was failing due to it only pulling 50 commits of history, which broke our git validation command. 

Sam fixed this, I will pick up that fix when I squash/rebase before merging. `make lint` passes locally though for these changes.

go.mod
```diff
@@ -1,8 +1,10 @@
 module github.com/firecracker-microvm/firecracker-containerd
 
 require (
+	github.com/BurntSushi/toml v0.3.1 // indirect
 	github.com/Microsoft/go-winio v0.4.11 // indirect
 	github.com/Microsoft/hcsshim v0.8.1 // indirect
+	github.com/StackExchange/wmi v0.0.0-20181212234831-e0a55b97c705 // indirect
```
samuelkarp: Is this being pulled in by containerd?

sipsma:
```
go mod why github.com/StackExchange/wmi      
                                                                                                                                                                                                                                                                                                                           
# github.com/StackExchange/wmi
github.com/firecracker-microvm/firecracker-containerd/internal
github.com/shirou/gopsutil/process
github.com/StackExchange/wmi
```

It appears to be the psutil library needed by the new service_integ_test package. That library also supports windows, so it makes sense I guess.

.buildkite/pipeline.yml
```diff
@@ -61,21 +63,43 @@ steps:
...
 
-  - label: ":running_shirt_with_sash: runtime tests"
+  - label: ":exclamation: event tests"
+    command:
+      >
+        docker run --rm \
+          localhost/firecracker-containerd-unittest-nonroot:${BUILDKITE_BUILD_NUMBER} \
+          '
+            cd /firecracker-containerd/eventbridge
+            go test . -v -count=1
```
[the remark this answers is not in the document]
sipsma: Done

.buildkite/pipeline.yml
```diff
@@ -61,21 +63,43 @@ steps:
             DISABLE_ROOT_TESTS='true' make test EXTRAGOARGS="-v -count=1"
           '
 
-  - label: ":running_shirt_with_sash: runtime tests"
+  - label: ":exclamation: event tests"
+    command:
```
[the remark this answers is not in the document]
sipsma: Done

agent/main.go
```diff
@@ -32,26 +33,19 @@ import (
...
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sys/unix"
+
+	"github.com/firecracker-microvm/firecracker-containerd/eventbridge"
 )
 
 const (
-	defaultPort = 10789
+	defaultPort      = 10789
+	defaultNamespace = "default"
```
samuelkarp: This is defined in `namespaces.Default`.

sipsma: Fixed while squashing commits

PR #120, opened 2019-03-04
 Specify network namespace for firecracker process

aaithal:
*Issue #, if available:* N/A

*Description of changes:* 
 * specify network namespace for firecracker process
 * start firecracker VM in non-default netns, when specified

*Testing done:* Modified the `taskworkflow` example to accept `-netns` argument. Tested end-to-end workflow using the same. 

```bash
$ sudo ./taskworkflow  -ip 10.0.5.33  -host_device tap301 -mac aa:bb:cc:dd:ee:ff -netns vlan-301
2019/03/04 19:32:10.943694 Creating containerd client
2019/03/04 19:32:10.944011 Created containerd client
2019/03/04 19:32:20.570249 Successfully pulled docker.io/library/nginx:latest image
2019/03/04 19:32:22.827486 Successfully created task: demo for the container
2019/03/04 19:32:22.827502 Completed waiting for the container task
2019/03/04 19:32:22.864207 Successfully started the container task
2019/03/04 19:32:25.864347 Executing http GET on 10.0.5.33
2019/03/04 19:32:25.865881 Response from [10.0.5.33]:
[<!DOCTYPE html>
<html>
<head>
<style>
    body {
        width: 35em;
        margin: 0 auto;
    }
</style>
</head>
<body>
<h1>Welcome to nginx!</h1>

</body>
</html>
]
10.0.3.109 - - [04/Mar/2019:19:32:25 +0000] "GET / HTTP/1.1" 200 612 "-" "Go-http-client/1.1" "-"
```


runtime/service.go
```diff
@@ -52,6 +54,7 @@ import (
 const (
 	defaultVsockPort     = 10789
 	supportedMountFSType = "ext4"
+	namedNetNSPath       = "/var/run/netns/"
```
sipsma: Just curious, what's the reason we are requiring the ns come from under `/var/run/netns`? As far as I knew you can bind a net NS fd anywhere you want on a filesystem, is this just a standard location?

runtime/service.go
```diff
@@ -702,6 +705,30 @@ func (s *service) startVM(ctx context.Context,
...
+		return s.machine.Start(vmmCtx)
+	}
+	// Get the network namespace handle.
+	netNS, err := ns.GetNS(namedNetNSPath + netNSName)
+	if err != nil {
+		return errors.Wrapf(err, "unable to find netns %s", netNSName)
+	}
+	// It's unsafe to switch network namespaces without locking down the OS
+	// thread. Lock it for the start VM operation.
+	pkgruntime.LockOSThread()
```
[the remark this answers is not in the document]
aaithal: nice catch! i didn't realize it was already doing that! i'll get rid of it.

runtime/service.go
aaithal: Named network namespaces are typically created under `/var/run/netns`. See `ip-netns(8)` [man page](http://man7.org/linux/man-pages/man8/ip-netns.8.html) for details. I have not seen too many (any tbh) implementations that use an alternative path.  You're right that the netns fd can certainly be bound anywhere on the filesystem. However, because the assumption is the use of named network namespaces, this path was chosen. An alternative could be letting clients specify `FirecrackerNetworkNamespacePath` instead of `FirecrackerNetworkNamespace`. But, it might be a bit much to specify was my take on it.

sipsma: Thanks, that makes sense. The other middle-ground option I can think of is check if the user specified an absolute or relative path; if they specified an absolute path use that unchanged, if it was relative append it to the end of `/var/run/netns/`.

If you agree that makes sense it's probably worth updating, but I don't consider it a blocker given it's just to support an obscure use-case.

aaithal: Since the most widely used library I'm aware of (for dealing with network namespaces in golang) makes this same assumption: https://github.com/vishvananda/netns/blob/13995c7128ccc8e51e9a6bd2b551020a27180abd/netns_linux.go#L81, I'd prefer to keep it the way it is. Let me know if you feel strongly about it and we can make that change.

sipsma: I see, no that works for me for now, easy to update in the future if ever needed.

aaithal:
> I think the right way to do this is with the jailer rather than making the runtime responsible for this.

Is there general agreement among the maintainers now about how firecracker-containerd will integrate with jailer? I agree that the right way to do this would be with jailer. This is something that unblocks folks who want to do this in the meantime. 
 
We can easily switch the internal implementation to jailer once #122 is unblocked.

examples/taskworkflow.go
```diff
@@ -33,26 +33,41 @@ import (
...
 func main() {
 	var ip = flag.String("ip", "", "ip address assigned to the container. Example: -ip 172.16.0.1")
 	var gateway = flag.String("gw", "", "gateway ip address. Example: -gw 172.16.0.1")
 	var netMask = flag.String("mask", "", "subnet gatway mask. Example: -mask 255.255.255.0")
+	var netNS = flag.String("netns", "", "firecracker VM network namespace. Example: -netNS testing")
+	var hostDevName = flag.String("host_device", defaultHostDevName,
+		"the host device name for the network interface, required when specifying 'netns'. Example: -host_device tap0")
+	var macAddress = flag.String("mac", defaultMacAddress,
+		"the mac address for the network interface, required when specifying 'netns'. Example: -mac AA:FC:00:00:00:01")
+	var kernelNetworkArgs = flag.Bool("kernel_nw_args", false,
```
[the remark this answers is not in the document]
nmeyerhans: +1 for hyphens rather than underscores.

I also would prefer something other than "nw" for network. Maybe `kernel-ip-arg` or `kernel-net-arg`?

runtime/service.go
aaithal: > If fully specified, we should use the concrete path rather than a relative path. This is useful when the ns is not bind-mounted anywhere but is instead one of the ns files from `/proc`, like `/proc/<pid>/ns/net`.

Agree. I'll make this change since two of you have asked for it. But, I still don't believe that clients would want to do that. It's much easier to manage/deal with named network namespaces and that's what will drive most of the usage of this field imo.

examples/taskworkflow.go
```diff
@@ -33,26 +33,41 @@ import (
...
+		"specifies if network params for the VMs should be included with kernel args. Example: -kernel_nw_args true")
```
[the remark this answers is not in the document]
aaithal: > Is this always true when `netns` is specified? Does it default to true when `netns` is specified? What about when `ip` is specified?

It's true only when someone sets it. Depending on how your VM is configured you or may not need it. My rootfs has a dhcp client. Hence, I don't need to set this at all. The `taskworkflow.md` file demonstrates use-cases with and without it. Please let me know if it's still unclear and I'll try to document more.

aaithal:
Closing this for now. The code/architecture has diverged a bit since I wrote this.

PR #123, opened 2019-03-05
*: Makefiles should depend on go.mod and go.sum

samuelkarp:

### DIFF
--- a/snapshotter/cmd/naive/Makefile
+++ b/snapshotter/cmd/naive/Makefile
@@ -11,9 +11,12 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
+GOMOD := $(shell go env GOMOD)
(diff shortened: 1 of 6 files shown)


snapshotter/cmd/naive/Makefile
```diff
@@ -13,7 +13,7 @@
 
 all: naive_snapshotter
 
-naive_snapshotter: *.go
+naive_snapshotter: *.go ../../../go.mod ../../../go.sum
```
mxpv: I would rather use this instead of `../../../`
```
GOMOD := $(shell go env GOMOD)
GOSUM := $(GOMOD:.mod=.sum)
```

samuelkarp: I had no idea that existed; that's great!  Will definitely switch.

samuelkarp: Done!

PR #132, opened 2019-03-10
Adding CRI Dockerfile

xibz:
The CRI Docker file will allow validation of CRI compliance

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,6 @@ sandbox-test-cri-run: sandbox-test-cri-build
 		--init \
 		--rm \
 		--privileged \
-		-v /tmp:/foo \
 		--security-opt seccomp=unconfined \
 		--ulimit core=0 \
 		localhost/sandbox-test-cri
--- a/Makefile
+++ b/Makefile
@@ -41,4 +41,16 @@ lint:
 install:
 	for d in $(SUBDIRS); do $(MAKE) -C $$d install; done
 
+sandbox-test-cri:
+	docker build -f sandbox/cri/Dockerfile -t "cri" .
+	docker run \
+		--init \
+		--rm \
+		--privileged \
+		-v /tmp:/foo \
+		--security-opt seccomp=unconfined \
+		--ulimit core=0 \
+		--device=/dev/kvm:/dev/kvm \
+		-t cri
+
 .PHONY: all $(SUBDIRS) clean proto deps lint install
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,8 @@ proto:
 
 clean:
 	for d in $(SUBDIRS); do $(MAKE) -C $$d clean; done
+	# --force is used to squash errors that pertain to file not existing
+	rm --force sandbox-test-cri-build-stamp
 
 deps:
 	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(GOPATH)/bin v1.12.3
@@ -41,8 +43,14 @@ lint:
 install:
 	for d in $(SUBDIRS); do $(MAKE) -C $$d install; done
 
-sandbox-test-cri:
-	docker build -f sandbox/cri/Dockerfile -t "cri" .
+sandbox-test-cri-build: sandbox-test-cri-build-stamp
+
+sandbox-test-cri-build-stamp:
+	docker build -f sandbox/cri/Dockerfile -t "localhost/sandbox-test-cri" .
+	@touch sandbox-test-cri-build-stamp
+
+sandbox-test-cri-run: sandbox-test-cri-build
+	test "$(shell id --user)" -eq 0
 	docker run \
 		--init \
 		--rm \
@@ -50,7 +58,8 @@ sandbox-test-cri:
 		-v /tmp:/foo \
 		--security-opt seccomp=unconfined \
 		--ulimit core=0 \
-		--device=/dev/kvm:/dev/kvm \
-		-t cri
+		localhost/sandbox-test-cri
 
-.PHONY: all $(SUBDIRS) clean proto deps lint install
+sandbox-test-cri: sandbox-test-cri-run
+
+.PHONY: all $(SUBDIRS) clean proto deps lint install sandbox-test-cri-run sandbox-test-cri-build sandbox-test-cri
--- a/Makefile
+++ b/Makefile
@@ -25,11 +25,6 @@ $(SUBDIRS):
 proto:
 	proto/generate.sh
 
-clean:
-	for d in $(SUBDIRS); do $(MAKE) -C $$d clean; done
-	# --force is used to squash errors that pertain to file not existing
-	rm --force sandbox-test-cri-build-stamp
-
 deps:
 	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(GOPATH)/bin v1.12.3
 	GO111MODULE=off go get -u github.com/vbatts/git-validation
@@ -61,5 +56,9 @@ sandbox-test-cri-run: sandbox-test-cri-build
 		localhost/sandbox-test-cri
 
 sandbox-test-cri: sandbox-test-cri-run
+
+clean:
+	for d in $(SUBDIRS); do $(MAKE) -C $$d clean; done
+	-rm sandbox-test-cri-build-stamp
 
 .PHONY: all $(SUBDIRS) clean proto deps lint install sandbox-test-cri-run sandbox-test-cri-build sandbox-test-cri
--- a/sandbox/cri/Dockerfile
+++ b/sandbox/cri/Dockerfile
@@ -41,7 +41,6 @@ FROM golang:1.12 as cri-build
 
 RUN apt-get update \
     && apt-get -y install --no-install-recommends \
-        cmake \
         curl \
         make \
         musl-tools \
--- a/sandbox/cri/Dockerfile
+++ b/sandbox/cri/Dockerfile
@@ -1,44 +1,21 @@
-FROM golang:1.11
+FROM golang:1.12 as base-build
 
-ADD . /firecracker-containerd/
-
-ARG TMP_BUILD_DIR=/tmp/build
-ARG FIRECRACKER_SRC_DIR="/firecracker"
-ARG FIRECRACKER_BUILD_DIR="$FIRECRACKER_SRC_DIR/build"
-ARG CARGO_REGISTRY_DIR="$FIRECRACKER_BUILD_DIR/cargo_registry"
-ARG FIRECRACKER_CONTAINERD_SRC_DIR="/firecracker-containerd"
-ARG FIRECRACKER_VERSION="0.15.0"
-
+ENV TMP_BUILD_DIR=/tmp/build
 ENV CARGO_HOME=/usr/local/rust
 ENV RUSTUP_HOME=/usr/local/rust
-ENV PATH="/go/bin:$CARGO_HOME/bin:$PATH"
-ENV CARGO_TARGET_DIR="$FIRECRACKER_BUILD_DIR/cargo_target"
+ENV RUST_VERSION=1.32.0
+ENV PATH="$CARGO_HOME/bin:$PATH"
+
+ARG FIRECRACKER_VERSION="0.15.0"
 
 # Install system dependecies
 #
 RUN apt-get update \
     && apt-get -y install --no-install-recommends \
-        binutils-dev \
         cmake \
         curl \
-        file \
-        g++ \
-        gcc \
-        git \
-        iperf3 \
-        iproute2 \
-        jq \
-        libcurl4-openssl-dev \
-        libdw-dev \
-        libiberty-dev \
-        libssl-dev \
-        lsof \
         make \
         musl-tools \
-        net-tools \
-        openssh-client \
-        pkgconf \
-        zlib1g-dev \
         libseccomp-dev \
         btrfs-tools \
         socat \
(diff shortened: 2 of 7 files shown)


sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,105 @@
...
+WORKDIR /go/src/github.com/containerd/containerd
+RUN git checkout v1.2.1
+RUN make
+RUN cp bin/* /usr/local/bin
+
+# Install runc
+RUN mkdir -p /go/src/github.com/opencontainers/runc
+RUN git clone https://github.com/opencontainers/runc /go/src/github.com/opencontainers/runc
+WORKDIR /go/src/github.com/opencontainers/runc
+RUN git checkout 96ec2177ae841256168fcf76954f7177af9446eb
```
mxpv: What rev is this? Can we use version tag instead?

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,105 @@
...
+RUN mkdir /var/run/firecracker-containerd
+
+# Install CRI
+WORKDIR /
+# CRI does not have support for Go modules yet.
+ENV GOPATH="/go"
+RUN go get github.com/containerd/cri
+RUN go get -u github.com/onsi/ginkgo/ginkgo
+WORKDIR /go/src/github.com/containerd/cri
+RUN make install.deps
```
mxpv: `install.deps` [installs](https://github.com/containerd/cri/blob/master/hack/install/install-deps.sh) containerd, runc, cni and other dependencies needed for cri. Do we really need to keep both `install.deps` and our own logic for installing deps ?

sipsma:
[Sam sent out an update to the quickstart guide](https://github.com/firecracker-microvm/firecracker-containerd/pull/130) that includes updating the versions of containerd and runc, you may want to update this Dockerfile similarly

Makefile
```diff
@@ -41,4 +41,16 @@ lint:
 install:
 	for d in $(SUBDIRS); do $(MAKE) -C $$d install; done
 
+sandbox-test-cri:
+	docker build -f sandbox/cri/Dockerfile -t "cri" .
```
samuelkarp: Please pick a name other than "cri".  How about "localhost/firecracker-containerd-sandbox-cri-test"?

Makefile
```diff
@@ -41,4 +41,16 @@ lint:
...
+	docker run \
```
samuelkarp: Like https://github.com/firecracker-microvm/firecracker-go-sdk/pull/87#discussion_r265366703, I'd like to see a requirement that this runs as root by checking the UID.

Makefile
```diff
@@ -41,4 +41,16 @@ lint:
...
+		--init \
+		--rm \
+		--privileged \
+		-v /tmp:/foo \
+		--security-opt seccomp=unconfined \
+		--ulimit core=0 \
+		--device=/dev/kvm:/dev/kvm \
+		-t cri
```
samuelkarp: Do you need a TTY in this container?

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,108 @@
+FROM golang:1.11
+
+ADD . /firecracker-containerd/
```
samuelkarp: Please use `COPY`, not `ADD`.

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,108 @@
...
+
+ARG TMP_BUILD_DIR=/tmp/build
```
samuelkarp: Why `ARG`? Do you intend for all of these to be overridden?

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,108 @@
...
+        socat \
+	sudo \
+	iptables \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install the Rust toolchain
+#
+# Install 1.32.0 due to Firecracker crashing with 148 code
+RUN mkdir "$TMP_BUILD_DIR" \
+    && curl https://sh.rustup.rs -sSf | sh -s -- -y \
```
samuelkarp: I'd like to see this done the same way as @sipsma is doing in https://github.com/firecracker-microvm/firecracker-containerd/pull/134/files#diff-6e72c6a06c3291e7fccf443ca341959cR70

Makefile
```diff
@@ -27,6 +27,8 @@ proto:
 
 clean:
 	for d in $(SUBDIRS); do $(MAKE) -C $$d clean; done
+	# --force is used to squash errors that pertain to file not existing
+	rm --force sandbox-test-cri-build-stamp
```
samuelkarp: Instead of using `--force`, prefix the command with `-` so it becomes `-rm`.  See the [`make` manual](https://www.gnu.org/software/make/manual/html_node/Errors.html) for details.

Makefile
```diff
@@ -41,4 +43,23 @@ lint:
 install:
 	for d in $(SUBDIRS); do $(MAKE) -C $$d install; done
 
-.PHONY: all $(SUBDIRS) clean proto deps lint install
+sandbox-test-cri-build: sandbox-test-cri-build-stamp
```
samuelkarp: nit: I would probably put all of these above `clean` because I tend to like that toward the bottom.  But I don't have a very strong opinion here.  @nmeyerhans do you have a preference?

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,95 @@
...
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends \
+        cmake \
+        curl \
+        make \
+        musl-tools \
+        libseccomp-dev \
+        btrfs-tools \
+        socat \
+	sudo \
```
samuelkarp: You have some inconsistent indentation.

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,95 @@
...
+ENV RUST_VERSION=1.32.0
+ENV PATH="$CARGO_HOME/bin:$PATH"
+
+ARG FIRECRACKER_VERSION="0.15.0"
+
+# Install system dependecies
+#
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends \
+        cmake \
```
samuelkarp: Do we need cmake?

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,95 @@
+FROM golang:1.12 as base-build
```
samuelkarp: It looks like this "base-build" image is only used for building Firecracker and not any Go programs.  Can you use the `rust:1.32` image instead and skip the need to install with rustup?

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,95 @@
...
+    && cp "$TMP_BUILD_DIR"/x86_64-unknown-linux-musl/release/firecracker /usr/local/bin \
+    && cp "$TMP_BUILD_DIR"/x86_64-unknown-linux-musl/release/jailer /usr/local/bin \
+    && mkdir /etc/containerd && touch /etc/containerd/config.toml \
+    && mkdir -p /go/src/github.com/containerd/containerd
+
+from golang:1.12 as cri-build
+
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends \
+        cmake \
```
samuelkarp: Do we need cmake for CRI?

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,95 @@
...
+WORKDIR /firecracker-containerd/
+COPY --from=base-build /usr/local/bin/firecracker /usr/local/bin/
+COPY --from=base-build /usr/local/bin/jailer /usr/local/bin/
+COPY --from=base-build /etc/containerd /etc/
+
+RUN mkdir -p /var/lib/firecracker-containerd/runtime \
+    && curl -fsSL -o /var/lib/firecracker-containerd/runtime/hello-vmlinux.bin https://s3.amazonaws.com/spec.ccfc.min/img/hello/kernel/hello-vmlinux.bin \
+    && curl -fsSL -o /var/lib/firecracker-containerd/runtime/hello-rootfs.ext4 https://s3.amazonaws.com/spec.ccfc.min/img/hello/fsfiles/hello-rootfs.ext4 \
+    && e2fsck -f -y /var/lib/firecracker-containerd/runtime/hello-rootfs.ext4 \
+    && resize2fs /var/lib/firecracker-containerd/runtime/hello-rootfs.ext4 1g \
```
samuelkarp: 1g?

sandbox/cri/config.toml
```diff
@@ -0,0 +1,189 @@
+# Firecracker
+[proxy_plugins]
+  [proxy_plugins.firecracker-naive]
+    type = "snapshot"
+    address = "/var/run/firecracker-containerd/naive-snapshotter.sock"
+  [proxy_plugins.firecracker-dm-snapshotter]
+    type = "snapshot"
+    address = "/var/run/firecracker-dm-snapshotter.sock"
```
samuelkarp: Any reason the paths are inconsistent for the dm snapshotter and naive snapshotter?

sandbox/cri/config.toml
```diff
@@ -0,0 +1,189 @@
...
+    # no_pivot disables pivot-root (linux only), required when running a container in a RamDisk with runc.
+    # This only works for runtime type "io.containerd.runtime.v1.linux".
+    # DEPRECATED: use Runtime.Options for runtime specific config for shim v2 runtimes.
+    #   For runtime "io.containerd.runc.v1", use the option `NoPivotRoot`.
+    no_pivot = false
+
+    # "plugins.cri.containerd.default_runtime" is the runtime to use in containerd.
+    [plugins.cri.containerd.default_runtime]
+      # runtime_type is the runtime type to use in containerd e.g. io.containerd.runtime.v1.linux
+      runtime_type = "aws.firecracker"
```
samuelkarp: Looks like this should be something like "io.containerd.runtime.v2.linux" rather than the name of the runtime, right?

sandbox/cri/config.toml
```diff
@@ -0,0 +1,189 @@
...
+
+      # "plugins.cri.containerd.default_runtime.options" is options specific to
+      # the default runtime. The options type for "io.containerd.runtime.v1.linux" is:
+      #   https://github.com/containerd/containerd/blob/v1.2.0-rc.1/runtime/linux/runctypes/runc.pb.go#L40
+      # NOTE: when `options` is specified, all related deprecated options will
+      #   be ignored, including `systemd_cgroup`, `no_pivot`, `runtime_engine`
+      #   and `runtime_root`.
+      [plugins.cri.containerd.default_runtime.options]
+        # Runtime is the binary name of the runtime.
+        Runtime = ""
```
samuelkarp: I think this is where the name of the runtime goes.

sandbox/cri/cri.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -ex
+
+mkdir /tmp/mnt
+sudo mount /var/lib/firecracker-containerd/runtime/hello-rootfs.ext4 /tmp/mnt
+sudo cp /firecracker-containerd/agent/agent /tmp/mnt/usr/local/bin
+sudo cp $(which runc) /tmp/mnt/usr/local/bin
+sudo cp /firecracker-containerd/sandbox/cri/fc-agent.start /tmp/mnt/etc/local.d
```
samuelkarp: Looks like you're missing a few lines:

```
ln -s /etc/init.d/local /tmp/mnt/etc/runlevels/default/local
ln -s /etc/init.d/cgroups /tmp/mnt/etc/runlevels/default/cgroups
```

sandbox/cri/firecracker-runtime.json
```diff
@@ -0,0 +1,14 @@
+{
+
+  "firecracker_binary_path": "/usr/local/bin/firecracker",
+  "socket_path": "./firecracker.sock",
```
samuelkarp: When you rebase, you can remove `socket_path`.

sandbox/cri/config.toml
xibz: What this is doing is looking for the container shim. It does this by taking a look at the runtime type and creating `container-shim-aws-firecracker` name from `aws.firecracker`. If we were to use `io.containerd.runtime.v2.linux` it would look for `container-shim-v2-linux`. No idea why the `io.containerd` is removed, though.

sandbox/cri/Dockerfile
xibz: Nope, will remove

Makefile
xibz: I moved it down as after pointing it out I prefer that as well.

sandbox/cri/Dockerfile
xibz: 1g may be overkill, but the reason why it is needed is CRI requires free space to be on the root img. I'll lower it and see what is a good number to use.

sipsma: If it needs to be fairly large still, another possibility to consider would be to keep it small during the docker build (which will keep the image bloat minimal) and add the resize step to the entrypoint.

The tradeoff would be a smaller image size for a little extra delay during container startup at runtime, so it's hard to say whether it's worth it, just something to consider if needed.

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+        libseccomp-dev \
+        btrfs-tools \
+        socat \
+        sudo \
+        iptables \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install the Rust toolchain
+#
+# Install 1.32.0 due to Firecracker crashing with 148 code
```
samuelkarp: You shouldn't need to install rust or rustup, given that you're using the `rust:1.32` base image.  It should already have rust 1.32 and the rustup tool installed.

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+# Install system dependecies
+#
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends \
+        curl \
+        make \
+        musl-tools \
+        libseccomp-dev \
+        btrfs-tools \
+        socat \
```
samuelkarp: Why do you need socat?

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+        sudo \
```
samuelkarp: Why do you need sudo?

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+        iptables \
```
samuelkarp: Why do you need iptables?

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends \
+        cmake \
+        curl \
+        make \
+        musl-tools \
+        libseccomp-dev \
+        btrfs-tools \
+        socat \
```
samuelkarp: Same questions here about socat, sudo, and iptables.

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+
+# Install system dependecies
+#
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends \
+        curl \
+        make \
+        musl-tools \
+        libseccomp-dev \
+        btrfs-tools \
```
samuelkarp: I don't believe Firecracker requires btrfs-tools.

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+from cri-build as firecracker-containerd-build
+ENV PATH="/go/bin:$PATH"
+
+# Build and install firecracker-containerd
+WORKDIR /firecracker-containerd/
+COPY --from=base-build /usr/local/bin/firecracker /usr/local/bin/
+COPY --from=base-build /usr/local/bin/jailer /usr/local/bin/
+COPY --from=base-build /etc/containerd /etc/
+
+RUN mkdir -p /var/lib/firecracker-containerd/runtime \
```
samuelkarp: I would move this `RUN` before the `COPY --from=base-build` lines to avoid downloading the kernel and rootfs every time firecracker or containerd change.

sandbox/cri/config.toml
```diff
@@ -0,0 +1,189 @@
+# Firecracker
+[proxy_plugins]
+  [proxy_plugins.firecracker-naive]
+    type = "snapshot"
+    address = "/var/run/firecracker-containerd/snapshotter.sock"
```
samuelkarp: You can't have both firecracker-naive and firecracker-dm-snapshotter using the same socket address.

(Also, I'd probably use "firecracker-naive" and "firecracker-dm" to be consistent with each other.)

sandbox/cri/config.toml
samuelkarp: Are you sure that's the runtime type field and not the runtime name field?  container-shim-aws-firecracker is a runtime of type io.containerd.runtime.v2.linux.

sandbox/cri/config.toml
```diff
@@ -0,0 +1,189 @@
...
+  # restrict_oom_score_adj indicates to limit the lower bound of OOMScoreAdj to
+  # the containerd's current OOMScoreAdj.
+  # This is useful when the containerd does not have permission to decrease OOMScoreAdj.
+  restrict_oom_score_adj = false
+
+  # "plugins.cri.containerd" contains config related to containerd
+  [plugins.cri.containerd]
+
+    # snapshotter is the snapshotter used by containerd.
+    snapshotter = "firecracker-naive"
```
samuelkarp: Are you planning to test with both firecracker-naive and firecracker-dm?

sandbox/cri/config.toml
xibz: I tried using `io.containerd.runtime.v2.linux` and it failed trying to find `container-shim-v2-linux`. I can mess around with the configuration some more, but I know this configuration talks to our runtime.

sandbox/cri/Dockerfile
xibz: socat is a required dependency of CRI. Something to do with portforward?

sandbox/cri/Dockerfile
xibz: because the `cri.sh` script uses sudo to mount the root image to install the agent and all that jazz.

sandbox/cri/Dockerfile
xibz: Ah cool, so apparently the `rust:1.32` image doesn't have a default tool chain specified which was resulting in an error. However, that can be fixed by adding the `rust default "$RUST_VERSION" as a `RUN` command.

sandbox/cri/Dockerfile
xibz: it's a CRI dependency

sandbox/cri/Dockerfile
xibz: CRI dependency

sandbox/cri/Dockerfile
mxpv: It doesn't ) I bet this is for `containerd`. @xibz how about using `no_btrfs` build tag instead?

xibz: Ah cool! Didn't know that was a thing. Thanks, will update.

mxpv: `make` should accept `BUILDTAGS` parameter. https://github.com/containerd/cri/blob/eb27e526f5d23c4c762d25f80351c632efe4f7ab/hack/install/install-containerd.sh#L40
Can you try it?

sandbox/cri/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+
+WORKDIR firecracker
+RUN git checkout tags/v"$FIRECRACKER_VERSION" \
+    && cargo build --release --features vsock --target-dir "$TMP_BUILD_DIR" \
+    && cp "$TMP_BUILD_DIR"/x86_64-unknown-linux-musl/release/firecracker /usr/local/bin \
+    && cp "$TMP_BUILD_DIR"/x86_64-unknown-linux-musl/release/jailer /usr/local/bin \
+    && mkdir /etc/containerd && touch /etc/containerd/config.toml \
+    && mkdir -p /go/src/github.com/containerd/containerd
+
+from golang:1.12 as cri-build
```
mxpv: While instructions are not case sensitive, the convention is to use uppercase letters.
https://docs.docker.com/v17.09/engine/reference/builder/#format

Makefile
```diff
@@ -41,4 +38,27 @@ lint:
...
+	docker build -f sandbox/cri/Dockerfile -t "localhost/sandbox-test-cri" .
+	@touch sandbox-test-cri-build-stamp
+
+sandbox-test-cri-run: sandbox-test-cri-build
+	test "$(shell id --user)" -eq 0
+	docker run \
+		--init \
+		--rm \
+		--privileged \
+		-v /tmp:/foo \
```
nmeyerhans: `/foo`? I didn't see any place where this is used.

sandbox/cri/Dockerfile
nmeyerhans: But you're running as root in the container, no?

sandbox/cri/cri.sh
```diff
@@ -0,0 +1,17 @@
...
+sudo ln -s /etc/init.d/local /tmp/mnt/etc/runlevels/default/local
+sudo ln -s /etc/init.d/cgroups /tmp/mnt/etc/runlevels/default/cgroups
+sudo umount /tmp/mnt
+naive_snapshotter -address /var/run/firecracker-containerd/snapshotter.sock -path /var/lib/firecracker-snapshotter &> /dev/null &
+pid=$!
+make test-cri
+exec "$@"
+
+kill -9 $pid
```
nmeyerhans: On line 15, you're `exec`ing another process, so execution will never reach this line. You should probably use the shell's `trap` builtin to execute cleanup code upon termination. Also, avoid `kill -9` except in an escalation sort of scenario. State management programs like snapshotters almost certainly are going to want to perform some pre-shutdown housekeeping. (I realize this is an ephemeral container, so if you _really_ want to use `SIGKILL`, at least include a comment stating why it's the right thing to use here.)

sandbox/cri/Dockerfile
xibz: Done!

Makefile
```diff
@@ -41,4 +38,27 @@ lint:
...
+sandbox-test-cri-build-stamp:
+	docker build -f sandbox/cri/Dockerfile -t "localhost/sandbox-test-cri" .
+	@touch sandbox-test-cri-build-stamp
+
+sandbox-test-cri-run: sandbox-test-cri-build
+	test "$(shell id --user)" -eq 0
+	docker run \
+		--init \
+		--rm \
+		--privileged \
```
nmeyerhans: Do we really, really need `--privileged`? Can we get away with some added capabilities instead?

sandbox/cri/Dockerfile
xibz: Yea, turns out I added sudo due to the CRI install scripts using sudo. To which, if it isn't installed, fails.

Makefile
xibz: Can you elaborate on what you mean by this? We only need this for mounting.

sandbox/cri/config.toml
xibz: I removed the devmapper and will only test against naive, for now at least.

Makefile
nmeyerhans: That's exactly my point. We do not need all the access that comes with `--privileged` just to call mount.

xibz: Ah, I tried to remove `--privileged` and add sudo to the command, but that fails. /=

nmeyerhans: As expected. `sudo` doesn't mean anything, especially considering that you're already running as root inside the container. The container itself is constrained. There are several ways to remove some or all of the constraints. `--privileged` removes them all, but that's not usually necessary. In this case, you definitely don't need to run with full privileges. Try adding `CAP_SYS_ADMIN` via docker's `--cap-add` flag.

xibz: Hm, that still doesn't work for me. Still get the good ol' `mount failed: Unknown error -1`.

sandbox/cri/Dockerfile
samuelkarp: This stage of the Dockerfile is not used to build CRI or containerd.

sandbox/cri/Dockerfile
samuelkarp: This stage of the Dockerfile is not used to build CRI or containerd.

sandbox/cri/Dockerfile
samuelkarp: This stage of the Dockerfile is not used to build CRI or containerd.

sandbox/cri/Dockerfile
samuelkarp: This stage of the Dockerfile is not used to build CRI or containerd.

sandbox/cri/Dockerfile
samuelkarp: Unresolved as you didn't answer my question.

xibz: Yea, cmake isn't needed. So removed.

xibz:
Closing due to inactivity and age

PR #134, opened 2019-03-11
Run tests in containers.

sipsma:
The original motivation for this change was to enable tests that verify
our shim exits as expected when a container exits. We decided this was
best enabled via running tests in containers, which allows you to
easily separate shim processes running in parallel from one another
and make assertions about the one created by your test in particular.
The Dockerfile used to build the environment in which tests run should
be useful in general too for enabling devs to quickly rebuild/test
changes.

While Buildkite does offer plugins for docker-compose and docker, they
are not used here for a few reasons:
1. docker-compose doesn't support docker-buildkit, which we use to
   parallelize builds and cache gomod/apt dependencies, which both
   speed up the builds (~4 mins vs. ~10 mins) and improve reliability
2. The plain docker Buildkite plugin doesn't support all the parameters
   we need to configure, such as setting the ipc namespace to host for
   the snapshotter tests

As of this commit, only the "firecracker-containerd-test-base" image is actually used by our CI system; the "firecracker-containerd-e2etest" and "firecracker-containerd-dev" images are not yet. However, I wanted to keep them here for a few reasons:
1. They can already provide utility by giving devs a way to quickly test changes to code (this was helpful for me while trying to get firecracker 0.15 working in the docker container)
2. They will be used in the CI system very soon to implement a test for the new event forwarder

If there are objections to this, I can remove those stages for now and add them later when I add the test for the event forwarder PR, let me know.


tools/docker/Dockerfile
```diff
@@ -0,0 +1,184 @@
...
+
+
+
+
+# Build firecracker itself
+# TODO can we cache cargo artifacts like we do go modules?
+FROM build-base as firecracker-build
+SHELL ["/bin/bash", "-c"]
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain 1.32.0 \
+&& source /root/.cargo/env \
```
xibz: Can we add tabbing here and the various other places? This will visually hint that this is grouped with the statement above.

tools/docker/Dockerfile
```diff
@@ -0,0 +1,184 @@
...
+
+
+
+# Build a rootfs for the microVM, including runc and firecracker-containerd's agent
+FROM alpine:3.8 as firecracker-vm-root
+COPY --from=runc-build /output/runc /usr/local/bin/
+COPY --from=firecracker-containerd-build /output/agent /usr/local/bin/
+ADD tools/docker/fc-agent.start /etc/local.d/fc-agent.start
+RUN apk add musl \
+busybox \
```
xibz: same here

tools/docker/Dockerfile
```diff
@@ -0,0 +1,184 @@
...
+alpine-baselayout \
```
xibz: Are all of these needed? Like why is ncurses needed?

tools/docker/Dockerfile
sipsma: Sure

tools/docker/Dockerfile
sipsma: This list was obtained by just getting the list of packages installed by firecracker's "hello-world" rootfs and then removing some SSH packages which I knew for sure were not needed.

I'll try out removing some more like ncurses to get it down to the bare minimum

tools/docker/fc-agent.start
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+mkdir -p /container
+exec > /container/agent-debug.log # Debug logs from the agent
+exec 2>&1
+touch /container/runtime
+mkdir /container/rootfs
+mount -t auto -o rw /dev/vdb /container/rootfs
+cd /container
+/usr/local/bin/agent -id 1 -debug &
```
nmeyerhans: Please make sure all files have a trailing newline, unless there are specific reasons why some shouldn't.

tools/docker/Dockerfile
```diff
@@ -0,0 +1,184 @@
...
+ca-certificates
+
+
+
+
+# Build firecracker itself
+# TODO can we cache cargo artifacts like we do go modules?
+FROM build-base as firecracker-build
+SHELL ["/bin/bash", "-c"]
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain 1.32.0 \
```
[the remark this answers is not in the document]
sipsma: Yeah makes sense, I avoided the rust base image initially just because I wanted to make 100% sure all the builds were using the same set of dependencies (that we apt install in `build-base`).

It turns out [the rust dockerfile](https://github.com/rust-lang/docker-rust/blob/edda09581f7c015c5cc2b8f0e04951a83b6f9727/1.32.0/stretch/Dockerfile) happens to also use `debian:stretch` as a base and works by downloading `rust-init` and verifying it against hardcoded SHA sums. I'll switch to using that base instead for the firecracker install.

tools/docker/Dockerfile
```diff
@@ -0,0 +1,184 @@
...
+&& source /root/.cargo/env \
+&& rustup target add x86_64-unknown-linux-musl \
+&& git clone https://github.com/firecracker-microvm/firecracker.git \
+&& cd firecracker \
+&& git checkout v0.15.2 \
+&& cargo build --release --features vsock --target x86_64-unknown-linux-musl \
```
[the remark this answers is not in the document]
sipsma: Also makes sense, though I believe we need root for a few select portions such as `apt install`.

How about I try to get all the root-specific stuff done in isolated build stages (such as `build-base`) and then use `USER <someone>` for all the others?

tools/docker/fc-agent.start
sipsma: Will Do

tools/docker/Dockerfile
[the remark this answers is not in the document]
sipsma: Using the Rust base image ended up causing complication with wanting to run the builds as a non-root user, so I just took a few lines from their dockerfile (download rust-init and verify its sha sum) and put it in here.

tools/docker/Dockerfile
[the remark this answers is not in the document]
sipsma: All of the compilation runs as root now. The parts that still execute as root are
1. apt install at the beginning 
2. alpine apk install and VM image creation
3. creation of the images used for testing (just some apt installs and copying artifacts from previous stages)

There are no more steps (to my knowledge) that compile code from unverified, third-party sources as root.

sipsma:
The travis CI build failed because of commits layered on top of the first one that I'm planning on squashing into one before merging, so I don't think it's a "real" failure.

.buildkite/pipeline.yml
```diff
@@ -12,27 +12,48 @@
...
-  # multiple go builds are downloading to the same cache.
-  - wait
+  - label: ":docker: Build"
+    env:
+      DOCKER_BUILDKIT: "1"
+    command:
+      >
+        docker build \
+          --progress=plain \
+          -f tools/docker/Dockerfile \
```
samuelkarp: Can we use long flags consistently?  `--file` and `--tag` instead of `-f` and `-t` respectively.

.buildkite/pipeline.yml
```diff
@@ -12,27 +12,48 @@
...
+          --target firecracker-containerd-test-e2e \
+          -t firecracker-containerd-test-e2e .
```
samuelkarp: We should use something like `--tag "localhost/firecracker-containerd-test-e2e:${BUILDKITE_BUILD_NUMBER}"` so that parallel invocations of `docker build` don't stomp over each other.

.buildkite/pipeline.yml
```diff
@@ -12,27 +12,48 @@
...
 
-  - label: 'build'
-    command: 'make'
+        docker build \
+          --progress=plain \
+          -f tools/docker/Dockerfile \
+          --target firecracker-containerd-test-base \
+          -t firecracker-containerd-test-base .
```
samuelkarp: Same here as above.

.buildkite/pipeline.yml
```diff
@@ -12,27 +12,48 @@
...
-  - label: ':rotating_light: snapshotter *root* tests'
-    commands:
-      - "cd snapshotter"
-      - "sudo make test EXTRAGOARGS='-v -count=1'"
+  - label: ":hammer: snapshotter tests"
+    command:
+      >
+        docker run --rm \
+          --privileged \
+          -v /dev:/dev \
```
samuelkarp: `--volume` instead of `-v`

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+# BUILD IMAGES
+#
+#########################################
+
+
+# Common tools needed for the build stages ahead. The final test images do not inherit directly from here, so this bloat
+# is dropped in those final end-use images.
+FROM debian:stretch as build-base
+ENV PATH /bin:/usr/bin:/usr/local/bin:/sbin:/usr/sbin:/usr/local/sbin:/usr/lib/go/bin
+RUN --mount=type=cache,id=aptcache,target=/var/cache/apt \
```
samuelkarp: This is to share the apt cache among containers?  Is there a lot of value in doing this given that we're unconditionally running `apt update` anyway?

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+	--mount=type=cache,id=aptlib,target=/var/lib/apt,sharing=locked \
+	mkdir -p /etc/apt/sources.list.d \
+	&& echo "deb http://ftp.debian.org/debian stretch-backports main" > /etc/apt/sources.list.d/stretch-backports.list \
+	&& apt update \
```
samuelkarp: I think it's typically recommended to do `DEBIAN_FRONTEND=noninteractive apt-get update` and `DEBIAN_FRONTEND=noninteractive apt-get install --yes --no-install-recommends`.  The `--no-install-recommends` should reduce the amount of time it takes to build the image, as fewer packages will need to be downloaded, extracted, and installed.

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+
+# Run as non-root for the rest of the build stages now that we've gotten apt installs out of the way.
+RUN useradd --create-home builder \
+	&& mkdir /output \
+	&& chown builder /output \
+	&& mkdir -p /home/builder/go
+
+# Buildkit caches don't support anything like a "--chown" flag yet, so ensure builder will have access to them
+RUN --mount=id=gomod,target=/home/builder/go/pkg/mod,type=cache \
+	chown -R builder /home/builder/go
```
samuelkarp: There's no guarantee that every `builder` user in each of these builds is going to have the same UID, meaning we could have some strange behavior when running parallel builds.  You can specify the UID in the `useradd` command above with the `--uid` flag, and that should make this safer.

(There are still issues with concurrent mutation of the Go module cache though, and the only way I can think of fixing that is to use a separate cache per build agent, but that negates some of the value of having a cache...)

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+	RUST_VERSION=1.32.0
+
+# TODO using ADD <url> should work instead of having to wget, but buildkit has a bug where --chown is ignored for URLs
+# https://github.com/moby/buildkit/issues/813
+RUN wget "https://static.rust-lang.org/rustup/archive/1.16.0/x86_64-unknown-linux-gnu/rustup-init" \
+	&& echo "2d4ddf4e53915a23dda722608ed24e5c3f29ea1688da55aa4e98765fc6223f71 rustup-init" | sha256sum -c - \
+	&& chmod +x rustup-init \
+	&& ./rustup-init -y --no-modify-path --default-toolchain $RUST_VERSION
+
+# TODO can we cache cargo artifacts like we do go modules?
```
samuelkarp: Yep.  I believe that it's in `~/.cargo/registry`.

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+
+
+
+# Build runc
+FROM build-base as runc-build
+RUN --mount=id=gomod,target=/home/builder/go/pkg/mod,type=cache \
+	mkdir -p /home/builder/go/src/github.com/opencontainers/runc \
+	&& git clone https://github.com/opencontainers/runc /home/builder/go/src/github.com/opencontainers/runc \
+	&& cd /home/builder/go/src/github.com/opencontainers/runc \
+	&& git checkout 6635b4f0c6af3810594d2770f662f34ddc15b40d \
```
samuelkarp: Should leave a comment explaining where this came from.

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+
+
+
+# Build containerd
+FROM build-base as containerd-build
+RUN --mount=id=gomod,target=/home/builder/go/pkg/mod,type=cache \
+	mkdir -p /home/builder/go/src/github.com/containerd/containerd \
+	&& git clone https://github.com/containerd/containerd.git /home/builder/go/src/github.com/containerd/containerd \
+	&& cd /home/builder/go/src/github.com/containerd/containerd \
+	&& git checkout v1.2.4 \
```
samuelkarp: 1.2.5 is out

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+
+#########################################
+#
+# TEST IMAGES
+#
+#########################################
+
+
+
+# Just enough to run basic firecracker-containerd tests
```
samuelkarp: How come you have this one separate from firecracker-containerd-test-e2e?  Just for the visual separation?

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+# and the snapshotter in the background. The cmd drops to a shell if not otherwise overridden.
+FROM firecracker-containerd-test-base as firecracker-containerd-test-e2e
+COPY --from=firecracker-build /output/* /usr/local/bin/
+COPY --from=containerd-build /output/* /usr/local/bin/
+COPY --from=firecracker-vm-root-builder /output/vm.ext4 /var/lib/firecracker-containerd/runtime/hello-rootfs.ext4
+COPY --from=firecracker-containerd-build /output/* /usr/local/bin/
+COPY --from=runc-build /output/* /usr/local/bin/
+
+ADD https://s3.amazonaws.com/spec.ccfc.min/img/hello/kernel/hello-vmlinux.bin /var/lib/firecracker-containerd/runtime/hello-vmlinux.bin
+ADD tools/docker/config.toml /etc/containerd/config.toml
```
samuelkarp: These should be `COPY` unless you have a specific need for `ADD`.

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+# Enough to start firecracker, firecracker-containerd and run full e2e tests. The entrypoint will start containerd
+# and the snapshotter in the background. The cmd drops to a shell if not otherwise overridden.
+FROM firecracker-containerd-test-base as firecracker-containerd-test-e2e
+COPY --from=firecracker-build /output/* /usr/local/bin/
+COPY --from=containerd-build /output/* /usr/local/bin/
+COPY --from=firecracker-vm-root-builder /output/vm.ext4 /var/lib/firecracker-containerd/runtime/hello-rootfs.ext4
+COPY --from=firecracker-containerd-build /output/* /usr/local/bin/
+COPY --from=runc-build /output/* /usr/local/bin/
+
+ADD https://s3.amazonaws.com/spec.ccfc.min/img/hello/kernel/hello-vmlinux.bin /var/lib/firecracker-containerd/runtime/hello-vmlinux.bin
```
samuelkarp: I'd prefer to see this as a `RUN` with `curl` and validating the downloaded file's sha256sum.

tools/docker/entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -ex
+
+/usr/local/bin/naive_snapshotter \
+     -address /var/run/firecracker-containerd/naive-snapshotter.sock \
+     -path /var/lib/firecracker-containerd/naive \
+     -debug &> /var/log/naive_snapshotter.out &
```
samuelkarp: Are we exposing `/var/log/naive_snapshotter.out` and `/var/log/containerd.out` somewhere?  These will be useful to see in test failures.

tools/docker/Dockerfile
nmeyerhans: Yeah, `apt` will whine when used without a controlling TTY. Better to use `apt-get`.

`DEBIAN_FRONTEND=noninteractive` isn't needed with the `update` operation, but it should definitely be used with `(dist-)upgrade` and `install` operations.

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+	mkdir -p /etc/apt/sources.list.d \
+	&& echo "deb http://ftp.debian.org/debian stretch-backports main" > /etc/apt/sources.list.d/stretch-backports.list \
+	&& apt update \
+	&& apt --target-release stretch-backports install -y golang-go \
+	&& apt install -y \
+		git \
+		wget \
+		build-essential \
+		libseccomp-dev \
+		btrfs-tools \
```
nmeyerhans: Why do we need `btrfs-tools`?

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+
+
+
+
+# Build firecracker itself
+FROM build-base as firecracker-build
+ENV RUSTUP_HOME=/home/builder/rustup \
+    CARGO_HOME=/home/builder/cargo \
+    PATH=/home/builder/cargo/bin:$PATH \
+	RUST_VERSION=1.32.0
```
nmeyerhans: nit: whitespace

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+
+# TODO using ADD <url> should work instead of having to wget, but buildkit has a bug where --chown is ignored for URLs
+# https://github.com/moby/buildkit/issues/813
+RUN wget "https://static.rust-lang.org/rustup/archive/1.16.0/x86_64-unknown-linux-gnu/rustup-init" \
```
nmeyerhans: I'd usually prefer curl over wget, with options such as `--max-time`, `--retry`, and `--silent` (maybe among others, depending on circumstances). You don't need to switch to curl, but it'd be nice make sure equivalent options are applied here (or not, if wget's defaults are reasonable).

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+
+
+
+# Just enough to run basic firecracker-containerd tests
+FROM debian:stretch as firecracker-containerd-test-base
+RUN --mount=type=cache,id=aptcache,target=/var/cache/apt \
+ 	--mount=type=cache,id=aptlib,target=/var/lib/apt,sharing=locked \
+	mkdir -p /etc/apt/sources.list.d \
+	&& echo "deb http://ftp.debian.org/debian stretch-backports main" > /etc/apt/sources.list.d/stretch-backports.list \
+	&& apt update \
```
nmeyerhans: Similar comments to above about `DEBIAN_FRONTEND` and `apt-get`

tools/docker/Dockerfile
```diff
@@ -0,0 +1,231 @@
...
+COPY --from=firecracker-build /output/* /usr/local/bin/
+COPY --from=containerd-build /output/* /usr/local/bin/
+COPY --from=firecracker-vm-root-builder /output/vm.ext4 /var/lib/firecracker-containerd/runtime/hello-rootfs.ext4
+COPY --from=firecracker-containerd-build /output/* /usr/local/bin/
+COPY --from=runc-build /output/* /usr/local/bin/
+
+ADD https://s3.amazonaws.com/spec.ccfc.min/img/hello/kernel/hello-vmlinux.bin /var/lib/firecracker-containerd/runtime/hello-vmlinux.bin
+ADD tools/docker/config.toml /etc/containerd/config.toml
+ADD tools/docker/firecracker-runtime.json /etc/containerd/firecracker-runtime.json
+ADD tools/docker/entrypoint.sh /entrypoint.sh
```
nmeyerhans: nit: It's often a good idea to drop file extensions on installation. If we ever decide to rewrite this script in something other than shell, it'd be nice not to have to worry about chasing down and updating all references. (Not a big deal here, just something to consider generally.)

.buildkite/pipeline.yml
nmeyerhans: Am I missing something or are we not actually using this image anywhere?

tools/docker/Dockerfile
sipsma: I just realized I forgot to address the second part of your comment here. I will make another quick PR update that uses the `sharing=locked` flag to prepopulate the cache in each stage and then in a second RUN step actually does the build with `sharing=private`, which should result in safe concurrent usage without locking the cache for too long.

sipsma: Got this working, but worth noting that once we upgrade to Go 1.12 we should be able to remove this workaround: https://github.com/golang/go/issues/26794

tools/docker/entrypoint.sh
samuelkarp: > need to ensure each test run is carved out its own directory on the host which is eventually cleaned up after a while

If you write to the current working directory, the Buildkite Agent will (attempt) to clean it up on exit.  (The "attempt" is because the Buildkite Agent does not run as root, and will get tripped up by file permissions.)

sipsma: Got this working, including uploading as buildkite artifacts! Thanks!

mxpv:
I'm not super fan of huge `Dockerfile` with lots of dependencies between stages as it's harder to maintain. I would rather split it on smaller files. But probably this is fine for now.

tools/docker/Dockerfile
```diff
@@ -0,0 +1,272 @@
...
+
+# Build a rootfs for the microVM, including runc and firecracker-containerd's agent
+FROM alpine:3.8 as firecracker-vm-root
+COPY --from=runc-build /output/runc /usr/local/bin/
+COPY --from=firecracker-containerd-build /output/agent /usr/local/bin/
+ADD --chown=builder tools/docker/fc-agent.start /etc/local.d/fc-agent.start
+RUN apk add openrc \
+	&& ln -s /etc/init.d/local /etc/runlevels/default/local \
+	&& ln -s /etc/init.d/cgroups /etc/runlevels/default/cgroups \
+	&& ln -s /etc/init.d/agetty /etc/runlevels/default/agetty.ttyS0 \
```
nmeyerhans: This is probably an issue with alpine, not your Dockerfile, but it seems that `/etc/init.d/agetty` tries to run `/sbin/agetty`, which doesn't exist in alpinelinux. The alpine busybox installation creates a `/sbin/getty` symlink, so it may be sufficient to symlink `/sbin/agetty` to that, if the invocation is compatible.

This also may not be critical here.

sipsma:
> I'm not super fan of huge `Dockerfile` with lots of dependencies between stages as it's harder to maintain. I would rather split it on smaller files somehow. But probably this is fine for now.

I agree this Dockerfile is pretty much at max capacity at this point. Unfortunately, splitting it up into separate files would make it more difficult for buildkit to automatically parallelize different build stages. If we expand it in the future we can consider splitting it up into separate files and then just appending them together into a single file during `docker build` as a Makefile target or something like that.

sipsma:
Combined all the fixup commits into one; there was a final fix needed which I noticed this morning after seeing a build on a fresh instance related to setting the go mod cache permissions. This is working now and both Travis+Buildkite builds are passing.

tools/docker/Dockerfile
sipsma: Just saw this, I will try removing it to see if it's needed at all and, if it is, the symlinking suggestion

sipsma: I removed this line entirely and I'm still able to start up a shell in a container in the VM just fine, so I'd guess that we only need the TTY working if we want a shell directly on the VM (as opposed to one inside a container inside the VM). Just leaving the line out for now. Thanks!

tools/docker/Dockerfile
```diff
@@ -0,0 +1,270 @@
...
+
+RUN curl --silent --show-error --retry 3 --max-time 30 --output rustup-init \
+	"https://static.rust-lang.org/rustup/archive/1.16.0/x86_64-unknown-linux-gnu/rustup-init" \
+	&& echo "2d4ddf4e53915a23dda722608ed24e5c3f29ea1688da55aa4e98765fc6223f71 rustup-init" | sha256sum -c - \
+	&& chmod +x rustup-init \
+	&& ./rustup-init -y --no-modify-path --default-toolchain $RUST_VERSION
+
+RUN --mount=type=cache,from=build-base,source=/home/builder/cargo/registry,target=/home/builder/cargo/registry \
+	source /home/builder/cargo/env \
+	&& rustup target add x86_64-unknown-linux-musl \
```
samuelkarp: Any particular reason the `rustup target add` is done as part of this `RUN` block and not the previous one?

.buildkite/pipeline.yml
```diff
@@ -12,27 +12,54 @@
...
+        docker build \
+          --progress=plain \
+          --file tools/docker/Dockerfile \
+          --target firecracker-containerd-test-naive \
+          --tag localhost/firecracker-containerd-test-naive:${BUILDKITE_BUILD_NUMBER} .
 
-  - label: ':hammer: snapshotter tests'
-    commands:
-      - "cd snapshotter"
-      - "DISABLE_ROOT_TESTS='true' make test EXTRAGOARGS='-v -count=1'"
```
samuelkarp: I'd like to see the non-root tests added back.  I think it's important that we verify that it's possible to disable root-required tests and still get meaningful results in CI.  This would be running a container without `--privileged`, without all the volumes, without `--ipc host` and running as a non-root user.

.buildkite/pipeline.yml
```diff
@@ -12,27 +12,54 @@
...
+  - label: ":docker: Build"
+    env:
+      DOCKER_BUILDKIT: "1"
+    command:
+      >
+        docker build \
+          --progress=plain \
+          --file tools/docker/Dockerfile \
+          --target firecracker-containerd-test-naive \
+          --tag localhost/firecracker-containerd-test-naive:${BUILDKITE_BUILD_NUMBER} .
```
samuelkarp: Why "naive" here?  We're testing more than the naive snapshotter...

tools/docker/firecracker-runtime.json
```diff
@@ -0,0 +1,13 @@
+{
+  "firecracker_binary_path": "/usr/local/bin/firecracker",
+  "socket_path": "./firecracker.sock",
```
samuelkarp: If you rebase, you can now remove the `socket_path` field.

tools/docker/Dockerfile
sipsma: No it must have just ended up there as I was modifying other things, agree it makes more sense in the previous block, fixed.

.buildkite/pipeline.yml
sipsma: The idea was that in this image we automatically start up the naive snapshotter; there could be other images we run e2e tests in that start up the devmapper snapshotter too.

I renamed this to `localhost/firecracker-containerd-e2etest-naive` and created another image called `localhost/firecracker-containerd-unittest` for the non-root tests, which might help clarify this a bit more?

tools/docker/firecracker-runtime.json
sipsma: Nice, updating

.buildkite/pipeline.yml
samuelkarp: Are you planning to have a separate image that starts the dm snapshotter?  Otherwise it might be better to name it as "e2etest" and run both snapshotters inside.

sipsma: Yeah having separate images per snapshotter was the idea (have a `# TODO` towards the bottom of the Dockerfile mentioning this as I did not implement the devmapper image at this time).

I didn't consider that we could just start up both snapshotters in the same image, but all else being equal I like the idea of having separate images for separate snapshotter implementations better just to prevent unnecessary dependencies (i.e. if I just want to run a test against the naive snapshotter I shouldn't have to necessarily also build the devmapper one).

.buildkite/pipeline.yml
sipsma: I see the original intention now, this was straightforward so I'll be updating with it.

.buildkite/pipeline.yml
samuelkarp: Maybe we should have a better selector in our tests for the snapshotters then, since right now it'll run all of the tests for both snapshotters?

sipsma: Yeah I was conflating some things here. The tests that will actually need the snapshotter to startup automatically don't exist yet (it's the test for the shim exit that is pending all this). I was just running these other tests in the one that automatically starts up the snapshotter out of convenience for now because it also needs the various `--privileged` and other flags to run anyways.

Either way, what I did now was just use two images `firecracker-containerd-unittest` and `firecracker-containerd-unittest-nonroot`. These are the only images we use right now; neither starts either snapshotter up automatically.

I am leaving the later stages that do start up the snapshotter automatically because they provide some utility immediately (by enabling manual debugging) and are going to be used in the very near future for the shim exit test.

Issue #59, opened 2018-12-27
Simplify the runtime configuration


samuelkarp:
The runtime configuration is fairly complicated and has a number of fields that need reasonable defaults.

* ~`firecracker_binary_path` - the relative path default is unreasonable, since the working directory changes every time containerd executes a new container (working directory is inside the container bundle)~
* [x] `socket_path` - needs a reasonable default, probably relative inside the working directory/bundle
* [x] `kernel_image_path` - could use a default lookup path
* [x] `kernel_args` - needs a reasonable default
* [x] `root_drive` - could use a default lookup path
* [x] `cpu_count` - might want to pick this dynamically from the OCI spec, with perhaps a reasonable default
* ~`console` - should it default to "stdio"?~

Related to https://github.com/firecracker-microvm/firecracker-containerd/issues/57

IRCody:
>socket_path - needs a reasonable default, probably relative inside the working directory/bundle

Not sure about all the others but I think `./firecracker.sock` makes sense as a default for socket_path.

samuelkarp:
> Not sure about all the others but I think ./firecracker.sock makes sense as a default for socket_path.

Yep, I think so too.

xibz:
I believe the jailer socket path (not settable) is `api.socket` and in firecracker's doc they use `firecracker.socket` every where. Should we try to be consistent? If the jailer is on by default, may be best to go with `api.socket`?

samuelkarp:
> `firecracker_binary_path` - the relative path default is unreasonable, since the working directory changes every time containerd executes a new container (working directory is inside the container bundle)

This has been changed in firecracker-go-sdk and we've now inherited that change such that it defaults to `firecracker` on `$PATH`.  This is a much more reasonable default.  I'll remove this item from the overview list.

samuelkarp:
> I believe the jailer socket path (not settable) is `api.socket` and in firecracker's docs they use `firecracker.socket` in their getting started guide. Should we try to be consistent? If the jailer is on by default, may be best to go with `api.socket`?

That [looks accurate](https://github.com/firecracker-microvm/firecracker/blob/v0.15.2/jailer/src/lib.rs#L386-L392).  I think for now I'm going to hard-code the socket (instead of making it configurable) as `./firecracker.sock`, but when we take a look at jailer integration (https://github.com/firecracker-microvm/firecracker-containerd/issues/122) we'll need to revisit this.

samuelkarp:
[...]

PR #139, opened 2019-03-19
runtime: hard-code Firecracker API socket path

samuelkarp:
*Issue #, if available:*
https://github.com/firecracker-microvm/firecracker-containerd/issues/59

*Description of changes:*
The Firecracker API socket is important for the runtime, as it uses the socket to communicate with Firecracker.  Allowing the runtime to fully control the API socket (including its path) removes a potential failure condition from misconfiguration.

The API socket is hard-coded to exist within the current working directory.  Part of the contract that containerd exposes for runtimes is that they are started with the current working directory changed to be that of the OCI bundle.  Using an API socket in the OCI bundle makes its location well-known and predictable to the runtime.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -212,10 +212,6 @@ configuration file has the following fields:
   `firecracker` located in its working directory.  A fully-qualified path to the
   `firecracker` binary is recommended, as the working directory typically
   changes every execution when run by containerd.
-* `socket_path` (required) - A path where a socket file should be created for
-  communicating with the Firecracker API.  A relative path like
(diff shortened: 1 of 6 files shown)


mxpv:
Would this require us to have separate runtime instance per microVM?

samuelkarp:
It does right now, as that's the way the runtime works today.  I know that we're thinking about changing the way the runtime works, but we don't yet know fully what we plan to do.  I think having the socket path defined in the global config file is bad as it makes it too easy to misconfigure and conflict.  If we do decide to have a single runtime, we can revisit this then and add it to whatever API is used to configure the runtime rather than a global config file.

PR #140, opened 2019-03-20
runtime: specify default kernel arguments

samuelkarp:
*Issue #, if available:*
https://github.com/firecracker-microvm/firecracker-containerd/issues/59

*Description of changes:*
Sets default kernel arguments so that it is not required to specify.

### DIFF
--- a/runtime/config_test.go
+++ b/runtime/config_test.go
@@ -1,0 +1,62 @@
+// Copyright 2018-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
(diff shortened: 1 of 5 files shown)


runtime/config_test.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func TestLoadConfigOverrides(t *testing.T) {
+	overrideKernelArgs := "OVERRIDE KERNEL ARGS"
+	configContent := fmt.Sprintf(`{"kernel_args":"%s"}`, overrideKernelArgs)
+	configFile, cleanup := createTempConfig(t, configContent)
+	defer cleanup()
+	cfg, err := LoadConfig(configFile)
+	if err != nil {
+		t.Error(err, "failed to load config")
```
mxpv: nit: can be just `assert.NoError`

PR #147, opened 2019-03-22
Implement usage for devmapper snapshotter

mxpv:
*Description of changes:*
This PR implements `Usage` for device mapper snapshotter.
It relies on the number of used blocks reported by `dmsetup status` command.

```
pool_device: 0 204800 thin-pool 0 13/25600 290/1600 - rw discard_passdown queue_if_no_space -
device3: 0 204800 thin 0 - <--- no file system
device2: 0 204800 thin 16128 204799 <---- just created with ext4 file system
device1: 0 204800 thin 18304 204799 <---- ext4 file system + 1 MB file
device1_snap: 0 204800 thin 20352 204799 <----- (parent device1) ext4 fs + 1 MB file from parent + 1 MB file extra (20352-18304)*512sectorSize/1024/1024=1MB
```

/ref https://github.com/containerd/containerd/pull/3022#issuecomment-475449586
/cc @cespo

### DIFF
--- a/snapshotter/devmapper/snapshotter.go
+++ b/snapshotter/devmapper/snapshotter.go
@@ -139,7 +139,22 @@ func (s *Snapshotter) Usage(ctx context.Context, key string) (snapshots.Usage, e
 		}
 
 		if info.Kind == snapshots.KindActive {
-			usage, err = s.getSnapshotUsage(ctx, id, &info)
+			deviceName := s.getDeviceName(id)
+			usage.Size, err = s.pool.GetUsage(deviceName)
+			if err != nil {
+				return err
+			}
+		}
+
+		if info.Parent != "" {
+			// GetInfo returns total number of bytes used by a snapshot (including parent).
+			// So subtract parent usage in order to get delta consumed by layer itself.
+			_, _, parentUsage, err := storage.GetInfo(ctx, info.Parent)
+			if err != nil {
+				return err
+			}
+
+			usage.Size -= parentUsage.Size
 		}
 
 		return err
@@ -206,44 +221,24 @@ func (s *Snapshotter) Commit(ctx context.Context, name, key string, opts ...snap
 	log.G(ctx).WithFields(logrus.Fields{"name": name, "key": key}).Debug("commit")
 
 	return s.withTransaction(ctx, true, func(ctx context.Context) error {
-		id, info, _, err := storage.GetInfo(ctx, key)
+		id, _, _, err := storage.GetInfo(ctx, key)
 		if err != nil {
 			return err
 		}
 
-		usage, err := s.getSnapshotUsage(ctx, id, &info)
+		deviceName := s.getDeviceName(id)
+		size, err := s.pool.GetUsage(deviceName)
 		if err != nil {
 			return err
+		}
+
+		usage := snapshots.Usage{
+			Size: size,
 		}
 
 		_, err = storage.CommitActive(ctx, key, name, usage, opts...)
 		return err
 	})
-}
-
-// getSnapshotUsage returns data size consumed by snapshot layer itself (excluing parent data usage)
-func (s *Snapshotter) getSnapshotUsage(ctx context.Context, id string, info *snapshots.Info) (snapshots.Usage, error) {
-	deviceName := s.getDeviceName(id)
-	size, err := s.pool.GetUsage(deviceName)
-	if err != nil {
-		return snapshots.Usage{}, err
-	}
-
-	// Subtract parent data usage if any
-	if info.Parent != "" {
-		_, _, usage, err := storage.GetInfo(ctx, info.Parent)
-		if err != nil {
-			return snapshots.Usage{}, err
-		}
-
-		size -= usage.Size
-	}
-
-	usage := snapshots.Usage{
-		Size: size,
-	}
-
-	return usage, nil
 }
 
 // Remove removes thin device and snapshot metadata by key
--- a/snapshotter/devmapper/snapshotter.go
+++ b/snapshotter/devmapper/snapshotter.go
@@ -121,11 +121,31 @@ func (s *Snapshotter) Update(ctx context.Context, info snapshots.Info, fieldpath
 	return info, err
 }
 
-// Usage not yet implemented
+// Usage returns the resource usage of an active or committed snapshot excluding the usage of parent snapshots.
 func (s *Snapshotter) Usage(ctx context.Context, key string) (snapshots.Usage, error) {
 	log.G(ctx).WithField("key", key).Debug("usage")
 
-	return snapshots.Usage{}, errors.New("usage not implemented")
+	var (
+		id    string
+		err   error
+		info  snapshots.Info
+		usage snapshots.Usage
+	)
+
+	err = s.withTransaction(ctx, false, func(ctx context.Context) error {
+		id, info, usage, err = storage.GetInfo(ctx, key)
+		if err != nil {
+			return err
+		}
+
+		if info.Kind == snapshots.KindActive {
+			usage, err = s.getSnapshotUsage(ctx, id, &info)
+		}
+
+		return err
+	})
+
+	return usage, err
 }
 
 // Mounts return the list of mounts for the active or view snapshot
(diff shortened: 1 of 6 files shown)


snapshotter/devmapper/snapshotter.go
```diff
@@ -125,7 +125,27 @@ func (s *Snapshotter) Update(ctx context.Context, info snapshots.Info, fieldpath
 func (s *Snapshotter) Usage(ctx context.Context, key string) (snapshots.Usage, error) {
```
xibz: comment is not longer valid, correct?

snapshotter/devmapper/snapshotter.go
```diff
@@ -125,7 +125,27 @@ func (s *Snapshotter) Update(ctx context.Context, info snapshots.Info, fieldpath
...
+	)
+
+	err = s.withTransaction(ctx, false, func(ctx context.Context) error {
+		id, info, usage, err = storage.GetInfo(ctx, key)
+		if err != nil {
+			return err
+		}
+
+		if info.Kind == snapshots.KindActive {
+			usage, err = s.getSnapshotUsage(ctx, id, &info)
```
xibz: Can you explain why we are stomping over the previous usage?

snapshotter/devmapper/snapshotter.go
```diff
@@ -186,11 +206,46 @@ func (s *Snapshotter) Commit(ctx context.Context, name, key string, opts ...snap
 	log.G(ctx).WithFields(logrus.Fields{"name": name, "key": key}).Debug("commit")
 
 	return s.withTransaction(ctx, true, func(ctx context.Context) error {
-		_, err := storage.CommitActive(ctx, key, name, snapshots.Usage{}, opts...)
+		id, info, _, err := storage.GetInfo(ctx, key)
```
xibz: And why is this usage not important? Is it assumed that the snapshot will always be active at this point?

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -258,6 +258,55 @@ func Version() (string, error) {
...
+		err    error
+		status DeviceStatus
+	)
+
+	output, err := dmsetup("status", deviceName)
+	if err != nil {
+		return nil, err
+	}
+
+	output = strings.Trim(output, "\n")
```
xibz: Doesn't `dmsetup` already trim `\n`? Is there multiple expected new lines?

snapshotter/devmapper/snapshotter.go
mxpv: `KindActive` means that this layer is currently being open for modifications, so `Usage` must return the most up to date info

snapshotter/devmapper/snapshotter.go
mxpv: Because we don't keep usage info for active snapshots in database.

snapshotter/pkg/dmsetup/dmsetup.go
mxpv: `dmsetup status` can output multiple lines. Here I want to make sure that the last parameter doesn't contain `\n` at the end

xibz: I think any parsing of the `dmsetup` function should be contained there. How do you feel about moving this in `dmsetup` with an additional if statement?

```go
if action == "status" {
	output = strings.Trim(output, "\n")
}
```

mxpv: Actually `dmsetup` already [contains](https://github.com/firecracker-microvm/firecracker-containerd/blob/81542ccb06325417a21faa19dd0ca87ecd4d9b31/snapshotter/pkg/dmsetup/dmsetup.go#L343) `strings.TrimSuffix(output, "\n")`. It makes sense to replace it with `strings.Trim(output, "\n")`

xibz: @mxpv - Yea, that's why I was confused :D

mxpv: Done

snapshotter/devmapper/snapshotter.go
```diff
@@ -186,11 +206,46 @@ func (s *Snapshotter) Commit(ctx context.Context, name, key string, opts ...snap
...
+// getSnapshotUsage returns data size consumed by snapshot layer itself (excluing parent data usage)
+func (s *Snapshotter) getSnapshotUsage(ctx context.Context, id string, info *snapshots.Info) (snapshots.Usage, error) {
+	deviceName := s.getDeviceName(id)
+	size, err := s.pool.GetUsage(deviceName)
+	if err != nil {
+		return snapshots.Usage{}, err
+	}
+
+	// Subtract parent data usage if any
+	if info.Parent != "" {
```
samuelkarp: Do you need to traverse through all of the ancestors?  It looks like this just accounts for the immediate parent.

snapshotter/pkg/dmsetup/dmsetup.go
```diff
@@ -258,6 +258,53 @@ func Version() (string, error) {
...
+}
+
+// Status provides status information for devmapper device
+func Status(deviceName string) (*DeviceStatus, error) {
+	var (
+		err    error
+		status DeviceStatus
+	)
+
+	output, err := dmsetup("status", deviceName)
```
samuelkarp: Do you need "--noflush"?

snapshotter/devmapper/snapshotter.go
mxpv: No, we need a difference between parent/child here. Have a look on `device1_snap` in PR description.

mxpv: Discussed offline, fixed in https://github.com/firecracker-microvm/firecracker-containerd/pull/147/commits/693f1f5d9a2fa7977cde92e314c999947673b0ae

snapshotter/pkg/dmsetup/dmsetup.go
samuelkarp: Discussed with @mxpv offline.  We're both not sure whether we want `--noflush`, but we can leave it off for now.

snapshotter/devmapper/snapshotter.go
```diff
@@ -186,7 +219,22 @@ func (s *Snapshotter) Commit(ctx context.Context, name, key string, opts ...snap
...
 
 	return s.withTransaction(ctx, true, func(ctx context.Context) error {
-		_, err := storage.CommitActive(ctx, key, name, snapshots.Usage{}, opts...)
+		id, _, _, err := storage.GetInfo(ctx, key)
+		if err != nil {
+			return err
+		}
+
+		deviceName := s.getDeviceName(id)
+		size, err := s.pool.GetUsage(deviceName)
```
samuelkarp: This is a `dmsetup` disk-based operation happening inside a transaction.  Can you split this into two separate transactions instead so that we don't have anything calling out into devicemapper in the middle of a transaction?

snapshotter/devmapper/snapshotter.go
```diff
@@ -121,11 +121,44 @@ func (s *Snapshotter) Update(ctx context.Context, info snapshots.Info, fieldpath
...
+		if info.Kind == snapshots.KindActive {
+			deviceName := s.getDeviceName(id)
+			usage.Size, err = s.pool.GetUsage(deviceName)
+			if err != nil {
+				return err
+			}
+		}
+
+		if info.Parent != "" {
+			_, _, parentUsage, err := storage.GetInfo(ctx, info.Parent)
```
samuelkarp: Can you add a comment here that `storage.GetInfo` returns the total number of bytes consumed by the parent (i.e., not just the delta of that layer) so that it's clear that we're working with all of the data?

snapshotter/devmapper/snapshotter.go
```diff
@@ -121,11 +121,44 @@ func (s *Snapshotter) Update(ctx context.Context, info snapshots.Info, fieldpath
...
+
+	err = s.withTransaction(ctx, false, func(ctx context.Context) error {
+		id, info, usage, err = storage.GetInfo(ctx, key)
+		if err != nil {
+			return err
+		}
+
+		if info.Kind == snapshots.KindActive {
+			deviceName := s.getDeviceName(id)
+			usage.Size, err = s.pool.GetUsage(deviceName)
```
samuelkarp: This is a `dmsetup` disk-based operation happening inside a transaction. Can you split this into two separate transactions instead so that we don't have anything calling out into devicemapper in the middle of a transaction?

snapshotter/devmapper/snapshotter.go
mxpv: `dmsetup status` is not going to be heavy disk operation. Essentially it ends up with reading the corresponding value from btree leaf ([1](https://github.com/torvalds/linux/blob/cffd425b90147296c627dfb13f7ea8f361d962b2/drivers/md/dm-thin-metadata.c#L1825), [2](https://github.com/torvalds/linux/blob/6cdc577a18a616c331f57e268c97466171cfc45f/drivers/md/dm-thin.c#L4387), [3](https://github.com/torvalds/linux/blob/6cdc577a18a616c331f57e268c97466171cfc45f/drivers/md/dm-init.c#L264), [4](https://github.com/torvalds/linux/blob/6cdc577a18a616c331f57e268c97466171cfc45f/drivers/md/dm-thin.c#L53)). I don't think it worth to optimize here, we'll get more noise than performance gain.

snapshotter/devmapper/snapshotter.go
mxpv: Done

snapshotter/devmapper/snapshotter.go
samuelkarp: Let's keep an eye on it going forward.  I'm concerned about the combination of the lock plus skipping `--noflush`, but we can move forward and keep this in mind if we experience issues in the future.

PR #148, opened 2019-03-25
agent: fixes panic on cancelAll function

xibz:
`cancelAll` would panic due to using `len(slice)` as an index.


sipsma:
Change itself looks good but we should update the makefiles and/or .`buildkite/pipeline.yml` to run the new unit test (it isn't now because this is the first unit test in the agent package).

xibz:
@sipsma - Good catch. Will update.

PR #150, opened 2019-03-28
Add design overview doc

mxpv:


docs/design-approaches.md
```diff
@@ -0,0 +1,235 @@
...
+A few details on Orchestrator / Runtime responsibilities:
+
+* Runtime/FC control manages Firecracker microVMs lifecycle e.g. creates new VM instances for new create container requests, knows how to reuse existing VMs, and decides when to shutdown instances.
+* Orchestrator uses a `VMID` for each container group and includes it in requests to Runtime to identify which microVM instance to use for serving a request. Runtime is responsible for matching given `VMID` with corresponding VM instance.
+* Orchestrator is able to get running microVM instances and query specific VM through containerd's service layer (get metrics, set metadata, get logs, etc) by `VMID`.
+* Runtime is responsible for tracking containers states and for VM termination if there are no running containers left.
+* Orchestrator is responsible for pulling container images via `containerd` client.
+
+
+## Runtime behavior
```
sipsma: Between our two PRs, whoever merges second should probably just link this Runtime section and the Agent section to the shim-design doc for more context on the final decision that was made

mxpv: Yep. I'll update this section with ref to your doc as soon as it gets merged.

docs/design-approaches.md
```diff
@@ -0,0 +1,235 @@
...
+
+
+## Goals
+
+Support running multiple containers inside a microVM in order to handle Orchestrator tasks with multiple container definitions.
+
+
+## Requirements
+
+1. Orchestrator needs to launch multiple containers in the same VM. The container launch can be staggered. For example, requests could be something like: Launch container A in a VM. After a period of time (once the orchestrator verifies that container A is healthy), launch container B in the same VM.
```
samuelkarp: nit: I tend to break lines in Markdown around 80 characters as it makes it easier to read in plain-text (doesn't cross the whole screen).  But we're not enforcing that anywhere so it's up to you.

docs/design-approaches.md
```diff
@@ -0,0 +1,235 @@
...
+2. containerd uses firecracker-containerd runtime and block device snapshotter. Requests are passed to runtime. Runtime is responsible for communicating with Agent, attaching/managing block devices and controlling container execution lifecycle.
+3. Runtime controls the list of drives to use for passing container mounts (more on this in the following sections).
+4. Runtime and/or FC-control [finds](https://github.com/firecracker-microvm/firecracker-containerd/blob/7fd9dcfb77c6b5e760ed0a8b2e1eecf44fe37c20/runtime/service.go#L560) available context [identifier](http://man7.org/linux/man-pages/man7/vsock.7.html) for Vsock and starts the VM (unless it's already started), waits for the VM to boot up and Agent to start responding to requests via vsock channel. After that the Runtime submits requests for managing containers through Agent (interface is described below).
```
samuelkarp: I think finding a CID is not high level and is subject to change depending on how Firecracker implements vsock.

docs/design-approaches.md
samuelkarp: I think you should just link to Erik's doc.

docs/design-approaches.md
```diff
@@ -0,0 +1,235 @@
...
+
+**Pros**:
+* GRPC/TTRPC can generate strongly typed interface for data exchange (see service example below)
+* No need in port negotiation, Agent's well-known port always used.
+
+**Cons**:
+* TTRPC doesn't support streaming (need to implement and submit a PR)
+* GRPC is heavier, potentially it consumes more resources ([need to clarify how much heavier it'll be for just one service](https://github.com/xibz/GRPCvsTTRPC/blob/master/README.md)).
+* Multiplexing (in general) may consume more memory & CPU
+
```
samuelkarp: There's a third option here, which is multiplexing at the connection layer instead of the application layer (i.e., underneath gRPC instead of within gRPC).  From our discussion with the Firecracker team, they are looking into options around Firecracker handling some aspects of the communication (i.e., exposing the correct vsock device inside the VM) and the outer application needing to handle other aspects (i.e., (de)multiplexing vsock datagrams from an AF_UNIX socket on the host.

mxpv:
> This doc also does not describe any recommendations

I think we don't have enough information to provide recommendations yet. For instance block devices and multiplexing depend on FC team explorations outcome.

docs/design-approaches.md
```diff
@@ -0,0 +1,255 @@
...
+  needs. It's where we can decide vm size, etc.
+* VMID - abstract identifier not related to any specific implementation. Means a group of containers somehow related by
+  same id (for example a group of containers from pod or task definition)
+
+
+## High-level overview
+
+![design-overview](img/design-overview.png)
+
+1. Orchestrator accepts task definition request from data-plane and breaks it into sub-requests suitable for sending to
```
samuelkarp: nit: you probably mean "from control-plane" instead of "from data-plane".

PR #151, opened 2019-03-28
Add GRPC plugin stub

mxpv:
*Description of changes:*
This PR adds:
- Proper protobuf generator Makefile and initial `fc-control` service definition
- containerd binary with embedded GRPC plugin (to simplify development process and avoid headache with golang plugins).


proto/sandbox.proto
```diff
@@ -0,0 +1,57 @@
...
+
+    // Stops existing Firecracker instance by VM ID
+    rpc StopVM(StopVMRequest) returns (google.protobuf.Empty);
+
+    // Gets VM instance socket file location
+    rpc GetVMAddress(GetVMAddressRequest) returns (GetVMAddressResponse);
+}
+
+// CreateVMRequest specifies creation parameters for a new FC instance
+message CreateVMRequest {
```
sipsma: (This is just a general question, not something that should probably be addressed in this CR)

Do you think there's any possibility of sharing the struct definitions here with the similar ones used in the firecracker-go-sdk? On the surface, it seems like there's a lot of overlap.

Just sharing definitions would probably be beneficial, but even better would be if we could just take the firecracker-go-sdk and point it at fc-control and start a VM up.

The main issue I can think of would be that the go-sdk doesn't currently have a notion of VMID, though I wonder if it would be much effort to add a layer there that incorporates it in a way such that it's compatible with both the firecracker socket and fc-control. Maybe that's easier said than done though, just curious what your thoughts are.

mxpv: There is an overlap and this is why we have to keep [taskfirecrackeropts.go](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/runtime/taskfirecrackeropts.go) to convert from one to another (not super happy with it either). SDK generates models from swagger file (pulled from FC repo), while here we generate from protobuf definition in order to use GRPC. 
I definitely be happy to avoid this overlap. We might want to consider generating protobuf from swagger as well, so we don't need to duplicate lots of contracts as one possible solution.

proto/sandbox.proto
```diff
@@ -0,0 +1,57 @@
...
+    // The number of dummy drives to reserve in advance before running FC instance.
```
samuelkarp: I'd vote for different ordering here, since we're defining now.

```protobuf
FirecrackerMachineConfiguration MachineCfg = 1;
string KernelImagePath = 2;
string KernelArgs = 3;
FirecrackerDrive RootDrive = 4;
repeated FirecrackerDrive AdditionalDrives 5;
repeated FirecrackerNetworkInterface NetworkInterfaces = 6;
int32 ContainerCount = 7;
```

That way we're grouping all of the explicit VM configuration first and adding the container-oriented components at the end.

sandbox/cmd/containerd/Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+SRC := $(shell find . -name '*.go')
+GOMOD := $(shell go env GOMOD)
+GOSUM := $(GOMOD:.mod=.sum)
+
+all: build
+containerd: build
+
+build: $(SRC) $(GOMOD) $(GOSUM)
+	go build -v -o containerd
```
samuelkarp: Would it be worth calling this binary "firecracker-containerd" to make it explicit that it's not the standard containerd binary?

proto/sandbox.proto
mxpv: Another another option: handle it similarly to how `containerd` handles OCI specs: [1](https://github.com/containerd/containerd/blob/932f883e49af925380a3fdce6b8db189915f93fb/containers/containers.go#L56) and [2](https://github.com/containerd/containerd/blob/932f883e49af925380a3fdce6b8db189915f93fb/container_opts.go#L216) (e.g. pack sdk's structs to `type.Any`)

firecracker-control/cmd/containerd/Makefile
```diff
@@ -0,0 +1,36 @@
...
+EXTRAGOARGS:=
+
+SRC := $(shell find . -name '*.go')
+GOMOD := $(shell go env GOMOD)
+GOSUM := $(GOMOD:.mod=.sum)
+
+all: build
+containerd: build
+
+build: $(SRC) $(GOMOD) $(GOSUM)
```
samuelkarp: Can you rewrite it like this?

```makefile
build: firecracker-containerd
firecracker-containerd: $(SRC) $(GOMOD) $(GOSUM)
	go build -v -o firecracker-containerd
```

firecracker-control/cmd/containerd/Makefile
```diff
@@ -0,0 +1,36 @@
...
+install: containerd
+	cp firecracker-containerd /usr/local/bin/
+
+test: $(SOURCES)
+	go test ./... $(EXTRAGOARGS)
+
+clean:
+	- rm -f firecracker-containerd
+
+.PHONY: all containerd clean install
```
samuelkarp: build and test should be added here too

firecracker-control/cmd/containerd/README.md
```diff
@@ -0,0 +1,4 @@
+# README
+
+This is custom `containerd` binary which embeds `firecracker-control` GRPC plugin (instead of using it as golang
```
samuelkarp: nit: "a custom `containerd` binary", "embeds the `firecracker-control` GRPC plugin", "instead of using it as a Go plugin", "simplify the development process"

PR #155, opened 2019-04-05
runtime: add some debug logs

samuelkarp:

### DIFF
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -120,6 +120,7 @@ func NewService(ctx context.Context, id string, remotePublisher events.Publisher
 }
 
 func (s *service) StartShim(ctx context.Context, id, containerdBinary, containerdAddress string) (string, error) {
+	log.G(ctx).WithField("id", id).Debug("StartShim")
 	cmd, err := s.newCommand(ctx, containerdBinary, containerdAddress)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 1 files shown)


runtime/service.go
```diff
@@ -201,6 +217,7 @@ func (s *service) Create(ctx context.Context, request *taskAPI.CreateTaskRequest
 	}
 	// TODO: should there be a lock here
 	if !s.agentStarted {
+		log.G(ctx).WithField("id", request.ID).Debug("calling startVM")
 		client, err := s.startVM(ctx, request, firecrackerConfig)
 		if err != nil {
 			log.G(ctx).WithError(err).Error("failed to start VM")
```
sipsma: nit: can also include the request.ID here too now

runtime/service.go
```diff
@@ -251,17 +268,18 @@ func (s *service) proxyStdio(ctx context.Context, stdin, stdout, stderr string,
...
 	f, err := fifo.OpenFifo(ctx, path, syscall.O_RDWR|syscall.O_NONBLOCK, 0700)
 	if err != nil {
-		log.G(ctx).WithError(err).Error("error opening fifo")
+		log.G(ctx).WithError(err).WithField("path", path).Error("error opening fifo")
 		return
 	}
 	conn, err := vsock.Dial(cid, port)
 	if err != nil {
-		log.G(ctx).WithError(err).Error("unable to dial agent vsock")
+		log.G(ctx).WithError(err).WithField("path", path).Error("unable to dial agent vsock")
```
sipsma: nit: could include something like `unable to dial agent vsock IO port` to distinguish from the other vsock dial we do

runtime/service.go
samuelkarp: Done

runtime/service.go
samuelkarp: Done

PR #158, opened 2019-04-08
docs: Host file isolation

samuelkarp:
Add a new design document on ensuring that host files remain isolated
between multiple microVMs.

### DIFF
--- a/docs/host-file-isolation.md
+++ b/docs/host-file-isolation.md
@@ -1,0 +1,146 @@
+# Host file isolation
+
+firecracker-containerd has a number of host-based components, based on files,
+that are accessible to a microVM and need to be isolated.  Isolation means that
+one microVM does not have the ability to influence the execution of another
+microVM or to gain access to data held by another microVM.  We believe this is
+important both for security (i.e., one VM tenant cannot influence the execution
+of another tenant's VM) and for repeatability (my container runs the same way
+every time).
+
+## File components
+
+### Kernel
+
+The kernel is the bootable component of the microVM.  We expect that users of
+firecracker-containerd will either use the same kernel for all microVMs, or have
+a small number of kernels shared among multiple microVMs.  Importantly, we do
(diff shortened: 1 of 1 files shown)


docs/host-file-isolation.md
```diff
@@ -0,0 +1,142 @@
...
+### Root filesystem
+
+Similar to the kernel, the root filesystem contains the user-mode components
+that are invoked upon boot of a microVM.  Like the kernel, we expect users of
+firecracker-containerd to either use the same root filesystem for all microVMs
+or share a small number of root filesystems among multiple microVMs.  Also like
+the kernel, we do not want to require a user of firecracker-containerd to
+provide separate root filesystems for each microVM.
+
+Because the root filesystem is shared among multiple microVMs, it must not be
```
sipsma: I think this can be clarified to be that while the mounted rootfs may be mutated by processes running in the Guest, any changes are ephemeral for the life of the VM and will not affect other VMs using the same base rootfs

docs/host-file-isolation.md
```diff
@@ -0,0 +1,142 @@
...
+Overlay has broad support as it is integrated with the Linux kernel and
+Firecracker already requires a new-enough kernel.  Overlay is also simple to set
+up.  We are already using devicemapper as our snapshotter, even though it is
+more challenging to set up.  Filesystem-integrated copy-on-write approaches
+require a filesystem that supports the feature; we do not currently require a
+specific filesystem to run firecracker-containerd.
+
+### Preventing mutation
+
+Linux provides different capabilities for preventing mutation to files:
```
sipsma: You could add "mounting filesystem as read-only" to this list I think

samuelkarp: Done

docs/host-file-isolation.md
samuelkarp: Done

PR #160, opened 2019-04-09
Group containers into different VMs by VMID

sipsma:
1. This is the first half of the shim updates needed for running multiple containers per-VM. It consists of the host-side changes needed to group containers up into separate VMs by VMID. The second part of the updates are going to be VM-side in `agent`; they will be a separate PR from this one.
   * Right now, multiple VMs can be spun up in parallel each with it's own container, but the above mentioned changes in `agent` will be needed to support multiple containers within each VM.
1. Many of the TODOs added will be removed by integrating with Max's fc-control plugin (#156).
1. There's a few small updates that will be needed to the shim design doc once this is approved; namely
   * the use of just the `address` file instead of that plus the pid file (this is also what containerd has done in their newer runc v2 shim)
   * the "start shim" routine will fork/exec a child when it creates a new shim for handling a new VM (it turns out containerd forces you to do this; you can't just let the existing process take over even though it's the same binary...)
1. The original strategy for the e2e test was to use MMDS. This worked but when it came time to productionizing it, there were a lot of issues to be sorted out with getting a container in the VM w/ curl installed, making the timing of the MMDS update request and container query of it work, etc. I realized a different approach of just mapping the VMID to a mac address and having the container print that was much simpler but just as robust a test, so I switched to that.

### DIFF
--- a/runtime/firecrackeroci/annotation.go
+++ b/runtime/firecrackeroci/annotation.go
@@ -1,0 +1,39 @@
+// Copyright 2018-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package firecrackeroci
+
+import (
+	"context"
+
+	"github.com/containerd/containerd/containers"
+	"github.com/containerd/containerd/oci"
+)
+
+const (
+	// VMIDAnnotationKey is the key specified in an OCI-runtime config annotation section
+	// specifying the ID of the VM in which the container should be spun up.
+	VMIDAnnotationKey = "aws.firecracker.vm.id"
+)
+
+// WithVMID annotates a containerd client's container object with a given firecracker VMID.
+func WithVMID(vmID string) oci.SpecOpts {
+	return func(_ context.Context, _ oci.Client, _ *containers.Container, s *oci.Spec) error {
+		if s.Annotations == nil {
+			s.Annotations = make(map[string]string)
+		}
+
+		s.Annotations[VMIDAnnotationKey] = vmID
+		return nil
+	}
+}
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -15,14 +15,23 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"math"
+	"strings"
+
+	// disable gosec check for math/rand. We just need a random starting
+	// place to start looking for CIDs; no need for cryptographically
+	// secure randomness
+	"math/rand" // #nosec
+
 	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sync"
 	"syscall"
 	"time"
 	"unsafe"
@@ -39,6 +48,7 @@ import (
 	"github.com/containerd/typeurl"
 	"github.com/firecracker-microvm/firecracker-go-sdk"
 	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
+	"github.com/gofrs/uuid"
 	ptypes "github.com/gogo/protobuf/types"
 	"github.com/mdlayher/vsock"
 	"github.com/pkg/errors"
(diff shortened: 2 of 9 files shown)


runtime/service.go
```diff
@@ -48,20 +57,72 @@ import (
...
+	// to find the shim addr file in the bundle dir, so we still have to create it or
+	// symlink it to the shimAddrFilePath
+	bundleAddrFilePath string
+
+	// shimLogFifoPath is a path to a FIFO for writing shim logs as found in the vmIDDir.
+	shimLogFifoPath string
+
+	// bundleLogFifoPath is a path to a FIFO for writing shim logs as found in the bundleDir.
+	// It is the path created by containerd for us, the shimLogFifoPath is just a symlink to one.
+	bundleLogFifoPath string
```
mxpv: It looks like `bundleLogFifoPath` and `shimLogFifoPath` are only required for start up logic (specifically creating symlinks). Can we evaluate those just once inside `createSymlinkXXX` and don't keep them on instance level?

runtime/service.go
```diff
@@ -48,20 +57,72 @@ import (
...
 // implements shimapi
 type service struct {
 	server        *ttrpc.Server
 	id            string
+	vmID          string
 	publish       events.Publisher
 	eventExchange *exchange.Exchange
 
+	// vmIDDir holds files, sockets and FIFOs scoped to a single given VM
+	vmIDDir string
```
mxpv: nit: vmIDDir -> vmDir ?

sipsma: Yeah agree that makes more sense, will update

runtime/service.go
sipsma: We can, the reason I wanted them on the instance level is that it centralizes our definition of paths in one place (which helps a lot for clarity I think) and in case they ever need to be used in more than one place in the code, they will already be accessible.

Given the cost of storing 2 strings is minimal and they are expected to be constant for the lifetime of the service, I'd prefer to keep them in the instance level unless there's other considerations I'm not thinking of, let me know.

mxpv: It's more about readability/maintainability. I would rather avoid polluting struct definitions with fields that are not going to be used. In case when we need to access in more than one place, I would rather use `getXxxPath` instead (but in this case its only one usage).

sipsma: How do you feel about updating to use `get*Path` methods for all the paths? I didn't do this originally because it felt silly to have a function for returning a value that's a known constant after initialization, but in retrospect it does make those values more immutable, which is nice.

I still don't want to just inline any paths inside the function where they are used because it makes it hard for someone unfamiliar with the code to know what the overall directory structure is and all the paths the code interacts with. They have to search all over the place rather than look in one central place where all paths are defined. This is something that's frustrated me on multiple occasions when reading various other codebases.

runtime/firecrackeroci/annotation.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+
+	"github.com/containerd/containerd/containers"
+	"github.com/containerd/containerd/oci"
+)
+
+const (
+	// VMIDAnnotationKey is the key specified in an OCI-runtime config annotation section
+	// specifying the ID of the VM in which the container should be spun up.
+	VMIDAnnotationKey = "aws.firecracker.vmID"
```
samuelkarp: I think annotations are meant to be DNS-style, so all lowercase.  Maybe "aws.firecracker.vm-id" or "aws.firecracker.vm.id"?

Also, per https://github.com/containerd/containerd/pull/3004 there will be support for `io.containerd.runc.v2.group=<group>` and `io.kubernetes.cri.sandbox-id=<group>` to indicate grouping in containerd.  Do we want to try using those now, or start off with our own annotation and migrate to this later?

runtime/service.go
mxpv: yep, `get*Path` is fine.

runtime/service.go
```diff
@@ -48,20 +57,72 @@ import (
...
+
+	// bundleDir is created by containerd and contains configuration scoped to a
+	// single given container
+	bundleDir string
+
+	// shimAddrFilePath is the path to the shim address file as found in the vmIDDir
```
samuelkarp: What is the address file?  A text file containing the path to the AF_UNIX socket for the shim?

runtime/service.go
```diff
@@ -76,8 +137,78 @@ var (
...
+
+	return cwd, nil
+}
+
+// If the VMID was provided by the client, we'll find it in the OCI config Annotations section.
+// If, however, no VMID was provided in the OCI config, we it may have been set by a "shim start"
+// parent process as an env var. resolveVMID checks those two locations and returns the VMID
+// if found in either.
+func resolveVMID(ctx context.Context) (string, error) {
+	// If the vmID is provided via env var, use that
```
samuelkarp: What's setting the environment variable?

runtime/service.go
```diff
@@ -48,20 +57,72 @@ import (
...
+
+	// shimSocketAddress is the abstract unix socket path at which our shim serves its API
+	// It is unique per-VM.
+	shimSocketAddress string
+
+	firecrackerSockPath        string
+	firecrackerLogFifoPath     string
+	firecrackerMetricsFifoPath string
+
+	startVMMutex sync.Mutex
```
samuelkarp: Is it worth replacing `startVMMutex` and `agentStarted` with a `sync.Once`?

runtime/firecrackeroci/annotation.go
sipsma: Ah good call, will update this to `aws.firecracker.vm.id` for now.

As for whether we should re-use the keys defined in containerd's implementation instead, that's an interesting question.

On one hand, it would be nice to automatically group containers into VMs using these already defined annotations so users can switch to firecracker-containerd more seamlessly.

On the other hand, I don't know if we want to entangle ourselves and our users with keys that have meaning out of our control. I.e. what if some other random part of a user's stack starts also interpreting `io.kubernetes.cri.sandbox-id` or `io.containerd.runc.v2.group` but does something the user doesn't want? That user would probably prefer to just have a key like `aws.firecracker.vm.id` so they can specify the firecracker VM without any extra side-effects.

Maybe what'd be best is for our shim implementation to interpret each one with some priority, i.e. use `aws.firecracker.vm.id` if set, if not check `io.kubernetes.cri.sandbox-id`, if not check another, etc.

I'll make an issue for following up on this.

sipsma: https://github.com/firecracker-microvm/firecracker-containerd/issues/162

PR #161, opened 2019-04-10
Updater linters

mxpv:
*Description of changes:*
- Update golangci-lint to v1.16.0
- Add a few new linters
- Use `bin` directory for `deps` dependencies (don't pollute `GOPATH/bin`)


.golangci.yml
```diff
@@ -19,6 +19,9 @@ linters:
     - prealloc
     - scopelint
     - gocritic
+    - interfacer # Suggests narrower interface types
+    - gocyclo
```
xibz: This linter seems a little ambiguous. What number are we looking at before we get a linting error in terms of complexity?

mxpv: Ambiguous why?
it uses abstract "units" https://github.com/golangci/golangci-lint/blob/97ea1cbb21bbf5e4d0e8bcc0f9243385e9262dcc/README.md#config-file

xibz: Ah, I didn't know it defaulted to 30. 30 seems ridiculously high and I don't think we'd ever hit it. Im assuming we can override these values, can we settle for something like 15?

mxpv: Done.
Agent's `main` need to be refactored now, that will be addressed later.

xibz: I thought about this some more and I am convinced that number is useless. Well, cyclomatic complexity is useless.

```go
for i := 0; i < 13; i++ {
    month := rand.Intn(15)
    switch(month) {
        case 1:
             fmt.Println("Jan")
        case 2:
             fmt.Println("Feb")
        case 3:
             fmt.Println("Mar"")
        case 4:
            fmt.Println("April"")
        case 5:
             fmt.Println("May"")
        case 6:
             fmt.Println("June"")
        case 7:
             fmt.Println("July"")
        case 8:
             fmt.Println("Aug"")
        case 9:
             fmt.Println("Sep"")
        case 10:
             fmt.Println("Oct"")
        case 11:
             fmt.Println("Nov"")
        case 12:
             fmt.Println("Dec"")
        default:
             fmt.Println("None"")
    }
}
```
The cyclomatic complexity here is 15. However, I can write something that has a complexity of 3 but I feel like this is a matter of preference and I think the numbers don't really do a good job of giving any useful information other than how many branches a function has.

```go
months := []string{
"Jan",
"Feb",
"Mar",
"April",
"May",
"June",
"July",
"Aug",
"Sept",
"Oct",
"Nov",
"Dec",
}

for i := 0; i < 13; i++ {
    month := rand.Intn(13)
    if month >= len(months) {
        fmt.Println("None")
        continue
    }
    fmt.Println(months[month])
}
```

mxpv: In general splitting big function into smaller one isn't a bad practice and we can turn it off anytime.

samuelkarp: I think I agree with Ben.  This is the kind of thing that we should be looking at during code review; an automated tool doesn't have the same judgement as a human.

mxpv: Ok, deleted.

Issue #159, opened 2019-04-09
Implement firecracker-control GRPC plugin


mxpv:
Implement a GRPC plugin for containerd to manage Firecracker VM instance(s) 

- [x] Basic implementation: https://github.com/firecracker-microvm/firecracker-containerd/pull/156
- [x] Generate containerd events (vm start, vm exit)
- [x] Remove TODOs after merging PRs: firecracker-microvm/firecracker-go-sdk#91, firecracker-microvm/firecracker-go-sdk#90
- [x] Handle containerd shutdown properly? (shutdown FC instances)
- [x] Add collision detection for generated IDs.
- [x] Figure out how to reuse containerd Client conn object with grpc plugins - https://github.com/containerd/containerd/pull/3208
- [x] Address synchronous VM creation - https://github.com/firecracker-microvm/firecracker-containerd/pull/167
- [x] Add fake devices reservation (copy-paste from Runtime): #154
- [x] VM configuration defaults.
- [x] ~Implement persistency. containerd is designed to be able to stop and restart without the containers under management being interrupted. We  want to persist state outside of memory so we can recover~
- [x] Throw away obsolete code from Runtime (https://github.com/firecracker-microvm/firecracker-containerd/pull/182)
- [x] Switch to TTRPC (https://github.com/firecracker-microvm/firecracker-containerd/issues/180, https://github.com/firecracker-microvm/firecracker-containerd/pull/196)
- [x] Update example to use GRPC plugin (https://github.com/firecracker-microvm/firecracker-containerd/pull/182)

PR #163, opened 2019-04-15
firecracker-control improvements

mxpv:
*Issue #, if available:*
https://github.com/firecracker-microvm/firecracker-containerd/issues/159

*Description of changes:*
This PR addresses some comments from https://github.com/firecracker-microvm/firecracker-containerd/pull/156 + adds containerd events.

Changelist:
- Use vsock file descriptor to lock cid until FC starts (in order to reduce races)
- Client is responsible for supplying VM id (so now you can implement scenario `subscribe topic==ID` + `create VM with ID` + `catch VM events`)
- Use SDK's machine interface for mocking (instead of maintaining our own)
- Publish containerd events (VM start/stop)


firecracker-control/local.go
```diff
@@ -48,25 +45,31 @@ func init() {
...
 
-// TODO: Remove this interface after merging PR #91
-// See https://github.com/firecracker-microvm/firecracker-go-sdk/pull/91
-type machineIface interface {
-	Start(context.Context) error
-	StopVMM() error
-	Wait(context.Context) error
-	SetMetadata(context.Context, interface{}) error
+const (
+	startEventName = "/vm/start"
```
sipsma: nit: may want to be a little more specific, i.e. `/firecracker-vm/start`, unless containerd has officially defined this is a topic to publish to for any VM manager shim.

firecracker-control/local.go
```diff
@@ -77,20 +80,35 @@ func newLocal(ic *plugin.InitContext) (*local, error) {
...
+		id     = req.GetVMID()
+		logger = log.G(ctx).WithField("vm_id", id)
+	)
+
+	if id == "" {
+		return nil, errors.New("invalid VM ID")
+	}
+
+	// Make sure VM ID unique
+	if _, err := s.getVM(id); err == nil {
```
sipsma: I think we need this `getVM` call to be atomic with the `s.vm[id] = newInstance` call at the bottom of this function, otherwise it seems like there will still be race conditions for concurrent requests.

firecracker-control/local.go
```diff
@@ -206,11 +214,22 @@ func (s *local) startMachine(ctx context.Context, id string, machine machineIfac
 		return err
 	}
 
+	if err := s.publisher.Publish(ctx, startEventName, &proto.VMStart{VMID: id}); err != nil {
+		logger.WithError(err).Error("failed to publish create event")
+		return err
```
sipsma: Should we actually fail the startMachine operation if we fail to publish the event?

If the event publish is just for informational purposes, it seems like it's enough to log the error but move on. Do we rely on the event publish for basic functionality anywhere?

firecracker-control/local.go
```diff
@@ -306,6 +325,7 @@ func (s *local) getVM(id string) (*instance, error) {
 	s.vmLock.RUnlock()
 
 	if !ok {
+		log.L.Warnf("attempt to get not existing VM id: %s", id)
```
sipsma: We call getVM to check if the VM exists when creating a new one, so will we see this log every time that happens? If so, I don't think a WARN log should be written for something that's expected to happen. Maybe letting the caller decide whether to log would be more flexible in this case.

firecracker-control/local.go
mxpv: I couldn't find any requirements for topics, so I think its fine to prefix them `/firecracker-vm`

firecracker-control/local.go
mxpv: Done

firecracker-control/local.go
mxpv: I don't know the exact use key for events yet. I'll clarify and will update this behavior. Let's leave it as is for now.

firecracker-control/local.go
```diff
@@ -77,20 +80,38 @@ func newLocal(ic *plugin.InitContext) (*local, error) {
...
+	var (
+		id     = req.GetVMID()
+		logger = log.G(ctx).WithField("vm_id", id)
+	)
+
+	if id == "" {
+		return nil, errors.New("invalid VM ID")
+	}
+
+	s.vmLock.Lock()
```
xibz: So basically this method is non-concurrent. Is this due to finding the vsock CID?

mxpv: See Erik's comment: https://github.com/firecracker-microvm/firecracker-containerd/pull/163#discussion_r275515176

xibz: Ah, if it is only racing on the map, I would much prefer the use of the `sync.Map` structure.

mxpv: That's not going to work. we need to check+add id atomically in order to prevent race conditions.

sipsma: Doesn't have to be addressed now, but it is a bit unfortunate that we aren't spinning up VMs in parallel anymore (though it remains to be seen if that is a performance bottleneck).

I think we could still spin up in parallel if we actually add an entry to `s.vm` for the ID right at the beginning of the method and just have a defer that removes the entry in case an error is encountered (each individual operation being wrapped in a quick lock/unlock).

I'm not convinced that complication is necessary at this stage, so no strong feelings either way for this PR.

xibz: Use the `LoadStore` and if it already exists, it'll return true and you can return an error. This has a smaller critical section

mxpv: Yep. Will add to TODO list.

mxpv: This gives same results as with mutex. See my thoughts regarding `sync.Map` here: https://github.com/firecracker-microvm/firecracker-containerd/pull/156#discussion_r273284575

xibz: But it is also smaller amounts of code and prevents a smaller critical section. I would be in favor of changing this since it is relatively small and easy. Right now, `CreateVM` is a sequential API.

mxpv: How is it going to be smaller?

xibz: You don't need the first if statement for one. The second if statement will be a `LoadStore` check and then cast. So my guess is one line less :p. But more importantly smaller critical section.

mxpv: So you create a VM first and than call `LoadStore` at the end?

PR #164, opened 2019-04-15
Support VM Agent managing multiple containers.

sipsma:

- [x] Use dynamic vsock ports for stdio (host decides and tells the vm what to use via protobuf msg)
- [x] Refactor agent to use one runc shim instance per-Container 
- [x] Refactor shared code between runtime+agent into common struct (called `VMTask` for now)
- [x] Working e2e test (updated `TestMultipleVMs_Isolated` to spin up 3 VMs each with 3 containers)
- [x] More doc strings + log statements all over the place
- [x] commented TODOs for error handling
- [x] fix race condition where container runs and exits before IO can be setup
- [x] further refactoring to get rid of duplicate code between runtime+agent
- [x] fix failure in TestShimExitsUponContainerKill_Isolated e2e test 
- [x] unit tests that need updating and a few new unit tests for the new code


runtime/vmdir.go
```diff
@@ -0,0 +1,83 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/containerd/containerd/runtime/v2/shim"
+	"github.com/firecracker-microvm/firecracker-containerd/internal"
+	"github.com/firecracker-microvm/firecracker-containerd/internal/bundle"
+	"github.com/pkg/errors"
+)
+
+type vmDir string
```
mxpv: I think it would make sense to move this to commons, so I can reuse this in `firecracker-control`

sipsma: For sure, will do

agent/service.go
```diff
@@ -136,44 +218,33 @@ func proxyIO(ctx context.Context, path string, port uint32, in bool) {
...
-		return nil, err
+		log.G(ctx).WithError(err).Error("error with stdio")
 	}
-	return extraData.RuncOptions, nil
 }
 
 // State returns process state information
 func (ts *TaskService) State(ctx context.Context, req *shimapi.StateRequest) (*shimapi.StateResponse, error) {
 	log.G(ctx).WithFields(logrus.Fields{"id": req.ID, "exec_id": req.ExecID}).Debug("state")
+	runcService, ok := ts.containers[req.ID]
```
mxpv: Should this be wrapped with `containersMu` as well? (or even `RWMutex`).

agent/service.go
```diff
@@ -136,44 +218,33 @@ func proxyIO(ctx context.Context, path string, port uint32, in bool) {
...
+	if !ok {
+		return nil, errors.Errorf("container with id \"%s\" not found", req.ID)
```
mxpv: nit: %q will do the job.

agent/service.go
sipsma: Oh good catch, yeah I'll probably just make a getContainer() method that gets the lock (and uses RWMutex)

internal/container/container.go
```diff
@@ -0,0 +1,191 @@
...
+		id:          containerID,
+		bundleDir:   bundleDir,
+		runcOptions: options,
+		fifoSet:     fifoSet,
+
+		stdinPort:  stdinPort,
+		stdoutPort: stdoutPort,
+		stderrPort: stderrPort,
+
+		ctx:    ctrCtx,
```
mxpv: nit: having `ctx` as a struct field is not not recommended [practice](https://dave.cheney.net/2017/08/20/context-isnt-for-cancellation). Is there any reason why  we can't pass it as func parameter instead?

sipsma: That makes sense. I didn't want to necessarily use the context that is passed in everywhere because the container exists longer than the lifetime of any single request (i.e. i/o copying lasts for the whole lifetime of the container).

I will refactor this to use the passed in context where it makes sense and use a non-context-based way of canceling the I/O (or something along those lines)

sipsma:
Just FYI, the e2e tests are passing the vast majority of the time, but I did see a one-off failure that appears to be an issue from before this PR. In the interest of not expanding this PR further I created [a separate issue I'll follow up on here](https://github.com/firecracker-microvm/firecracker-containerd/issues/169).

agent/service.go
```diff
@@ -15,165 +15,193 @@ package main
...
+func acceptVSock(ctx context.Context, port uint32) (net.Conn, error) {
+	listener, err := vsock.Listen(port)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to listen on vsock")
+	}
+
+	var conn net.Conn
+	for range time.Tick(10 * time.Millisecond) {
+		// accept is non-blocking so try to accept until we get a connection
+		// TODO: investigate if there is a way to distinguish
```
mxpv: This might help: https://github.com/containerd/ttrpc/blob/master/server.go#L94

internal/mock.go
```diff
@@ -0,0 +1,97 @@
...
+	ptypes "github.com/gogo/protobuf/types"
+
+	"github.com/containerd/containerd/runtime/v2/task"
+)
+
+var (
+	MockTaskService task.TaskService = &mockTaskService{}
+)
+
+type mockTaskService struct {
```
mxpv: 1. Can you rename it to `*_mock_test.go` to exclude this from regular builds.
2. Is there any reason not to use `mockgen` instead? ([example](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/firecracker-control/local_test.go#L14))

agent/service.go
```diff
@@ -15,165 +15,193 @@ package main
...
-		cancels: []context.CancelFunc{cancel},
+		eventExchange: eventExchange,
+		cancel:        cancel,
 	}
 }
 
-// Create creates a new initial process and container using runc
-func (ts *TaskService) Create(ctx context.Context, req *shimapi.CreateTaskRequest) (*shimapi.CreateTaskResponse, error) {
-	log.G(ctx).WithFields(logrus.Fields{"id": req.ID, "bundle": req.Bundle}).Info("create")
+func logPanic(logger *logrus.Entry) {
```
mxpv: nit: looks like we recover here in the first place and logging is a consequence. I'd propose `recoverPanic(ctx context.Context)`

internal/vm/task.go
```diff
@@ -0,0 +1,335 @@
...
+	m.tasks[task.ID] = task
+	return task, nil
+}
+
+func (m *TaskManager) get(containerID string) (*Task, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	if m.tasks == nil {
+		m.tasks = make(map[string]*Task)
```
mxpv: nit: can you move this to `NewTaskManager` so we don't have to perform this checks in multiple places.

tools/docker/Dockerfile
```diff
@@ -264,11 +266,23 @@ SHELL ["/bin/bash", "-c"]
 # Test image that starts up containerd and the naive snapshotter. The default CMD will drop to a bash shell. Overrides
 # to CMD will be provided appended to /bin/bash -c
 FROM firecracker-containerd-unittest as firecracker-containerd-e2etest-naive
+# Install tini init to handle zombies created by our double-forked shims.
```
sipsma: I've now encountered the issue with needing an init to reap zombies outside this PR, so I've updated #170 with this change to the Dockerfile and will be removing it from this PR

internal/mock.go
sipsma: 1. Yes
2. The only reason was I had no idea it existed, that's awesome! Will update

agent/service.go
sipsma: Well we recover, write the log, but then Fatal (which should exit the process), so we don't keep on running.

I can change the name to `logPanicAndDie` or something like that

agent/service.go
sipsma: This comment was carried over from the previous implementation. I'll take a look at that though either way, thanks. If it's a one line change I'll include it in this PR, but otherwise I'll save it for a next iteration.

internal/mock.go
sipsma: Once looking into this more I'm not sure how to have mockgen generate a mock for an interface defined in a different project (task.TaskService is defined in containerd).

I also realized that right now an even simpler option is to just make this struct in task_test.go:
```
type mockTaskService struct{
    task.TaskService
}
```

so I'll go with that for now.

agent/service.go
mxpv: I didn't notice `Fatal`. Do we really want to panic Agent? As I understand there no restart mechanism, so we'll lose all containers in a VM.

sipsma: Right now I think yeah it's best to prefer just crashing. Panics should really only be thrown if something so terrible has happened that the program is better off crashing (i.e. a nil pointer exception that, if ignored, could result in inconsistent or corrupted state).

In the future if we find that we are getting crashes in situations where we are better off still running then we can update this function in the future to not Fatal.

agent/service.go
sipsma: It was pretty straightforward, updated with that as a separate commit

runtime/service.go
```diff
@@ -67,45 +65,43 @@ func init() {
...
+	// type assertions
+	_ taskAPI.TaskService = &service{}
+	_ shim.Init           = NewService
+
+	sysCall = syscall.Syscall
 )
 
 // implements shimapi
 type service struct {
+	vm.TaskManager
```
xibz: I would much prefer if this was not embedded as it can make debugging more difficult.

internal/vm/task.go
```diff
@@ -0,0 +1,338 @@
...
+
+		cancelCh: make(chan struct{}),
+	}
+
+	m.tasks[task.ID] = task
+	return task, nil
+}
+
+func (m *taskManager) get(containerID string) (*Task, error) {
+	m.mu.RLock()
```
xibz: I think we should use a `sync.Map` or `RWLock` here, that way `get`s are not locking each other out.

internal/bundle/bundle.go
```diff
@@ -0,0 +1,143 @@
...
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package bundle
```
xibz: What is this package for? I see some logic for directories and as well as OCI stuff. Can we add a `doc.go` file describing the point of this package?

internal/vm/task.go
```diff
@@ -0,0 +1,338 @@
...
+	m.remove(ctx, containerID, task)
+}
+
+// RemoveAll calls Remove() on all tasks registered with the taskManager.
+func (m *taskManager) RemoveAll(ctx context.Context) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	for containerID, task := range m.tasks {
+		m.remove(ctx, containerID, task)
```
xibz: How does this not deadlock? You are calling `m.mu.Lock` and then again in `Remove`

agent/main.go
```diff
@@ -141,12 +135,13 @@ func main() {
...
 	log.G(ctx).Info("shutting down agent")
-	if _, err := runcTaskService.Shutdown(ctx, &shimapi.ShutdownRequest{ID: id, Now: true}); err != nil {
-		log.G(ctx).WithError(err).Error("runc shutdown error")
+
+	if err != nil {
+		log.G(ctx).WithError(err).Error("shim error")
+		os.Exit(1)
 	}
+
+	os.Exit(0)
```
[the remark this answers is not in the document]
sipsma: Yeah I suppose exit(0) is what will happen anyways so I can leave this out

internal/bundle/bundle.go
sipsma: It's just some helper utilities for interacting with a Bundle dir. Can add doc.go though, sure

internal/vm/task.go
sipsma: This is calling `remove`, not `Remove`. `remove` is the internal-only method that doesn't acquire a lock

internal/vm/task.go
sipsma: This is using an RWLock, it's calling `RLock()`

internal/vm/task.go
```diff
@@ -0,0 +1,338 @@
...
+			}()
+
+			_, err = io.CopyBuffer(fifoFile, sock, buf)
+			fifoFile.Close()
+
+		default:
+			logger.Fatalf("undefined io direction") // should not be possible, direction is a bool
+		}
+
+		if err != nil {
```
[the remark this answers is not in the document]
sipsma: `fifoFile` is closed in all the previous code branches already before this line

runtime/service.go
sipsma: Happy to change this, but I'm curious what about embedded fields makes debugging more difficult?

runtime/service.go
```diff
@@ -199,165 +133,165 @@ func NewService(ctx context.Context, id string, remotePublisher events.Publisher
...
 	return s, nil
 }
 
-// vmDir holds files, sockets and FIFOs scoped to a single given VM
-func (s *service) vmDir() string {
-	return filepath.Join(varRunDir, s.vmID)
+// vmDir holds files, sockets and FIFOs scoped to a single given VM.
+// It is unique per-VM and containerd namespace.
+func (s *service) vmDir() vm.Dir {
+	return vm.Dir(filepath.Join(varRunDir, s.namespace, s.vmID))
```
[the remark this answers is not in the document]
sipsma: I might be misunderstanding the concern, but we don't really care whether it's unique or not. Namespaces are expected to be re-used for multiple containers, same with vmID. If multiple containers have the same namespace and use the same vmID then that just means they will be executing in the same VM.

runtime/service.go
```diff
@@ -199,165 +133,165 @@ func NewService(ctx context.Context, id string, remotePublisher events.Publisher
...
-}
+	cmd.Dir = s.vmDir().RootPath()
 
-// firecrackerMetricsFifoPath is the path to the firecracker VMM's metrics fifo
-func (s *service) firecrackerMetricsFifoPath() string {
-	return filepath.Join(s.vmDir(), firecrackerMetricsFifoName)
-}
+	cmd.Env = append(os.Environ(),
+		fmt.Sprintf("%s=%s", vmIDEnvVarKey, s.vmID),
+		"GOMAXPROCS=2")
```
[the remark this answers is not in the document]
sipsma: The git diff doesn't make this obvious, but this is what we were doing prior to this change also. It's something containerd does in their shim implementations, I think for a mix of performance and debuggability concerns.

Can add a comment here either way as it's definitely not apparent offhand why this is here.

runtime/service.go
```diff
@@ -369,75 +303,125 @@ func (s *service) StartShim(ctx context.Context, id, containerdBinary, container
 
 	// If we're here, there is no pre-existing shim for this VMID, so we spawn a new one
 	defer func() {
-		log.G(ctx).WithField("id", id).Debug("closing shim socket")
+		log.G(ctx).WithField("id", containerID).Debug("closing shim socket")
 		shimSocket.Close()
 	}()
 
-	err = os.MkdirAll(s.vmDir(), 0700)
+	err = s.vmDir().Create()
```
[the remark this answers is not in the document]
sipsma: It depends on the circumstance IMO, but in this case I prefer what I have because it makes it easy to add stuff like
```
var err error
defer func() {
  if err != nil 
    cleanup()
  }
}()

err = doSomething()
if err != nil {
   return err
}
```

We aren't doing that at this exact moment in this function, but we actually will need to in the very near future when I fix #169 right after this PR is merged, so I'd prefer to leave it as is for now.

runtime/service.go
xibz: Generally when you go searching for a method or field, it isn't on the object but the embedded one, making it not as obvious when viewing the code, and if you start embedding multiple structs it becomes even crazier.

runtime/service.go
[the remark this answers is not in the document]
sipsma: Ah, even better, it turns out it's not necessary for us to do this explicitly, containerd will set it for us: https://github.com/containerd/containerd/blob/810b3c37bb13662a289b9f6b028b5b5435be72b7/runtime/v2/shim/shim.go#L117

So I'll just leave this out

internal/vm/task.go
xibz: Oh ffs. Totally missed the R lol.

internal/vm/task.go
xibz: Ah, missed the little r implementation.

runtime/service.go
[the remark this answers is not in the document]
xibz:
```go
// This is fine too
if err = doSomething(); err != nil {
    return err
}
```

runtime/service.go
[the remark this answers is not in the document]
xibz: Ah, okay cool. My misunderstanding. For some reason I thought we cared about that.

internal/vm/task.go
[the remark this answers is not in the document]
xibz: Yea, I figured as much but would it make sense to remove the branch closes for a single close in this if statement?

runtime/service.go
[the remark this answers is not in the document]
sipsma: Ohhh geeze, I've been confused about that for a while then. I always though that notation with `;` resulted in the setting of the variable `err` being scoped to the if block. I never thought to just do `=` instead of `:=` 😅

Makes sense though, will update.

internal/vm/task.go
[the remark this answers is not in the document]
sipsma: In this situation, no I don't think so. We want to close the FIFO in a very particular order that differs between the two case statements above this.

When we are copying `FIFOtoVsock`, we want to close the FIFO before the vsock. This allows us to first cut off the FIFO, then drain buffered data and then finally close the vsock.

When we are copying `VSockToFIFO`, we the exact opposite, to first cut off the vsock, then drain buffered data and then finally close the FIFO.

runtime/service.go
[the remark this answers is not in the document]
xibz: Haha, no worries. To be fair, the notation isn't very popular in languages.

internal/vm/task.go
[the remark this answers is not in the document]
xibz: Ah, I see. Hm, Im just worried about adding another switch case and having someone forget to close the fifo. Can we add a comment illustrating that each case in the switch statement should close the fifo since that isn't handled outside of the switch scope.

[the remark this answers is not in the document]
sipsma: There shouldn't be any possibility of adding another case as the underlying type here is a `bool`, so there can only be 2 cases.

Adding the comment though for extra clarity.

runtime/service.go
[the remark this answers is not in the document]
sipsma: I started doing this, but then I encountered situations like this where I wanted to change
```
uuid, err := uuid.NewV4()
if err != nil {
  return "", err
}

s.vmID = uuid.String()
```
to
```
if uuid, err := uuid.NewV4(); err != nil {
  return "", err
}

s.vmID = uuid.String()
```

But that doesn't work because `uuid` is scoped to just the if conditional now. It gets even worse for the err checking with the `shimSocket`.

Honestly, given all that, I'd just prefer to keep it as is. It just seems simpler and less surprising to not use the `;` syntax unless there's a strong reason to.

PR #170, opened 2019-04-19
Use makefile targets for all CI tests.

sipsma:

This will make it much easier for developers to run CI tests locally.


tools/docker/Dockerfile
```diff
@@ -264,11 +264,22 @@ SHELL ["/bin/bash", "-c"]
 # Test image that starts up containerd and the naive snapshotter. The default CMD will drop to a bash shell. Overrides
 # to CMD will be provided appended to /bin/bash -c
 FROM firecracker-containerd-unittest as firecracker-containerd-e2etest-naive
+# Install tini init to handle zombies created by our double-forked shims.
+# Unfortunately, we need to do this rather than only use docker's "--init" flag because docker
+# currently puts its default init binary at /dev/init, but we need to bind-mount /dev from
+# the host in order for the snapshotter to work.
+# This should no longer be an issue after github.com/moby/moby #37655 is released.
```
samuelkarp: This PR: https://github.com/moby/moby/pull/37655/?  I can't quite tell how that's related...

agent/Makefile
```diff
@@ -10,11 +10,12 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
-EXTRAGOARGS:=
+EXTRAGOARGS?="-v -count=1"
```
samuelkarp: Do we want to change this default everywhere?  It was set that way in CI to ensure that CI always ran all tests, but in normal development one might want to take advantage of Go's test result caching and skip tests where no code is modified.

tools/docker/Dockerfile
sipsma: Ugh good catch, I meant to link https://github.com/moby/moby/pull/37665, will fix

agent/Makefile
sipsma: Yeah fair point, it would make more sense to just set this env var in the buildkite yaml file for now and default to nothing in the makefile, will update

Makefile
```diff
@@ -44,4 +46,23 @@ deps:
 install:
 	for d in $(SUBDIRS); do $(MAKE) -C $$d install; done
 
-.PHONY: all $(SUBDIRS) clean proto deps lint install
+docker-images:
```
xibz: Can we separate this out into multiple tasks? That way we can add a stub file for each one, and if someone wants to rebuild a specific image it isn't too much work.

tools/docker/Dockerfile
```diff
@@ -264,11 +264,22 @@ SHELL ["/bin/bash", "-c"]
 # Test image that starts up containerd and the naive snapshotter. The default CMD will drop to a bash shell. Overrides
 # to CMD will be provided appended to /bin/bash -c
 FROM firecracker-containerd-unittest as firecracker-containerd-e2etest-naive
+# Install tini init to handle zombies created by our double-forked shims.
+# Unfortunately, we need to do this rather than only use docker's "--init" flag because docker
+# currently puts its default init binary at /dev/init, but we need to bind-mount /dev from
```
xibz: Would it make sense to mount our `/dev` to a different path and then copy the contents to `/dev`?

Makefile
sipsma: Yeah for sure, I split them into separate targets and made the `docker-images` target just run all of them (in order so that we maximize re-use of build cache since the different images are mostly layers on top of each other)

tools/docker/Dockerfile
sipsma: I don't think that will work here as we want the actual bind mount of /dev from the host so if udev creates a device, our container will also see it. If we just copied when the container starts we will just be getting a snapshot of what /dev looked like at that time rather than ongoing updates to it.

PR #176, opened 2019-05-06
Update containerd to most recently needed commit.

sipsma:
Specifically, commit 82f2ac7edec03dfc301d99d8910c9e53a9d71d90, which ensures we
have the following needed commits:
* 82f2ac7edec03dfc301d99d8910c9e53a9d71d90
  * Stop logging error on v2 multi shim log failure
* 810b3c37bb13662a289b9f6b028b5b5435be72b7
  * Support disabling default setup of shim logger.
* 665715b8a6c595a1a96e6b60cbf8b544ad63cc78
  * Allow reuse client for external GRPC services

Those commits are not currently in a tagged release. Once they are, it should
make sense to update to use a tagged release branch again instead of a commit ID
off of containerd's master branch.


runtime/service.go
```diff
@@ -993,21 +998,20 @@ func (s *service) startVM(ctx context.Context,
...
 		// using the eventbridge client. All events that are published on the
 		// agent's exchange will also be published on our own
-		if err := <-eventbridge.Attach(ctx, eventBridgeClient, s.eventExchange); err != nil && err != context.Canceled {
+		if err := <-eventbridge.Attach(context.Background(), eventBridgeClient, s.eventExchange); err != nil && err != context.Canceled {
 			log.G(ctx).WithError(err).Error("error while forwarding events from VM agent")
 		}
 	}()
 
-	go s.monitorTaskExit(ctx)
+	go s.monitorTaskExit(context.Background())
```
mxpv: I don't know whether its worth to do it here, but I would prefer something like:
```
func ttrpcHandler(ctx) {
    cid = findCid(ctx)
    buildConfiguration(ctx)
    startVM(context.Background())
    monitorTask(context.Background())
}
```
Mixed context behavior is bad.

sipsma: I guess I'm not really sure how that will help as it still results in a mixed use of context in `ttrpcHandler`.

We want the request ctx (the one that will be cancelled after the request is done) for functions like `dialVsock` and `findNextAvailableVsock`. However, we want `Background` for resources that are being created that will live longer than the duration of the request. We thus can't just provide `Background` to `startVM`, it sometimes needs the request context and sometimes needs `Background`.

We just need to be aware going forward that the `ctx` parameter of these methods lives only for the lifetime of the request and never provide it to something that is intended to live longer than the request. I don't like this in that it requires us to know how each function that accepts a context will use the context, but I'm not sure we can fully solve this problem as we still need to interface with 3rd party libraries that accept context.

I think the least that can be done right now is rename `ctx` to `requestCtx` everywhere. That way it's at least easier to remember that it will be cancelled after the request is done. I'll update with that.

runtime/service.go
```diff
@@ -755,7 +756,11 @@ func (s *service) Shutdown(ctx context.Context, req *taskAPI.ShutdownRequest) (*
 	}
 
 	// Exit to avoid 'zombie' shim processes
-	defer os.Exit(0)
+	defer func() {
+		s.shutdown()
```
xibz: Probably want to check to see if `shutdown` is not `nil` before calling.

runtime/service.go
mxpv: It wouldn't help to avoid in `ttrpcHandler`, but rather will help to avoid spreading this behavior to other functions. And you have better control where to keep cancellation behavior and where to ignore it.

sipsma: That's what I'm saying though, we need the behavior in other functions. We need to either provide the request context or the background context in `startVM` depending on what operation `startVM` is performing. Request context is for anything that should be done when the request is done, background context is for anything that should continue existing after the request is done.

I'm renaming all the context objects to either `requestCtx` or `shimCtx` depending on whether containerd's shim code is providing us a context scoped to a single request or to the lifetime of the shim. There's also some low-hanging opportunities to refactor some of other functions to not even need context anymore (the ones that just needed a logger), which I'll update with too.

agent/service.go
```diff
@@ -538,8 +540,6 @@ func (ts *TaskService) Shutdown(ctx context.Context, req *taskAPI.ShutdownReques
 	log.G(ctx).WithFields(logrus.Fields{"id": req.ID, "now": req.Now}).Debug("shutdown")
 	ctx = namespaces.WithNamespace(ctx, defaultNamespace)
 
-	// We don't want to call runc.Shutdown here as it just os.Exits behind.
```
samuelkarp: How come these comments are removed?

sipsma: It's no longer true, containerd made some nice updates that got rid of any need to call `os.Exit` from `Shutdown`: https://github.com/containerd/containerd/commit/ae87730ad2511ae39239ac7aac75074768aa3c1b

runtime/service.go
sipsma: I made these changes here: https://github.com/sipsma/firecracker-containerd/commit/133b0919ad49305123052ddd59fdd260c227d681

I just pushed it to a separate branch for now because the changes are pretty sizable on top of the original purpose of this PR (just update containerd+ttrpc) and aren't strictly required for functionality.

Any objections to merging this PR as is and handling the rest of the updates in an immediate followup PR for that commit I linked to above?

runtime/service.go
sipsma: done

runtime/service.go
samuelkarp: SGTM

PR #178, opened 2019-05-07
Re-use proto messages for both ttrpc and grpc.

sipsma:
Before this change, each protobuf message type was generated for both grpc and
ttrpc. However, if both definitions get loaded into the same process (i.e. a
process that's translating between the two such as the fccontrol plugin), the
protobuf names will conflict with each other.

This change results in protobuf message types being generated only once. Now
only the service definitions are generated separately for both grpc and ttrpc.


mxpv:
Do you have any reasons to nest `/proto/services/fccontrol/{grpc,ttrpc}` ? We expect no more than 1 service, so it adds more complexity than needed.

sipsma:
> Do you have any reasons to nest `/proto/services/fccontrol/{grpc,ttrpc}` ? We expect no more than 1 service, so it adds more complexity than needed.

Well it's more complex to do something like `/proto/{grpc,ttrpc}` because then we have to include the `fccontrol.proto` file in the same directory as the other protos, which makes the Makefile there more complicated (can't just use wildcards anymore, have to use a wildcard but filter out `fccontrol.proto` and then run different commands for it).

As far as something shorter like `/proto/fccontrol/{grpc,ttrpc}`, I don't really see any difference in complexity between that and`/proto/services/fccontrol/{grpc,ttrpc}`. The slight extra benefits of the longer path name are that it makes it really clear what's being imported when you add an import statement in Go and makes it easier to add more services in the future in the need ever arises.

PR #182, opened 2019-05-10
Integrate fccontrol plugin with shim. 

sipsma:
This is a fairly significant reorganization that combines the work done on the
fccontrol plugin with the shim. Significant aspects include:
1. The shim implements the fccontrol interface (over ttrpc) and handles the
   actual VM operations. The plugin is now only responsible for starting shims
   and forwarding fccontrol requests from clients to the correct shim (based on
   VMID).
2. The fccontrol plugin is now completely stateless (as a result of the above
   change).
3. A shim and its VM are now started together, so the shim start routine no
   longer directly starts a shim process; it just gets a VMID and issues a
   CreateVM call to the plugin. This lets the plugin take care of looking to see
   if the shim for the incoming VM already exists (and starting it if not).

This change was biased towards keeping the behavior and interface unchanged, but
there are a few significant updates from the previous state of things:
1. The shim no longer by default shuts down when all tasks have exited; instead
   it continues to run until either StopVM is called or Shutdown is called when
   no tasks are left. Its lifecycle and the VM lifecycle are now the same
   essentially.
   * This results in less surprising behavior; a shim is created via a CreateVM
     call and can exist without any tasks, so it makes sense that it should be
     able to continue to exist after each task has exited too.
   * Otherwise, it's easy for users to, for example, create a VM and issue
     several CreateTask calls for that VM, but the first Task is created and
     exits quickly before the other tasks were setup, which results in the shim
     exiting and the rest of the tasks failing to be created.
   * For situations where the previous behavior is more desirable, a new
     CreateVM parameter, ExitAfterAllTasksGone, can be set to true. This is
     used by the shim start routine when the client did not specify any VMID
     (and thus don't care about VMs, just their task).
2. You can no longer pass Firecracker machine options as part of the
   CreateTask extra data field. If you create a task with a VMID that doesn't
   exist yet or no VMID at all, the VM will be created using the default
   options as configured in the runtime's config file.
   * This greatly simplifies the amount of branches our code has to handle and
     also greatly simplifies the interface we are giving to users without
     reducing the functionality. If users want custom VM options, they just
     need to use the CreateVM call. If they don't care about custom VM options,
     they can continue to just issue a CreateTask request and get the defaults
     specified in our runtime config file.

- [x] Working e2e (passing e2e tests)
- [x] ~~Use actual plugin in CI system instead of the special containerd binary with plugin baked in~~ This is not going to be pursued right now, we are just going to use the baked containerd binary for now
- [x] ~~Cleanup some parts of the shim logging (it may just be cleanest for the plugin to create a FIFO for the shim)~~ For now, went with just setting the stdout/stderr of the shim process and forwarding the logs to containerd. See the commit message for the new commit for more details.
- [x] Finish cleanup of old way of passing VM parameters via Task (runc options doesn't need to be embedded in firecracker options anymore)


firecracker-control/local.go
```diff
@@ -81,263 +67,246 @@ func newLocal(ic *plugin.InitContext) (*local, error) {
...
+		return nil, status.Errorf(codes.AlreadyExists, "VM with ID %q already exists", id)
+	} else if err != nil {
+		err = errors.Wrapf(err, "failed to open shim socket at address %q", shimSocketAddress)
+		s.logger.WithError(err).Error()
+		return nil, err
+	}
 
-func (s *local) newInstance(ctx context.Context, id string, req *proto.CreateVMRequest) (*instance, error) {
-	logger := log.G(ctx).WithField("vm_id", id)
+	// If we're here, there is no pre-existing shim for this VMID, so we spawn a new one
```
mxpv: If we receive two requests simultaneously with the same ID, the code below will attempt to create a VM twice?

sipsma: If we receive two requests with the same ID one will get to this line and another will follow the `isEADDRINUSE` branch. Opening the `shimSocket` works essentially like a lock; only one process on the box can have the shimSocket for a given vmID open at a given time, anyone else that tries to open it gets `EADDRINUSE`.

runtime/service.go
```diff
@@ -436,9 +361,296 @@ func (s *service) nextVSockPort() uint32 {
...
+		return nil
+	case <-time.After(firecrackerStartTimeout):
+		return status.Error(codes.DeadlineExceeded, "timed out waiting for VM start")
+	}
+}
+
+func (s *service) CreateVM(requestCtx context.Context, request *proto.CreateVMRequest) (*empty.Empty, error) {
+	errCh := make(chan error, 1)
+
+	s.vmStartOnce.Do(func() {
```
xibz: Can we break this into multiple functions? This function is a little long, so I think having things like `createMachine` will at least group things together nicely.

runtime/service.go
```diff
@@ -171,96 +182,60 @@ func NewService(shimCtx context.Context, id string, remotePublisher shim.Publish
...
 
-	err = cmd.Start()
-	if err != nil {
-		return nil, err
-	}
+	// TODO we have to create separate listeners for the fccontrol service and shim service because
+	// containerd does not currently expose the shim server for us to register the fccontrol service with too.
+	// This is likely addressable through some relatively small upstream contributions; the following is a stop-gap
+	// solution until that time.
+	fcSocketFDEnvVal := os.Getenv(internal.FCSocketFDEnvKey)
```
xibz: Where is `FCSocketFDEnvKey` setting the file descriptor?

sipsma: local.go sets the env var and the FD in `newShim` (the FD is set as part of `ExtraFiles`)

runtime/service.go
```diff
@@ -436,9 +361,296 @@ func (s *service) nextVSockPort() uint32 {
...
+func (s *service) waitVMReady() error {
+	select {
+	case <-s.vmReady:
+		return nil
+	case <-time.After(firecrackerStartTimeout):
+		return status.Error(codes.DeadlineExceeded, "timed out waiting for VM start")
+	}
+}
+
+func (s *service) CreateVM(requestCtx context.Context, request *proto.CreateVMRequest) (*empty.Empty, error) {
```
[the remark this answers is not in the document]
sipsma: Sure, I'm surprised the linter did not yell at me for that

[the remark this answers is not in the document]
sipsma: In general I agree but this is one of those functions that does so much interaction with the system and other APIs that you end up mocking almost everything out. In my experience, unit tests like that require much more maintenance than they provide benefit (at which point e2e tests are better IMO).

I will see if breaking this function up as you suggested below makes this more unit testable and update with unit tests if it seems like they will provide benefit over just e2e.

runtime/service.go
sipsma: Yeah at min I can extract out everything in the `Once.Do`, I'll also see if breaking that function up some more is beneficial.

runtime/service.go
```diff
@@ -904,19 +1129,24 @@ func findNextAvailableVsockCID(requestCtx context.Context) (uint32, error) {
...
 	file, err := os.OpenFile(vsockDevicePath, syscall.O_RDWR, 0600)
 	if err != nil {
-		return 0, errors.Wrap(err, "failed to open vsock device")
+		return 0, nil, errors.Wrap(err, "failed to open vsock device")
 	}
-	defer file.Close()
+
+	var closeFile = true
+	defer func() {
+		if closeFile {
```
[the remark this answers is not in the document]
sipsma: Sure

firecracker-control/local.go
```diff
@@ -81,263 +67,251 @@ func newLocal(ic *plugin.InitContext) (*local, error) {
...
-		vm:          make(map[string]*instance),
-		rootPath:    ic.Root,
-		publisher:   ic.Events,
-		findVsockFn: findNextAvailableVsockCID,
+		containerdAddress: ic.Address,
+		logger:            log.G(ic.Context),
 	}, nil
 }
 
 // CreateVM creates new Firecracker VM instance
```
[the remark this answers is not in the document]
sipsma: Sure, agree it could be a lot more descriptive, I'll update others similarly

firecracker-control/local.go
```diff
@@ -81,263 +67,251 @@ func newLocal(ic *plugin.InitContext) (*local, error) {
...
-
-	instance, err := s.newInstance(ctx, id, req)
-
-	s.vmLock.Lock()
-	defer s.vmLock.Unlock()
+	s.logger.Debugf("using namespace: %s", ns)
 
+	// We determine if there is already a shim managing a VM with the current VMID by attempting
+	// to listen on the abstract socket address (which is parameterized by VMID). If we get
+	// EADDRINUSE, then we assume there is already a shim for the VM and just hand off to that
```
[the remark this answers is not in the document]
sipsma: Yeah this comment made more sense when it used to be in the Shim Start. I'll rewrite it for the new context.

The idea now is that we return an `AlreadyExists` error here, which can then be interpreted by the Shim Start to mean that there is already a shim which can be handed off to.

firecracker-control/local.go
```diff
@@ -81,263 +67,251 @@ func newLocal(ic *plugin.InitContext) (*local, error) {
...
 
-	go func() {
-		ctx := namespaces.WithNamespace(context.Background(), ns)
+	return resp, nil
+}
 
-		if err := machine.Wait(ctx); err != nil {
-			logger.WithError(err).Error("failed to wait the VM")
-		}
+func (s *local) newShim(ns, vmID, containerdAddress string, shimSocket *net.UnixListener, fcSocket *net.UnixListener) (*exec.Cmd, error) {
```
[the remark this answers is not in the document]
sipsma: Well it definitely started out from their example shim implementation a while back, but it's been modified more and more as time has gone on. Right now, it's still somewhat recognizable but has diverged a fair bit: https://github.com/containerd/containerd/blob/master/runtime/v2/runc/v2/service.go#L117

I'd say in general our entire original shim implementation copied quite a bit of their runtime v2 code at the time (besides just this newShim function). While I *suspect* that's what the containerd authors intended for others to do, I have no idea whether it would be appropriate to attribute to them or not. Let me know what you think and how to attribute them if you think that makes sense.

As for upstream changes, the best I can think of would be if they provided a default "newShimCmd" that we could get and then modify with our customizations, which would let us de-dupe the `cmd.Args` and `cmd.SysProcAttr` lines.

However, that really only saves a couple lines. I'm not convinced it would actually provide much benefit.

firecracker-control/local.go
```diff
@@ -81,263 +67,251 @@ func newLocal(ic *plugin.InitContext) (*local, error) {
...
 
-	return nil
-}
-
-// StopVM stops running VM instance by VM ID
-func (s *local) StopVM(ctx context.Context, req *proto.StopVMRequest) (*empty.Empty, error) {
-	logger := log.G(ctx).WithField("vm_id", req.GetVMID())
+	cmd.Env = append(os.Environ(),
+		fmt.Sprintf("%s=%s", internal.VMIDEnvVarKey, vmID),
+		fmt.Sprintf("%s=%s", internal.FCSocketFDEnvKey, "4")) // TODO remove after containerd is updated to expose ttrpc server to shim
```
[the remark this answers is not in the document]
sipsma: We do actually (as long as we trust the Go stdlib to work as advertised) because we are setting it via `ExtraFiles`: https://golang.org/pkg/os/exec/#Cmd

In theory we could calculate this number "4" from the index of where we appended the file to `ExtraFiles` which would make it robust to any other code changes. It's kind of borderline whether that extra complication is worth it given we should hopefully be able to remove this entirely after updating containerd to just expose the ttrpc server to the shim (and the fact that containerd itself just hardcodes `3` when they do this exact same thing with the shimSocket), but I'll give updating with that a shot.

runtime/service.go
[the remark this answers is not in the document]
sipsma: I ended up adding e2e test coverage for these paths instead. The unit tests were still just gonna mock out so much that the effort required to add e2e coverage instead was equal or less. Let me know if there's any objections to this.

runtime/service.go
```diff
@@ -436,10 +345,266 @@ func (s *service) nextVSockPort() uint32 {
...
+	defer logPanicAndDie(s.logger)
+
+	errCh := make(chan error, 1)
+	s.vmStartOnce.Do(func() {
+		errCh <- s.createVM(requestCtx, request)
+		close(errCh)
+	})
+
+	select {
+	case err := <-errCh:
```
samuelkarp: I think it might be simpler to structure this without a channel, since we're not really doing any cross-goroutine communication or concurrency.  What about something like this?

```go
calledCreate := false
var err error
s.vmStartOnce.Do(func() {
	err = s.createVM(requestCtx, request)
	calledCreate = true
}

if !calledCreate {
	return nil, status.Error(codes.AlreadyExists, "shim cannot create VM more than once")
}

// everything else
```

runtime/service.go
```diff
@@ -436,10 +345,266 @@ func (s *service) nextVSockPort() uint32 {
...
+			return nil, err
+		}
+
+		// creating the VM succeeded, setup monitors and publish events to celebrate
+
+		if request.ExitAfterAllTasksGone {
+			go s.monitorTaskExit()
+		}
+
+		go func() {
```
samuelkarp: Can the event stuff be moved out into a separate function?

runtime/service.go
```diff
@@ -171,96 +183,61 @@ func NewService(shimCtx context.Context, id string, remotePublisher shim.Publish
...
 }
 
 func (s *service) StartShim(shimCtx context.Context, containerID, containerdBinary, containerdAddress string) (string, error) {
-	s.logger.WithField("id", containerID).Debug("StartShim")
-
-	namespace, err := namespaces.NamespaceRequired(shimCtx)
-	if err != nil {
-		return "", errors.Wrap(err, "failed to extract namespace from context")
-	}
+	logrus.SetOutput(os.Stderr) // containerd reads the socket address from stdout only during shim start
```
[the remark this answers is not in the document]
sipsma: That's what I wanted to do (to just always use Stderr), but that results in another issue that requires going down a rabbit hole to explain fully.

If you always use Stderr for the logger here, then the plugin will have to assume that anything written to Stderr is already pre-formatted via logrus and thus should be passed through to containerd logs unmodified ([see the 4 lines here for more context](https://github.com/sipsma/firecracker-containerd/blob/fccontrol/firecracker-control/local.go#L291)).

That becomes a problem if we hit [this line in containerd's shim code](https://github.com/containerd/containerd/blob/master/runtime/v2/shim/shim.go#L147), which occurs if the shim is exiting with an error. That line prints to stderr directly without any logrus formatting, so the end result is your containerd logs have this random line with no timestamp or context.

The solution here works because the StartShim method is never hit by the plugin; the plugin only ever starts up the actual runtime shim, it doesn't invoke the start shim or delete shim routines.

Obviously it's tempting to just update containerd to print that line with the proper logrus formatting. However, that would result in the actual error object containerd eventually constructs server-side having the full logrus formatting, which would probably be confusing when returned to clients.

Overall, the fundamental reason this is all so messy right now is that we can't rely on containerd's normal way of providing shim logging due to [the issues I mention in this commit (part of this PR)](https://github.com/sipsma/firecracker-containerd/commit/d27ed36ceb1ed81ca462d52022c6b85edd4316ab).

I think the best long-term way of addressing this would be to figure out how to update containerd to not tie a shim log FIFO's existence to the existence of a bundle directory. That's a fairly sizable endeavor though separate from the current effort.

All that being said, I did realize while going through this all in my head again that 1. the comment I have is slightly inaccurate (containerd reads `CombinedOutput`, which is both stdout and stderr, not just stdout) and 2. I need to remove the `.Debug` log call below as, if debug is on, that will result in the debug log line being interpreted as part of the address string incorrectly.

I think what I'll do for the moment is change the StartShim logger to not use `os.Stderr` and instead actually use the log FIFO provided by containerd, which is okay because StartShim will always be invoked inside a bundle dir (an assumption that doesn't hold for the other parts of our code). That's probably the "minimal kludge" solution for now until we can look into the better long-term solution I mentioned above.

runtime/service.go
sipsma: I had that originally but liked the idea of unifying `err` and `calledCreate` with a channel, which is why I went with what I have. It just seems nice that the thing that determines if you ran will also tell you what, if anything, went wrong during the run.

This is splitting hairs in my mind though; is the current solution actually confusing enough to warrant updating? Happy to update if so.

runtime/service.go
sipsma: Each individual function (the `Attach` goroutine, the `VMStart` publish and the `VMStop` publish) can be separately factored out into their own function, but it wouldn't really be possible or make sense to put all or a combination of them in one single function.

I can certainly do that, but it's one of those situations where we would be factoring out 3 line functions that only get called exactly once in the whole program. It doesn't provide any DRY benefits and doesn't IMO help readability that much because they are so short in the first place. I feel like all it does is make it harder to know what's actually happening since it forces you to jump around to random functions across the codebase instead of just seeing them directly in the only place they are actually used.

This is another case of splitting hairs though, if you or anyone else feels strongly I have no problem factoring these out.

runtime/service.go
samuelkarp: The use of a channel caused me to spend some time wondering about whether the `case err := <-errCh:` would be hit reliably, so it took me a bit more time to read.  There's also a bit of a cosmetic thing about reducing indentation.  But there's nothing wrong about this code, these are just stylistic.  I prefer the way I suggested, but not to such a huge amount.

> It just seems nice that the thing that determines if you ran will also tell you what, if anything, went wrong during the run.

Yeah, I really wish Go had a `Result<Ok<V>, Err<E>>`.

sipsma: Okay that makes sense, I'll update then, no need to make things unnecessarily confusing when all else is equal.

And hopefully Go generics will get us `Result` someday...

runtime/service.go
samuelkarp: What about something like this?

```go
func (s *service) CreateVMEvents(vmid string) (createFunc, stopFunc func() error) {
	go func() {
		if bridgeErr := eventbridge.Attach(s.shimCtx, s.eventBridgeClient, s.eventExchange); bridgeErr != nil {
			s.logger.WithError(err).Error("error while forwarding events from VM to agent")
		}
	}()
	createFunc = func() error {
		return s.eventExchange.Publish(s.shimCtx, StartEventName, &proto.VMStart{VMID: vmid})
	}
	stopFunc = func() error {
		return s.eventExchange.Publish(s.shimCtx, StopEventName, &proto.VMStop{VMID: vmid})
	}
	return
}
```

I don't really like mixing the nuts and bolts of how an event is created (protobufs) with the logic of actually starting the VM.

mxpv: We can consider runc shim's approach, that makes events publishing a lot more cleaner ([1](https://github.com/containerd/containerd/blob/master/runtime/v2/runc/v2/service.go#L295), [2](https://github.com/containerd/containerd/blob/master/runtime/v2/runc/v2/service.go#L690))
But I'm not sure that it worth doing that for a just few events.

sipsma: I get the point about not mixing in the proto stuff with everything else, so I just added two methods `publishVMStart` and `publishVMStop`. The names are obvious enough that I think it's fine to move the code elsewhere even though they are only called once each.

For the rest, I ended up taking a slightly different approach that I believe addresses the concerns:
1. I moved the `eventbridge.Attach` call to a common method with `eventbridge.Republish`. This makes logical sense in that now all of the republishing/forwarding we do is setup in one place.
2. I created a method similar to the existing `monitorTaskExit` called `monitorVMExit` that does that `Wait` call and publishes the Stop event once that returns.

Let me know if that addresses the concerns.

PR #184, opened 2019-05-10
Removes loop devices before starting job.

xibz:
Removes loop devices per job

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -23,6 +23,16 @@ steps:
 
   - wait
 
+    # Global concurrency has been set to one in the website configuration.
+    # With that we can now clean up loop devices before starting jobs to ensure
+    # we don't hit the loop device limit
+  - label: ":lint-roller: loop device cleanup"
+    agents:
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -21,6 +21,16 @@ steps:
       DOCKER_IMAGE_TAG: "$BUILDKITE_BUILD_NUMBER"
     command: make docker-images
 
+  - wait
+
+    # Global concurrency has been set to one in the website configuration.
```
samuelkarp: I think this is the wrong way to do this.  We have a fairly large compute resource for running these jobs, and setting concurrency to 1 means we'll have a lot of idle capacity and introduce artificial delays when multiple PRs are in progress.

xibz: Yea, I agree, but this is currently a short term solution til' we get something figured out. This should fix two issues. One, the load of loop devices created will now be cleaned up. Second, if two PRs are running they will stomp over each other during snapshotter tests causing them to fail. Until we come up with a better solution for those two issues, let's go with this

PR #154, opened 2019-04-02
runtime:  Adding stubbed drives when initializing VM 

xibz:
Signed-off-by: xibz <impactbchang@gmail.com>

**TODO**
- [x] Add agent logic to handle stubbed block devices
- [x]  Add documentation
- [x] Add more tests
- [x] Clean up code
- [x] Add e2e test

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #185, opened 2019-05-10
dangling whitespace linter isn't friendly with tests

xibz:
Some tests require spaces at the end of strings for proper string
comparison

The linter, git-validation, checks more than just commit messages, but the actual diff that is going to be merge. And if there are any danglers, it'll error out.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,6 +21,7 @@ linters:
     - gocritic
     - interfacer # Suggests narrower interface types
     - scopelint # Checks for unpinned variables
+    - gofmt
 
 issues:
   exclude-use-default: false
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ distclean: clean
 
 lint:
 	$(BINPATH)/ltag -t ./.headers -excludes "tools $(SUBMODULES)" -check -v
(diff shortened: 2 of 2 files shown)


samuelkarp:
What's an example where this is failing?  Would it be better for us to exclude certain kinds of files or certain folders from this linter instead of disabling it entirely?

xibz:
@samuelkarp - The main issue is this is a git validator which doesn't have any context to the language. So in the case where multiline strings are a thing, it'll fail if there is a dangling whitespace in any of the multilined portion of the string. Basically it is not doing dangling whitespace determination correctly and I don't understand why a git tool is even doing that. I've enabled the `gofmt` in `golang-ci` instead which will do a lot better job since it actually uses the AST.

samuelkarp:
@xibz Can you give me a concrete example?  I understand your explanation but I'm having a hard time visualizing a specific change that is rejected by the existing validator.

xibz:
@samuelkarp -
```
str := `foo
bar<WHITESPACE>
`
```

This would cause the linter to explode since it doesn't understand Go's syntax. There is a change pending currently in #154 that compares `lsblk` output which has multiline dangling whitespace in its string.

samuelkarp:
Thank you!

PR #186, opened 2019-05-14
docs: root filesystem design considerations

samuelkarp:
This is a first draft at putting things down that we've talked about in-person.  Still some open questions for the future, but hopefully this gives some context.

### DIFF
--- a/docs/root-filesystem.md
+++ b/docs/root-filesystem.md
@@ -1,0 +1,64 @@
+# Root Filesystem
+
+firecracker-containerd requires a kernel and root filesystem to be specified for
+the Firecracker virtual machine where containers run.  This document specifies
+the hard requirements for that root filesystem, the development tools we built,
+and some thoughts around where we might land for a production recommendation.
+
+## Requirements
+
+* The firecracker-containerd agent, configured to start on boot
+* runc
+* Standard Linux filesystem mounts like procfs and sysfs
+* A cgroup v1 filesystem configured and mounted, with the controllers you intend
+  to use for your containers (typical controllers include things like blkio,
+  cpu, cpuacct, cpuset, devices, freezer, hugetlb, memory, net_cls, net_prio,
+  perf_event, pids, and rdma, though you may not use all of them)
+* Any supporting software you may be interested in leveraging inside the VM
+  (e.g., a local caching DNS resolver)
+* Any supporting dynamically-linked libraries necessary to run the components
+  inside (e.g., libc, libseccomp, etc)
+* (If sharing the root filesystem image among multiple VMs) Ability to run
+  successfully from a read-only device, in order to prevent a VM from
+  manipulating the filesystem in use by another VM.
+
+## Development tools
+
+We've built a tool to generate root filesystems that are suitable for
+development.  The tool can be invoked with `make image` from the root directory
+of this repository.  It will generate a Debian-based root filesystem as a
+squashfs filesystem with the firecracker-containerd agent configured to start
+through systemd, a runc binary built from the git submodule in the
+`_submodules/runc` folder, and support for multiple VMs by an overlay filesystem
+on top of the squashfs base.
+
+To minimize on-disk overhead, we start with the "minbase" debootstrap
+configuration, and only add minimal dependencies on top of it. We ensure that
+obviously extraneous files such as the content of `/usr/share/doc` and
+`/var/cache/apt` are removed. Further, we reduce startup latency by defining a
+custom systemd target that minimizes the services that systemd tries to start at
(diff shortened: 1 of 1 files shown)


docs/root-filesystem.md
```diff
@@ -0,0 +1,52 @@
...
+## Further thoughts
+
+For production use, we want to achieve the following goals:
+
+* Safe to share between multiple VMs
+* Fast to initialize and start the agent
+* Reasonably pared-down set of installed software
+
+The current development filesystem meets the first goal already.  We have not
+yet started on the second and third goals.  We're considering a few things, but
```
nmeyerhans: We can certainly consider more aggressive optimizations at some point, but to say that we haven't done any is not quite accurate. For example, to minimize on-disk overhead, we start with the "minbase" debootstrap configuration, and only add minimal dependencies on top of it. We ensure that obviously extraneous files such as the content of `/usr/share/doc` and `/var/cache/apt` are removed. Further, we reduce startup latency by defining a custom systemd target that minimizes the services that systemd tries to start at boot.

samuelkarp: Thanks for the corrections!  Will update.

docs/root-filesystem.md
```diff
@@ -0,0 +1,52 @@
...
+* Standard Linux filesystem mounts like procfs and sysfs
+* A cgroup v1 filesystem configured and mounted, with the controllers you intend
+  to use for your containers (typical controllers include things like blkio,
+  cpu, cpuacct, cpuset, devices, freezer, hugetlb, memory, net_cls, net_prio,
+  perf_event, pids, and rdma, though you may not use all of them)
+* Any supporting software you may be interested in leveraging inside the VM
+  (e.g., a local caching DNS resolver)
+* Any supporting dynamically-linked libraries necessary to run the components
+  inside (e.g., libc, libseccomp, etc)
+* (If sharing the root filesystem among multiple VMs) Ability to run
```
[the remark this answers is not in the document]
samuelkarp: Fixed

docs/root-filesystem.md
samuelkarp: Updated.  I incorporated a large part of your comment in the text in the above section, since it seemed like a good description of the work that has already been done for the image.

docs/root-filesystem.md
```diff
@@ -0,0 +1,60 @@
...
+
+firecracker-containerd requires a kernel and root filesystem to be specified for
+the Firecracker virtual machine where containers run.  This document specifies
+the hard requirements for that root filesystem, the development tools we built,
+and some thoughts around where we might land for a production recommendation.
+
+## Requirements
+
+* The firecracker-containerd agent, configured to start on boot
+* runc
```
sipsma: (This is more of a side-thought, up to you whether it's worth mentioning as a possibility in this doc)

I feel like we could get rid of the requirement for runc with only a moderate amount of effort. In theory, agent can just have runc statically compiled in and should be able to just use libcontainer rather than shell out to the separate runc binary. This would require some upstream changes, but nothing extraordinary.

This would be nice in that we can just drop the runc requirement entirely and only ask that users include the agent binary. It would also prevent any issues where agent and runc version get out of sync, should that possibility ever arise.

docs/root-filesystem.md
```diff
@@ -0,0 +1,60 @@
...
+`/var/cache/apt` are removed. Further, we reduce startup latency by defining a
+custom systemd target that minimizes the services that systemd tries to start at
+boot.
+
+More information on the tool and how to use it can be found in [the README.md
+file](../tools/image-builder/README.md).
+
+## Further thoughts
+
+For production use, we want to achieve the following goals:
```
sipsma: Have we put any consideration into size limitations on the rootfs? By which I mean the maximum amount of data that should be allowed to be written to the upper layer of the overlayfs.

I don't think we don't need to decide on numbers, but it's worth mentioning how that limit is (or isn't) enforced.

sipsma: Should we care about disk I/O performance inside the VM? It seems like it would be good to at least set ballpark expectations for users even if it's just "do not expect good performance; it's not being optimized for".

docs/root-filesystem.md
samuelkarp: I've added this as a question.

My inclination is to keep them separate so that it's easier to patch runc, but I don't have a strong opinion yet.

docs/root-filesystem.md
samuelkarp: Added a question.

samuelkarp: Added a question.

Issue #180, opened 2019-05-08
Consider using TTRPC for all containerd services+plugins 


sipsma:
Containerd recently added support for TTRPC plugins: https://github.com/containerd/containerd/commit/a8a805cad39960b437a0a32afec5f434423a0e93

This should allow us to replace the GRPC FCControl plugin with a TTRPC-based plugin, which in turn would allow us to no longer generate service definitions for both GRPC and TTRPC, instead using purely TTRPC. There may also be some performance benefits to the switch.

At the moment, the TTRPC support doesn't appear to be in tagged release and has a few rough edges, but a quick trial of switching the FCControl plugin to use TTRPC showed it to basically work.

Some of the rough edges we'd probably want to be addressed (perhaps through upstream contributions) before doing this switch:
1. The TTRPC address appears to be hardcoded in containerd code to be `<containerd grpc address>.ttrpc`. Ideally, this should be something clients can be told about rather than require them to also hardcode `+".ttrpc"`.
2. You can create a GRPC client by just doing `containerd.New(containerdAddress)`, but there doesn't appear to be an equivalent method for TTRPC, so you have to do the `net.Dial` yourself. Ideally, there should be a `containerd.NewTTRPC(address)` or something.

PR #196, opened 2019-06-04
Use TTRPC for plugin

mxpv:
*Issue #, if available:*
https://github.com/firecracker-microvm/firecracker-containerd/issues/180

*Description of changes:*
Use TTRPC in firecracker-control plugin, so we no longer have to maintain two versions of `Firecracker` service. 


examples/taskworkflow.go
```diff
@@ -56,30 +59,32 @@ func main() {
...
+	ctx := namespaces.WithNamespace(context.Background(), namespaceName)
 	image, err := client.Pull(ctx, "docker.io/library/nginx:latest",
 		containerd.WithPullUnpack,
 		containerd.WithPullSnapshotter("firecracker-naive"),
 	)
 	if err != nil {
 		return errors.Wrapf(err, "creating container")
+	}
 
+	pluginClient, err := ttrpcutil.NewClient(containerdAddress + ".ttrpc")
```
sipsma: nit: could add a `const containerdTTRPCAddress = containerdAddress + ".ttrpc"` at the top

examples/taskworkflow.go
```diff
@@ -56,30 +59,32 @@ func main() {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to create ttrpc client")
 	}
 
-	fcClient := fccontrol.NewFirecrackerClient(client.Conn())
+	fcClient := fccontrol.NewFirecrackerClient(pluginClient.Client())
```
sipsma: Would it be possible to change `NewFirecrackerClient` to accept the ttrpc socket path instead of a pre-existing ttrpc client? So it would be like `fcClient := fccontrol.NewFirecrackerClient(containerdAddress + ".ttrpc")`?

That's will make things a little simpler for users of the client who aren't familiar with how it works. We could also have a separate `NewFirecrackerClientFrom(existingTTRPCClient)` to maintain the existing interface in case that's ever needed too (though I'm not sure it is in our codebase at this time).

mxpv: The client is [auto generated](https://github.com/containerd/ttrpc/blob/a5bd8ce9e40bc7c065a11c6936f4d032ce6bfa2b/plugin/generator.go#L111) by ttrpc. While I see your point, I don't think it worth changing it manually or submitting a PR with breaking changes.

sipsma: Ah right, can we add a util function somewhere else then? It just doesn't make sense to force users to always jump through these multiple hoops in order to create a client when they are going to be doing the same thing pretty much every time.

mxpv: Discussed privately, will add a client wrapper helper in a separate PR.

PR #199, opened 2019-06-05
Extending timeout for snapshotter tests

xibz:
Extends snapshotter tests timeout. I ran several tests today on this on my CI box, but none of my builds were failing and were succeeding within 5 minutes. However, I did encounter some timing issues yesterday and was able to resolve them by simply extending the timeout which mostly succeeded within 15 minutes but had one instance where a single test took 23 minutes...


mxpv:
Are there specific tests that take long?

xibz:
@mxpv - Looks like the test still failed. It's the naive snapshotter with the containerd test suite. I am thinking we remove the devmapper and naive snapshotter and use the containerd devmapper instead. Then we don't even need to test any of the snapshotters, which alleviates this problem of the naive snapshotter being too slow.

samuelkarp:
> I am thinking we remove the devmapper and naive snapshotter and use the containerd devmapper instead.

@xibz I'd like to wait until a version of containerd that includes the devmapper snapshotter is released before we remove it from here.

PR #204, opened 2019-06-10
Disable golang-ci linting on tests

xibz:
golang-ci causes some issues with some test patterns, like using t.Run
and passing variables from the parent scope which is commonly done with
test vectors in Go.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,3 +1,6 @@
+run:
+  tests: false # disable running linters on tests
+
 linters:
   disable-all: true
   enable:


samuelkarp:
At least for now I'd prefer to work around the issues with the linter and `t.Run` instead of disabling it.

```go
for _, tc := range tests {
	tc := tc
	t.Run(tc.Name, func(t *testing.T) {
		// use tc here
	}
}
```

See https://github.com/firecracker-microvm/firecracker-containerd/blob/8fd1217fcd14c4d7bb6c8d95a120815a8c382018/runtime/helpers_test.go#L106

xibz:
@samuelkarp - It disables it on tests only. That solution, imo, is very hacky and every time we write a vector test that uses `t.Run` we will need to do this. I don't think this brings too much to tests anyways.  I don't see it ever being fixed in the linter due to the linter would need to understand the code's semantics. One of the maintainers said in the thread that maybe scopelint shouldn't be used regularly.

Issue #195, opened 2019-06-03
Define+enforce validation of VMID, Namespace and other API inputs


sipsma:
[VMID is currently unvalidated in terms of length, invalid characters, etc.](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/firecracker-control/local.go#L82).

We should do some validation; it is used to construct filesystem paths and abstract unix socket paths, which is a starting point for defining length contraints and invalid characters (i.e. we probably don't want to allow anything that would result in `../../..` or similar).

We also construct paths from the [containerd namespace](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/internal/vm/dir.go#L37), [container/task ID](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/internal/vm/dir.go#L92), and potentially other parameters provided by clients. We need to do validation on those too (probably even if containerd currently has some form of validation on them for an extra layer of assurance). We should audit our codebase for any other examples of user inputs being used to construct paths.

samuelkarp:
Reopening since #205 only covered the VM ID and not the containerd namespace or container/task ID inputs.

mxpv:
Addressed in https://github.com/firecracker-microvm/firecracker-containerd/pull/218 . Now it validates all path related IDs (vm id, container id, and namespace).

PR #205, opened 2019-06-11
Add VM ID validator

mxpv:

*Issue #, if available:*
https://github.com/firecracker-microvm/firecracker-containerd/issues/195

*Description of changes:*
This PR adds VM ID validation to the plugin.

### DIFF
--- a/firecracker-control/local.go
+++ b/firecracker-control/local.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/containerd/containerd/identifiers"
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/namespaces"
 	"github.com/containerd/containerd/plugin"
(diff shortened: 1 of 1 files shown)


firecracker-control/local.go
```diff
@@ -321,3 +324,26 @@ func (s *local) newShim(ns, vmID, containerdAddress string, shimSocket *net.Unix
 func isEADDRINUSE(err error) bool {
 	return err != nil && strings.Contains(err.Error(), "address already in use")
 }
+
+// validateID checks ID is valid and safe to use as a part of a file system path
+func validateID(id string) error {
+	const (
+		maxIDLen          = 255 // Maximum filename length is 255 characters on ext4
+		invalidCharacters = "\t\n\\/.:?* "
```
samuelkarp: So I don't think most of those are actually invalid.  But I'd like to propose an alternate approach here of defining a limited, valid character set rather than an invalid character set.  Something like this:  `[A-Za-z9-0-_]`.

sipsma: We could also consider re-using containerd's code for identifier validation: https://github.com/containerd/containerd/blob/6ba92a7921c558012c9ca87f2ca82432a683a527/identifiers/validate.go

samuelkarp: I'd be comfortable with containerd's validator as well.

mxpv: That's cool. I didn't notice containerd has it. Done.

PR #211, opened 2019-06-15
Exec support

### LINKED ISSUE #77: Support exec
We need to be able to support `task.Exec` for executing additional processes inside the containers.

sipsma:
**Note to Reviewers**: There's two commits here; the first actually resolves a separate issue that I discovered while in the middle of adding Exec support. I'm including it in the same PR here because all the choices made while fixing it had a huge impact on what the code for Exec would end up looking like.

See the individual commit messages for more details.

Closes #77 


sipsma:
FYI: I disabled a single linter check "context.Context should be the first parameter of a function" because it was getting applied incorrectly to functions like this one: https://github.com/sipsma/firecracker-containerd/blob/exec/internal/vmio/vsock.go#L85

This function is in fact using a context.Context as it's first parameter, it just so happens the second parameter is also a context.Context object. This seems like a bug in the linter.

internal/vmio/vmio.go
```diff
@@ -0,0 +1,56 @@
...
+//
+// If any error occurs (in synchronous or async code), the function should
+// return it within a call to errG.Go to ensure its propogated to the caller
+// successfully.
+type IOConnector func(
+	reqCtx context.Context,
+	shimCtx context.Context,
+	logger *logrus.Entry,
+	errG *errgroup.Group,
+) <-chan io.ReadWriteCloser
```
mxpv: While I understand the motivation here and how this code evolved over time, this looks way too complex and scary from the perspective of adding new connectors (which we're going to add). This turned into a leaky abstraction. An idiomatic way in golang is not to expose concurrency primitives in APIs, leaving this decision for API user. While it works for one implementation, it might now work for another one. Also having two context parameters is bad. I think a good example of a similar solution would be the `IO` interface from `go-runc` (https://github.com/containerd/go-runc/blob/9007c2405372fe28918845901a3276c0915689a1/io.go#L25), which faces a similar problem. It's not a call to action considering the complexity, but rather my thoughts.

sipsma: >  This turned into a leaky abstraction. An idiomatic way in golang is to not to expose concurrency primitives in APIs, leaving this decision for API user.

We are the API user here though. This is not an attempt to create a general purpose library (it's under the `internal` pkg); it's just meant to adapt other libraries (the `fifo` package, [adapted here](https://github.com/sipsma/firecracker-containerd/blob/ef8f0b7984c624a96df3fec69c2da6cb63d57ad7/internal/vmio/fifo.go#L59), and the `vsock` package, [adapted here](https://github.com/sipsma/firecracker-containerd/blob/ef8f0b7984c624a96df3fec69c2da6cb63d57ad7/internal/vmio/vsock.go#L70)) to our particular IO setup needs.

The problem being solved requires we sometimes do parts of setup synchronously, sometimes do the setup asynchronously, sometimes perhaps a mix of both. I don't know how we can do that without having some concurrency primitives in the our internal interface. Open to suggestions though of course.

> I think a good example of a similar solution would be the IO interface from go-runc

Our situation is much more complex than the RunC interface; they just need to define a set of io endpoints for a single process. The `IOConnector` interface defines a way of connecting two sets of io endpoints for independent processes that are using arbitrarily different types of IO, additionally taking into account the requirement for a mix of both synchronous and asynchronous setup. I agree it's more complicated but it's because the problem is inherently much more complicated.

What specifically makes adding more implementations of a connector complex+scary? The idea is just that you do synchronous setup as normal, do asynchronous setup inside the provided errgroup and return a channel on which the final result will be published when ready. Like I said before, the code already has multiple examples of adapting functions that know nothing about `IOConnector` to the interface.

Please don't take this as just dismissing the concerns; I'm completely open to suggestions for simplifying in a way that also maintains the requirements we have. Also If there's anything I just can document more clearly or something like that too let me know.

mxpv: > our particular IO setup needs.

This are needs of the `vsock` implementation. And this complexity leaks to the interface, so other implementations have to carry the overhead. Ideally we should expose just `ReadWriteCloser` instead.

> What specifically makes adding more implementations of a connector complex+scary? 

Specifically even an empty implementation brings a channel, adds errgroup, 2 contexts and func callback. Far more than I would expect from an entity that should expose an interface close to `ReadWriteCloser`. 

I can think of an alternative implementation for vsock (compiled in my head).

```go
type Vsock struct {
	read *io.PipeReader
}

func NewVsock(port uint32) io.ReadCloser {
	reader, writer := io.Pipe()

	go func() {
		listener, err := vsock.Listen(port)
		if err != nil {
			return err
		}

		conn, err := listener.Accept()
		if err != nil {
			return err
		}

		// close writer on timeout
		// same trick for WriteCloser
		// use ctx for cancellation

		defer conn.Close()

		err = io.Copy(writer, conn)
		if err != nil {
			return err
		}
	}()

	return &Vsock{read: reader}
}

func (v *Vsock) Read(p []byte) (n int, err error) {
	return v.read.Read(p)
}

func (v *Vsock) Close() error {
	return v.read.Close()
}

```

sipsma: The needs are not just for vsock. They are for vsock, fifo and arbitrary future implementations we may need to add, which all together require a mix of sync and async setup and read/writes. What we need isn't just another `ReadWriteCloser` implementation; we are already provided those by the various libraries we are utilizing. What we need is an interface for **initializing** arbitrary `ReadWriteCloser`s and copying between them.

The sketch you wrote out would be nice, but it doesn't deal with the real-world complications that result in the `IOConnector` interface. The major ones being:
1. We want `CreateTask` and `ExecProcess` to fail if IO fails to initialize. That means that the errors returned by `Listen` and `Accept` in the example above need to be returned to the caller. However, we also need the setup to be asynchronous. If we made all setup synchronous, we actually create deadlock as the IO initialization depends on the `Create`/`Exec` call being made and the `Create`/`Exec` call depends on IO initializing successfully.
   * This combination of a need for doing something optionally asynchronously but also returning an error to the caller is where the `errgroup` comes from. The `errgroup` also allows the caller to block until the initialization is complete.
   * Like the existing examples show, you can do all sorts of asynchronous stuff in the `errgroup` or you can just do synchronous stuff and only use the `errgroup` to report the error (with an overhead of one line of code). It's a really nice, flexible interface in that respect. I don't think it's much of a cognitive burden.
2. We should really, really try to avoid increasing the amount of IO buffering we're doing as we already need to copy from task->agent->runtime->client for stdout and stderr. With an extra `io.Pipe()` that's now grown to task->io.Pipe->agent->io.Pipe->runtime->io.Pipe->client. If we have hundreds of VMs running and tasks are pumping out large amounts of data on stdout/stderr, all that extra buffering and copying is going to start adding up in terms of CPU, latency and memory.
   * So instead of creating an extra layer on top of `ReadWriteCloser`s, we just use a channel to return one of the existing implementations from the `fifo` and `vsock` libraries. This is where the returned channel comes from.
   * This has another nice side benefit of allowing us to re-use optimizations such as [this one in the golang `io` library that check if the object provided has any other methods that could result in better performance](https://github.com/golang/go/blob/7a4d02387fa16cd2a88c30357346e5cf0ae282b1/src/io/io.go#L381-L389). Instead of trying to keep up with every interface possible by adding `Read`, `Write`, `ReadFrom`, `WriteFrom`, etc., we just re-use what already exists by not wrapping objects and using standard golang channels.

For the contexts, looking at this with a fresh pair of eyes after the weekend, I see a way to have just one.
   * Originally, I did just have the `reqCtx`, which is all you think you'd need. However, that actually resulted in occasional race conditions with the `fifo` package implementation because it uses the context provided to cancel its own internal async initialization. If that didn't happen by the time the `reqCtx` got cancelled after the request was done, then you'd get mysterious errors about read/writing from an already closed FIFO. This is why the `shimCtx` got added.
   * What I'll do instead is provide just the single `vmProc.ctx` object. I already have code that cancels that context if `Create` or `Exec` fails, so it should be an adequate replacement for both of the existing contexts. I'll update with this.

So, given the above, I can get rid of one of the contexts, which leaves a context, an errgroup for reporting IO initialization errors and a channel for returning the `ReadWriteConnector`. I think each one of these are completely justified and gives us a flexible interface that allows us to plug in arbitrary implementations in the future without needing to always go back and refactor the runtime/agent implementations to deal with new async/sync setup needs.

I'm still of course open to suggestions that further simplify things without compromising on the points above and will think about it myself too, but ultimately I just don't see how an `errgroup` (which is just a light wrapper around a `sync.WaitGroup`) and a channel are that much of burden given the complications of our requirements. They are just two simple synchronization primitives that each serve a specific need.

mxpv: > That means that the errors returned by Listen and Accept in the example above need to be returned to the caller. 

Have a look on pipe's `CloseWithError` ([link](https://github.com/golang/go/blob/530097fe60c3d9feffabf52ea910bf35af9da6fb/src/io/pipe.go#L167)). You can use to to propagate an error to the caller. So when a caller tries to Read, it can receive  an error from `Accept` or `Listen` as well.

>  avoid increasing the amount of IO buffering

`io.Pipe` doesn't perform IO buffering ([link](https://github.com/golang/go/blob/530097fe60c3d9feffabf52ea910bf35af9da6fb/src/io/pipe.go#L181)). Data is copied directly.

> What we need is an interface for initializing

I think that should be done in `NewXXX` of every concrete implementation, but not in the interface.

> If we made all setup synchronous, we actually create deadlock as the IO initialization depends on the Create/Exec call being made and the Create/Exec call depends on IO initializing successfully.

We still can make it async if necessary (see my sketch above) + add some timeout mechanism. 
Let's say we do Listen, Accept and write to pipe. Or timeout if we're not writing to the pipe within X seconds. In this case a caller tries to `Read` and receives either data or error timeout, that we send through `CloseWithError`.

sipsma: > Have a look on pipe's CloseWithError
> ...
> In this case a caller tries to Read and receives either data or error timeout, that we send through CloseWithError.

Synced up offline about this too, but I don't see how we can use `CloseWithError`. We need to check if there was an error initializing the `ReadWriteCloser` and then use it if there wasn't an error. We can't wait until the actual Read/Write happens as that call can block indefinitely in the case of success and thus block Create/Exec from succeeding indefinitely.

> io.Pipe doesn't perform IO buffering

Ah I see now, that's really cool! Thanks for linking to the implementation. It uses a bunch of channels and other synchronization internally which look like they could essentially replace the channel we are creating directly. The function signature would thus change from
`func(context.Context, *logrus.Entry, *errgroup.Group) <-chan io.ReadWriteCloser`
to
`func(context.Context, *logrus.Entry, *errgroup.Group) io.ReadWriteCloser`

However, it doesn't eliminate the need for the `errgroup` AFAICT and probably results in more code being written for each implementation of the `IOConnector` func as now each one is doing an `io.Copy` in addition to the `io.Copy` that still happens between the pipes (though I understand now that doesn't actually imply more buffering). If the only benefit is that the `io.Pipe` call is making a channel for us instead of us just making it directly, I'm not convinced it's actually worth it given the extra coding overhead, but I will give it a shot to see what it looks like in practice. Will get back on this.

> I think that should be done in NewXXX of every concrete implementation, but not in the interface

This pattern isn't being used here because we technically aren't using an golang interface per-se, it's just a function signature type definition. This is because there's no need for structs to hold any persistent data.

sipsma: I gave `io.Pipe` a try here: https://github.com/sipsma/firecracker-containerd/commit/e209fdc5a0f7ed9b695570541d2d05e8361ec4eb

The only step of any significance that was eliminated was [this wait on the `ReadWriteCloser` channel here in `proxyIO`, not in the `IOConnector` implementions](https://github.com/sipsma/firecracker-containerd/commit/e209fdc5a0f7ed9b695570541d2d05e8361ec4eb#diff-f1cc210471d7d8666b600a626cdf9f84L341).

Besides that, you still have to do all the same stuff as the previous implementation except now:
1. Each implementation of the `IOConnector` func has to take care of setting up copying for the io.Pipe. I'm not seeing that this is anything but boilerplate as we still need to do the same `io.Copy` as before in `startStdioProxy`, resulting in 3 calls to `io.Copy` for every stream in every process (previously it was just once, with none in the `IOConnector` funcs).
2. There's more complication involved in thinking about when each end of every pipe and connection should be closed (and in what order so data can't get dropped)

The downsides are not the end of the world, but I don't see that the switch to `io.Pipe` provides any benefits either. I think it would have worked out if we didn't still need the `errgroup`, but like I said before AFAICT there's nothing in the `io.Pipe` interface that can replace that without creating deadlock or race conditions.

I'm going to leave it as in the PR for now, let me know if there's any objections or other ideas.

mxpv: That's fine, I don't want this PR to be blocked. I'll play with `IO` and submit a separate PR if I manage to improve this.

sipsma: I did some more thinking about this and made two more improvements that seem to work and result in a slightly easier to understand interface:
1. I just have `IOConnector` return two channels without being passed an errgroup. The synchronization ends up being exactly the same, it just results in only having to think about channels when writing an `IOConnector` func, which is a little bit nicer.
2. I wrapped `IOConnectorSet` in an interface called `IOProxy`. The intention here was to support completely different implementations of `IOProxy` that don't even use `CopyBuffer`. My hope is this makes it easy to plug in your support for Binary logging (it can just be a different `IOProxy` implementation), but worst case scenario it should just have no effect on your efforts there as it didn't change any functionality.

internal/vm/ioproxy.go
```diff
@@ -0,0 +1,186 @@
...
+func (connectorPair *IOConnectorPair) proxy(
+	proc *vmProc,
+	logger *logrus.Entry,
+	timeoutAfterExit time.Duration,
+) (ioInitDone <-chan error, ioCopyDone <-chan error) {
+	initDone := make(chan error)
+	copyDone := make(chan error)
+
+	inCh, inErrCh := connectorPair.ReadConnector(proc.ctx, logger.WithField("direction", "read"))
+	outCh, outErrCh := connectorPair.WriteConnector(proc.ctx, logger.WithField("direction", "write"))
```
mxpv: Just for my understanding. When error channel returns `nil`. Does it mean that `ReadWriteCloser` object is ok for use?

sipsma: > When error channel returns nil. Does it mean that ReadWriteCloser object is ok for use?

Yes, and if the error channel returns a non-nil object than the ReadWriteCloser should be nil. Just like the normal pattern of return values for go functions.

> In other words would this be an equivalent to chan Pair<ReadWriteCloser, error> ?

I don't think we'd be able to do that actually as the initialization finishing and the copying happening are two different events that happen at different times. You need two separate channels to communicate when those different events happen.

internal/vm/task.go
```diff
@@ -15,322 +15,269 @@ package vm
...
+	}
 
-		// All FIFO+vsocks should be closed by this point in the previous switch cases.
-		if err != nil {
-			logger.WithError(err).Error("error with stdio")
-		}
-	}()
+	// Wait for the io streams to be flushed+closed before returning. Error is ignored
+	// here as any io error is already logged in the IOProxy implementation directly
+	<-proc.ioCopyDone
```
xibz: Should we have a timeout here in case we are waiting indefinitely?

internal/vm/vsock.go
```diff
@@ -0,0 +1,138 @@
...
+					return
+				default:
+					// accept is non-blocking so try to accept until we get a connection
+					conn, err := listener.Accept()
+					if err == nil {
+						returnCh <- conn
+						return
+					}
+
+					if isTemporaryNetErr(err) {
```
xibz: nit: I think would much prefer if we had the temporary logic stuff outside of the if statement or vice versa

```go
if isTemporaryNetErr(err) {
    logger.WithError(err).Debug("temporary stdio vsock accept failure")
    continue
}
   
logger.WithError(err).Error("non-temporary stdio vsock accept failure")
errCh <- err
return
```

internal/vmio/vsock.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}
+
+	logger.WithError(lastErr).WithFields(logrus.Fields{"context_id": contextID, "port": port}).Error("vsock dial failed")
+	return nil, lastErr
+}
+
+// VSockDialConnector provides an IOConnector interface to the VSockDial function.
+func VSockDialConnector(contextID, port uint32) IOConnector {
+	return func(procCtx context.Context, logger *logrus.Entry, errG *errgroup.Group) <-chan io.ReadWriteCloser {
```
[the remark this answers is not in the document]
sipsma: I would use a struct if the parameters (`contextID` and `port`) needed to be accessed by different functions or multiple times, but they are only used by a single function for a single invocation.

If we made this a struct, AFAIK it would still be allocated on the heap because it is passed far out from the original function call (to other goroutines).

internal/vm/vsock.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+// VSockDial attempts to connect to a vsock listener at the provided cid and port with a hardcoded number
+// of retries.
+func VSockDial(reqCtx context.Context, logger *logrus.Entry, contextID, port uint32) (net.Conn, error) {
+	// VM should start within 200ms, vsock dial will make retries at 100ms, 200ms, 400ms, 800ms, 1.6s, 3.2s, 6.4s
+	const (
+		retryCount      = 7
+		initialDelay    = 100 * time.Millisecond
+		delayMultiplier = 2
```
samuelkarp: I don't really know that we need exponential backoff for a vsock dial.  We expect exactly one client to be attempting to dial anyway and we're not implementing throttling anywhere.  I think a fixed delay between retries makes enough sense here.

internal/vmio/vsock.go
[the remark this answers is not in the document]
xibz: Okay, that makes sense. And yes, the struct would also be allocated on the heap, but just once. This has two allocations, although minor it is good to know!

internal/vm/vsock.go
sipsma: Agreed, added a commit with that and also some other cleanup of the related functions (like adding a timeout to the `vsockAccept` function).

internal/vm/ioproxy.go
```diff
@@ -0,0 +1,214 @@
...
+			logger.WithError(err).Error("error copying io")
+			copyDone <- err
+		}
+	}()
+
+	return initDone, copyDone
+}
+
+type ioConnectorSet struct {
+	stdin  *IOConnectorPair
```
xibz: Can this be an array of `IOConnectorPair`s instead?

sipsma: In theory sure but you would need to specify still that index 0 of the slice is stdin, index 1 is stdout, index 2 is stderr because there can be branching behavior for the different streams, i.e: https://github.com/sipsma/firecracker-containerd/blob/611800875f31dbb4f210561017a0b3e4ff2e19f5/internal/vm/ioproxy.go#L164

Given that, it seems like a lot more straightforward to just have 3 specific fields. I think it's highly unlikely we would want to support more logging FDs in the future in this context, so our code shouldn't need to handle arbitrary numbers.

xibz: That branching logic could still work with an array, no? All you care about is the proxy function and passing in a name, which could be stored on the connector set.

sipsma: Sure, but what does that get us relative to just using a field for each of std<in/out/err>? It seems like it is just a more roundabout way of specifying them

xibz: It cleans up the code, so you can put it in a for loop. It generalizes it making it easier to add things in the future, if needed.

samuelkarp: > It generalizes it making it easier to add things in the future, if needed.

At this point, I don't really anticipate needing additional IO streams or FDs passed to exec'd processes.  Arbitrary FDs would be fairly difficult across a vsock.

Named fields here are clearer in the code that uses them; you can see immediately that it's `stderr` instead of `io[2]` and need to remember the mapping of file descriptors to their common name.

sipsma: I can make a for loop now with
```
for stream := range []*IOConnectorPair{set.stdin, set.stdout, set.stderr} {
   ...
}
```

but then I need to add more fields to the struct (for the name of stream and what timeouts to use) just to save a total of probably ~10 lines of code. I agree with Sam, I don't feel like that results in more clarity.

xibz: It's just for setup and the code is identical between the different io streams. I'm more for putting this in an array, as I think having them named doesn't yield anything in this context. However, since there seems to be more in favor of keeping it the way it is, I'll go ahead and approve this PR.

sipsma:
> The only thing I'd still like to see is unit tests around the new IO interfaces, but I'm comfortable having that in a follow-up PR instead of this one.

I will give this a shot in a separate PR for the `FIFOConnector`, `VSockDialConnector` and `VSockAcceptConnector` functions. The other code in `ioproxy.go` should already be covered as part of the unit tests in `task_test.go` and the e2e test in `service_integ_test.go`.

sipsma:
Rebased this on top of the stub drive PR

PR #212, opened 2019-06-26
runtime: Add error messages in integration test failures

xibz:
If an integration test fails with a non-zero exit code, we should log
what error occurred

### DIFF
--- a/runtime/service_integ_test.go
+++ b/runtime/service_integ_test.go
@@ -295,6 +295,7 @@ func TestMultipleVMs_Isolated(t *testing.T) {
 							fmt.Printf("stderr output from vm %d, container %d: %s", vmID, containerID, stderrOutput)
 						}
 
+						require.NoError(t, exitStatus.Error(), "failed to retrieve exitStatus")
 						require.Equal(t, uint32(0), exitStatus.ExitCode())
 						require.Equal(t, vmIDtoMacAddr(uint(vmID)), strings.TrimSpace(stdout.String()))
 					case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


runtime/service_integ_test.go
```diff
@@ -295,7 +295,7 @@ func TestMultipleVMs_Isolated(t *testing.T) {
 							fmt.Printf("stderr output from vm %d, container %d: %s", vmID, containerID, stderrOutput)
 						}
 
-						require.Equal(t, uint32(0), exitStatus.ExitCode())
+						require.Equal(t, uint32(0), exitStatus.ExitCode(), fmt.Sprintf("failed with %v", exitStatus.Error()))
```
samuelkarp: `require.Equal` should already be outputting both the expected and actual values; that's part of why we use it over `if exitStatus.ExitCode() != 0`.  Are you not seeing this output?

sipsma:
[The `.Error()` method appears to only return a non-nil error if something went wrong during the `Wait()` call's asynchronous code](https://github.com/containerd/containerd/blob/9990eed3149a23291032d49ddb2607ce21bc7b69/process.go#L150). So I don't think we want to log `.Error()` when checking the ExitCode as they aren't actually related.

We should however definitely be checking that `.Error()` is non-nil in the line above the check for the ExitCode, which was previously missed.

runtime/service_integ_test.go
```diff
@@ -295,6 +295,7 @@ func TestMultipleVMs_Isolated(t *testing.T) {
 							fmt.Printf("stderr output from vm %d, container %d: %s", vmID, containerID, stderrOutput)
 						}
 
+						require.NoError(t, exitStatus.Error())
```
samuelkarp: I'd suggest doing `require.NoError(t, exitStatus.Error(), "failed to retrieve exitStatus")`.

PR #216, opened 2019-06-27
Allow stdio redirection inside VM 

### LINKED ISSUE #78: Allow stdio redirection inside VM
Some use-cases are better served by having stdio contained entirely within the microVM, instead of crossing the VM<->host boundary through vsock.

mxpv:
*Issue #, if available:*
closes #78 

*Description of changes:*
These are minimal changes needed in order to support binary-io as per our discussion.


internal/vm/agent.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"net/url"
+)
+
+// IsAgentOnlyIO checks whether Stdout target is agent only in order to allow
+// log redirection entirely within the VM (currently cio.BinaryIO and cio.LogFile)
+func IsAgentOnlyIO(stdout string) bool {
+	parsed, err := url.Parse(stdout)
+	if err != nil {
```
sipsma: I'm guessing that most of the errors returned here are expected (`stdout` is just not a url), but should we at least log this error at Debug in case we get into a situation in the future where this starts returning unexpected errors?

mxpv: Probably wouldn't hurt. Done.

sipsma:
It would be good to have some e2e test coverage for this too, but no need to be in this PR necessarily

PR #219, opened 2019-07-03
agent: Adding retry around mounting logic

### LINKED ISSUE #214: Occasional "invalid argument" error when mounting image drive inside VM
One example (it has been very hard to reproduce): https://buildkite.com/firecracker-microvm/firecracker-containerd/builds/570#732d0eb9-666f-4c7b-b926-ecaccb4a319c

It looks like this started happening intermittently after the stub drive PR was merged: https://buildkite.com/firecracker-microvm/firecracker-containerd/builds?branch=master

Based on the logs, it seems to be that Agent is encountering an "invalid argument" during Create. I'm not sure, but I suspect it was this line as it's the only one that was not wrapped in a descriptive error: https://github.com/firecracker-microvm/firecracker-containerd/blob/c74f45d4c659dd01c2df6c6bffac1d30d398ab8b/agent/service.go#L146

I've since updated that error to be wrapped (as part of the Exec changes), so we can confirm next time it's encountered. If it does turn out to be that line returning "invalid argument", it seems like there may be some occasional bug, possibly a race condition, involving the mounting of the container image drive inside the VM. More investigation is needed.

xibz:
Firecracker requires some time to populate the config space change which
was causing failures in our CI due to calling patch drive followed by
an immediate call of mounting that drive.

https://github.com/firecracker-microvm/firecracker/issues/1159#issuecomment-507984468

Fixes #214

### DIFF
--- a/agent/error.go
+++ b/agent/error.go
@@ -1,0 +1,25 @@
+// Copyright 2018-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package main
+
+import (
+	"syscall"
+)
+
+// isRetryableMountError will check to see if the error passed in is an
+// syscall.EINVAL
+func isRetryableMountError(err error) bool {
+	errno, ok := err.(syscall.Errno)
+	return ok && errno == syscall.EINVAL
+}
--- a/agent/service.go
+++ b/agent/service.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 	"runtime/debug"
 	"strings"
+	"time"
 
 	"github.com/containerd/containerd/cio"
 	"github.com/containerd/containerd/log"
@@ -154,7 +155,25 @@ func (ts *TaskService) Create(requestCtx context.Context, req *taskAPI.CreateTas
 		return nil, fmt.Errorf("Drive %q could not be found", driveID)
 	}
 
-	err = bundleDir.MountRootfs(drive.Path(), "ext4", nil)
+	const (
+		maxRetries = 100
+		retryDelay = 10 * time.Millisecond
+	)
+
+	// We retry here due to guest kernel needing some time to populate the guest
+	// drive.
+	// https://github.com/firecracker-microvm/firecracker/issues/1159
+	for i := 0; i < maxRetries; i++ {
+		if err = bundleDir.MountRootfs(drive.Path(), "ext4", nil); isRetryableMountError(err) {
+			logger.WithError(err).Warnf("retrying to mount rootfs %q", drive.Path())
+
+			time.Sleep(retryDelay)
+			continue
+		}
+
+		break
+	}
+
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to mount rootfs %q", drive.Path())
 	}
--- a/internal/bundle/bundle.go
+++ b/internal/bundle/bundle.go
@@ -60,9 +60,10 @@ func (d Dir) RootfsPath() string {
 	return filepath.Join(d.RootPath(), internal.BundleRootfsName)
 }
 
-// MountRootfs creates the RootfsPath and mounts the provided sourcePath there.
+// MountRootfs creates the RootfsPath if it does not exist and mounts the
+// provided sourcePath there.
 func (d Dir) MountRootfs(sourcePath string, mountType string, mountOpts []string) error {
-	err := os.Mkdir(d.RootfsPath(), 0700)
+	err := os.MkdirAll(d.RootfsPath(), 0700)
 	if err != nil {
 		return err
 	}
--- a/agent/error_test.go
+++ b/agent/error_test.go
@@ -1,0 +1,58 @@
+// Copyright 2018-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"syscall"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsRetryableMountError(t *testing.T) {
+	cases := []struct {
+		Name     string
+		Error    error
+		Expected bool
+	}{
+		{
+			Name:     "nil case",
+			Error:    nil,
+			Expected: false,
+		},
+		{
+			Name:     "syscall.Errno EINVAL case",
+			Error:    syscall.EINVAL,
+			Expected: true,
+		},
+		{
+			Name:     "syscall.Errno ENOENT case",
+			Error:    syscall.ENOENT,
+			Expected: false,
+		},
+		{
+			Name:     "regular error case",
+			Error:    fmt.Errorf("foo bar"),
+			Expected: false,
+		},
+	}
+
+	for _, c := range cases {
+		c := c // see https://github.com/kyoh86/scopelint/issues/4
+		t.Run(c.Name, func(t *testing.T) {
+			assert.Equal(t, c.Expected, isRetryableMountError(c.Error))
+		})
+	}
+}


agent/service.go
```diff
@@ -154,7 +155,25 @@ func (ts *TaskService) Create(requestCtx context.Context, req *taskAPI.CreateTas
 		return nil, fmt.Errorf("Drive %q could not be found", driveID)
 	}
 
-	err = bundleDir.MountRootfs(drive.Path(), "ext4", nil)
+	const (
+		maxRetries = 3
+		retryDelay = 100 * time.Millisecond
```
nmeyerhans: 100ms seems like quite a long time to wait for the first retry. How did you arrive at this number?

sipsma:
I'm not necessarily opposed to this solution, but I'm wondering if there is a more specific way for us to wait for the OS to rescan the device using syscalls instead of just retries.

It seems like there may be [a command line utility, partprobe, that does this](https://linux.die.net/man/8/partprobe) (if I'm understanding everything correctly), but I wouldn't want to take a dependency on that program existing in the VM.

I think it's worth seeing how that utility works; if it's something as simple as writing/reading a file from /sys or similar, I'd prefer that over trying to figure out an optimal retry interval+timeout.

agent/service.go
```diff
@@ -154,7 +155,25 @@ func (ts *TaskService) Create(requestCtx context.Context, req *taskAPI.CreateTas
...
+	)
+
+	// We retry here due to Firecracker needing some time to populate the guest
+	// drive.
+	// https://github.com/firecracker-microvm/firecracker/issues/1159
+	for i := 0; i < maxRetries; i++ {
+		if err = bundleDir.MountRootfs(drive.Path(), "ext4", nil); err != nil {
+			logger.WithError(err).Errorf("failed to mount rootfs %q", drive.Path())
```
samuelkarp: This should probably be a warning instead of an error since it gets retried.

agent/service.go
xibz: I was thinking originally 50ms, but decided on 100ms to be a little more safer. Are we happy with changing this to 50ms?

xibz:
@sipsma - I would agree if the waiting was inconsistent. However, the wait time between retries will be fairly small and at worst case would need to loop only once. I set the `maxRetries` to three just for a little more of a error range, but I doubt it'll ever hit that third loop.

agent/service.go
nmeyerhans: I suspect even 25 is more than enough. I'd maybe start with 10.

Ideally we profile the `virtio_config_changed` kernel function in various scenarios (different numbers of device attachments, etc) and gather some specific numbers on which to base this decision...

nmeyerhans:
> I'm not necessarily opposed to this solution, but I'm wondering if there is a more specific way for us to wait for the OS to rescan the device using syscalls instead of just retries.
> 
> It seems like there may be [a command line utility, partprobe, that does this](https://linux.die.net/man/8/partprobe) (if I'm understanding everything correctly), but I wouldn't want to take a dependency on that program existing in the VM.
> 
> I think it's worth seeing how that utility works; if it's something as simple as writing/reading a file from /sys or similar, I'd prefer that over trying to figure out an optimal retry interval+timeout.

Pretty sure that partprobe is just using a bunch of `ioctl` calls to get information about drives it already knows about. We could try something similar (see some of the `BLK` ioctls in [include/uapi/linux/fs.h](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/fs.h) for example), but still, we're just polling. I'm not sure it's a whole lot better.

If there's a call we can make that'll block until a currently running `virtio_config_changed` call completes, that could be interesting, but I don't think there is such a function.

agent/service.go
xibz: @nmeyerhans - I feel writing that benchmark is a pretty large task, no? Basically you'd need to write C to benchmark the kernel function, but you'd also have to make some sort of stub linux kernel. Am I thinking about this right? Or did you have something else in mind?

sipsma:
> I would agree if the waiting was inconsistent. However, the wait time between retries will be fairly small and at worst case would need to loop only once. I set the maxRetries to three just for a little more of a error range, but I doubt it'll ever hit that third loop.

I don't think we can know if the waiting is consistent or not. We can't assume we know how fast the host/VM will be (they won't all be the same that are running our CI tests) or that there aren't still rare conditions in which even the hosts we test on still take more than 300ms. That's why I wanted to at least see if there's a better option.

> If there's a call we can make that'll block until a currently running virtio_config_changed call completes, that could be interesting, but I don't think there is such a function.

That's alright then. I saw for some block device types there is a `/sys/block/<name>/device/rescan` file, but poking around inside a FC VM it doesn't appear that's implemented for virtio. Given that, then I agree polling is the next best option.

agent/service.go
```diff
@@ -154,7 +155,25 @@ func (ts *TaskService) Create(requestCtx context.Context, req *taskAPI.CreateTas
...
+	const (
+		maxRetries = 3
+		retryDelay = 25 * time.Millisecond
+	)
+
+	// We retry here due to Firecracker needing some time to populate the guest
+	// drive.
+	// https://github.com/firecracker-microvm/firecracker/issues/1159
+	for i := 0; i < maxRetries; i++ {
+		if err = bundleDir.MountRootfs(drive.Path(), "ext4", nil); err != nil {
```
sipsma: Can we only do a retry if err is `EINVAL` so we don't retry when some unrelated error occurs?

agent/service.go
sipsma: I agree some actual metrics would be nice, but if we don't do the benchmarking right away, I'd ask we at least increase the total wait time to something more like 1 second (10 ms retry interval is good though). Just in an attempt to account for hosts that are slower and other unpredictable situations in which the VM takes more than 30ms to rescan the block device.

nmeyerhans: Agreed; in the common case, the rescan operation finishes very quickly. We often don't need a retry at all. When we do need a retry, 10ms is likely to be more than sufficient, and gives us a positive user experience (the task starts quickly). In the case of a particularly slow host, a full second (or more) of maximum retries will ensure that we do eventually succeed. The only drawback there is that it takes a bit longer for fatal errors to become apparent, but I am OK with that.

xibz: Fixed!

agent/service.go
xibz: Good call! Fixed

agent/service.go
```diff
@@ -154,7 +155,25 @@ func (ts *TaskService) Create(requestCtx context.Context, req *taskAPI.CreateTas
...
+		maxRetries = 100
+		retryDelay = 10 * time.Millisecond
+	)
+
+	// We retry here due to Firecracker needing some time to populate the guest
+	// drive.
+	// https://github.com/firecracker-microvm/firecracker/issues/1159
+	for i := 0; i < maxRetries; i++ {
+		if err = bundleDir.MountRootfs(drive.Path(), "ext4", nil); isRetryableMountError(err) {
+			logger.WithError(err).Warnf("failed to mount rootfs %q", drive.Path())
```
sipsma: nit: Including `will retry` or something similar in the error message would allay concerns for someone seeing that in the logs that's not familiar with the background here.

agent/service.go
```diff
@@ -154,7 +155,25 @@ func (ts *TaskService) Create(requestCtx context.Context, req *taskAPI.CreateTas
 		return nil, fmt.Errorf("Drive %q could not be found", driveID)
 	}
 
-	err = bundleDir.MountRootfs(drive.Path(), "ext4", nil)
+	const (
+		maxRetries = 100
+		retryDelay = 10 * time.Millisecond
+	)
+
+	// We retry here due to Firecracker needing some time to populate the guest
```
sipsma: nit: I think it's more accurate to say the guest kernel needs time to rescan the guest drive

agent/error.go
```diff
@@ -0,0 +1,24 @@
...
+// isRetryableMountError will check to see if the error passed in is an
+// syscall.EINVAL
+func isRetryableMountError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	errno, ok := err.(syscall.Errno)
+	if !ok {
+		return false
```
mxpv: This can be rewritten slightly shorter. Nil check not needed as in this case `ok` will be just `false`.
```go
func isRetryableMountError(err error) bool
    errno, ok := err.(syscall.Errno)
    return ok && errno == syscall.EINVAL
}
```

agent/service.go
nmeyerhans: Yes, the asynchronous operation that needs to complete is 100% in the guest kernel. Firecracker is out of the picture at this point.

internal/bundle/bundle.go
```diff
@@ -60,11 +60,15 @@ func (d Dir) RootfsPath() string {
...
 
-// MountRootfs creates the RootfsPath and mounts the provided sourcePath there.
+// MountRootfs creates the RootfsPath if it does not exist and mounts the
+// provided sourcePath there.
 func (d Dir) MountRootfs(sourcePath string, mountType string, mountOpts []string) error {
-	err := os.Mkdir(d.RootfsPath(), 0700)
-	if err != nil {
-		return err
+	rootfsPath := d.RootfsPath()
+	if _, err := os.Stat(rootfsPath); os.IsNotExist(err) {
```
mxpv: Just use `os.MkdirAll`

agent/service.go
xibz: Done!

agent/service.go
xibz: Ah, that's correct. Reworded!

agent/error.go
xibz: Fixed

internal/bundle/bundle.go
xibz: Good call! Fixed

agent/service.go
nmeyerhans: JFTR, I've done a small amount of benchmarking of the `virtio_config_changed` kernel function. When patching a single drive, it completes in roughly 0.01ms. So... 10ms ought to be more than sufficient, even with lots of drives.

PR #221, opened 2019-07-10
Adding support for v0.17.0 of the SDK

xibz:
This updates to use the newest version of the SDK which unconvered an
issue with one of the context that was passed during machine creation.
Originally it was using the request context which is a short lived
context and was causing the VM to be killed during execution. To fix
that the shim context is now used.


go.sum
```diff
@@ -46,6 +46,8 @@ github.com/docker/go-units v0.3.3 h1:Xk8S3Xj5sLGlG5g67hJmYMmUgXv5N4PhkjJHHqrwnTk
 github.com/docker/go-units v0.3.3/go.mod h1:fgPhTUdO+D/Jk86RDLlptpiXQzgHJF7gydDDbaIK4Dk=
 github.com/firecracker-microvm/firecracker-go-sdk v0.15.2-0.20190410183546-4f11ac73a965 h1:QKioOzqmg7DY/YDhGLYZsP8bq2T3BahVHdyPsf0uq/E=
 github.com/firecracker-microvm/firecracker-go-sdk v0.15.2-0.20190410183546-4f11ac73a965/go.mod h1:QcNsz2gYkcTI/zAQl3dYeLwf7KxtjKnt7aGklhn9yYk=
+github.com/firecracker-microvm/firecracker-go-sdk v0.15.2-0.20190627223500-b2e8284e890c h1:jAAxKV5SD/y2nNKNLHgAUmfgbrvfN4xBiKR+b2qK2pM=
```
sipsma: nit: We should probably run `go mod tidy` as it looks like there's still checksums in here for the older version of the firecracker-go-sdk.

go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/docker/go-events v0.0.0-20170721190031-9461782956ad // indirect
 	github.com/docker/go-metrics v0.0.0-20181218153428-b84716841b82 // indirect
 	github.com/docker/go-units v0.3.3
-	github.com/firecracker-microvm/firecracker-go-sdk v0.15.2-0.20190410183546-4f11ac73a965
+	github.com/firecracker-microvm/firecracker-go-sdk v0.15.2-0.20190627223500-b2e8284e890c
```
sipsma: Why is the version of the `v0.17.0` sdk still referenced as `v0.15.2` here? Can we update this to specify `v0.17.0`? Or is that a separate issue that would require changes in the firecracker-go-sdk? I don't remember how this versioning string works in go modules off the top of my head.

mxpv: @sipsma I suspect this is because there is no corresponding tag `v0.17.0` in firecracker-go-sdk. Latest releases there is v0.15.1. This is somewhat confusing.

sipsma: Okay makes sense. That's something we should resolve but doesn't need to be part of this PR.

go.sum
xibz: Fixed!

PR #223, opened 2019-07-15
Give task and exec FIFO dirs unique names in VM.

sipsma:
Before this, if the initial process of a task had the same ID as an exec in the
task, they would end up using the same FIFO dirs.

Note: This isn't related to any ongoing larger work, I just happened to notice it as I was looking at some code around here for unrelated reasons :-)

### DIFF
--- a/agent/service.go
+++ b/agent/service.go
@@ -113,11 +113,17 @@ func unmarshalExtraData(marshalled *types.Any) (*proto.ExtraData, error) {
 	return extraData, nil
 }
 
+func fifoName(taskID, execID string) string {
+	return fmt.Sprintf("exec-%s-task-%s", execID, taskID)
+}
+
 // Create creates a new initial process and container using runc
 func (ts *TaskService) Create(requestCtx context.Context, req *taskAPI.CreateTaskRequest) (*taskAPI.CreateTaskResponse, error) {
 	defer logPanicAndDie(log.G(requestCtx))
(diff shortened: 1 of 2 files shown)


agent/service.go
```diff
@@ -113,11 +113,17 @@ func unmarshalExtraData(marshalled *types.Any) (*proto.ExtraData, error) {
...
+}
+
 // Create creates a new initial process and container using runc
 func (ts *TaskService) Create(requestCtx context.Context, req *taskAPI.CreateTaskRequest) (*taskAPI.CreateTaskResponse, error) {
 	defer logPanicAndDie(log.G(requestCtx))
+	taskID := req.ID
+	execID := "" // the exec ID of the initial process in a task is an empty string by containerd convention
 
-	logger := log.G(requestCtx).WithFields(logrus.Fields{"id": req.ID, "bundle": req.Bundle})
+	logger := log.G(requestCtx).WithFields(logrus.Fields{"id": taskID, "bundle": req.Bundle})
```
mxpv: As we operate both task/exec ID, would it make sense to clarify the "id" field here?

sipsma: Ah sure, that's what I was doing in Exec anyways. Updated it to just be the same logger we use in Exec. I also don't include the bundle dir in the log message anymore because it's not actually providing any useful information (it was the agent logging the bundle dir used in by the runtime outside the VM).

PR #224, opened 2019-07-16
Upgrade CI to build with go 1.12.

sipsma:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ dist: xenial
 language: go
 
 go:
+  - 1.12.x
   - 1.11.x
 
 env:
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -17,7 +17,7 @@
 # COMMON IMAGES
 #
 #########################################
-FROM debian:stretch as base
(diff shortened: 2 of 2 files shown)


mxpv:
I think it make sense to keep things consistent and update to 1.12 everywhere, including:
https://github.com/firecracker-microvm/firecracker-containerd/blob/master/.travis.yml#L5
https://github.com/firecracker-microvm/firecracker-containerd/blob/master/docs/getting-started.md#clone-this-repository-and-build-the-components

sipsma:
> I think it make sense to keep things consistent and update to 1.12 everywhere, including:
> https://github.com/firecracker-microvm/firecracker-containerd/blob/master/.travis.yml#L5
> https://github.com/firecracker-microvm/firecracker-containerd/blob/master/docs/getting-started.md#clone-this-repository-and-build-the-components

Good catch, though I actually left the getting-started guide as is for now because it says "Go 1.11 or later", which is still true. Along the same vein I updated Travis to run with both 1.11 and 1.12 (which is apparently possible) so we can make sure it builds with both even if our CI is always using the latest go version.

Issue #65, opened 2019-01-10
Remove directories for removed containers on the host


samuelkarp:
```
[ec2-user@ip-172-31-30-18 ~]$ sudo /usr/local/bin/ctr run --snapshotter firecracker-naive --runtime aws.firecracker --tty docker.io/library/busybox:latest test
/ # echo hello
hello
/ # exit
[ec2-user@ip-172-31-30-18 ~]$ sudo /usr/local/bin/ctr c rm test
[ec2-user@ip-172-31-30-18 ~]$ sudo /usr/local/bin/ctr run --snapshotter firecracker-naive --runtime aws.firecracker --tty docker.io/library/busybox:latest test
ctr: mkdir /run/containerd/io.containerd.runtime.v2.task/default/test: file exists: unknown
```

Even though I've deleted the container (and its snapshot) on the host, ~the microVM rootfs~ the host retains the bundle directory that was created for the container.  We should ~either~ clean up the bundle directory when the container is removed (or stopped?)~, or we should ensure that the microVM rootfs is read-only (or separate per microVM)~.

samuelkarp:
Per discussion with @IRCody yesterday, he thinks that the error relates to a path on the host instead of inside the microVM.  When I look on the host, I do see what he's talking about:

```
[ec2-user@ip-172-31-30-18 ~]$ sudo ls /run/containerd/io.containerd.runtime.v2.task/default/test
address  config.json  firecracker.sock	log  rootfs  shim.pid  work
```

I'm not sure what is supposed to clean up this directory, maybe the runtime?

IRCody:
I think the runtime is supposed to clean it up.  I think [this comment](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/runtime/service.go#L470) is wrong and we actually do need to clean some stuff up here.

Reading the [docs](https://github.com/containerd/containerd/blob/master/runtime/v2/README.md#delete) it looks like there are two scenarios:
* containerd will call cleanup() over RPC.
* containerd is expecting to call the runtime binary with the "delete" subcommand when RPC calls fail expects that runtime binary to cleanup any state left by the other one. That delete call gets processed [here](https://github.com/containerd/containerd/blob/master/runtime/v2/shim/shim.go) which calls cleanup.

samuelkarp:
There appear to be additional leftover files in `/run/containerd/fifo`.

samuelkarp:
[...]

PR #225, opened 2019-07-17
Make sure containers are deleted correctly

kzys:
*Issue #, if available:*

#65

*Description of changes:*

While we cannot repro #65 anymore, we'd like to have some tests to
prevent regression.

### DIFF
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -124,7 +124,7 @@ type service struct {
 	agentClient              taskAPI.TaskService
 	eventBridgeClient        eventbridge.Getter
 	stubDriveHandler         stubDriveHandler
-	exitAfterAllTasksDeleted bool
+	exitAfterAllTasksDeleted bool // exit the VM and shim when all tasks are deleted
 
 	machine          *firecracker.Machine
 	machineConfig    *firecracker.Config
@@ -284,7 +284,8 @@ func (s *service) StartShim(shimCtx context.Context, containerID, containerdBina
 
 	logrus.SetOutput(logFifo)
 
-	log.G(shimCtx).WithField("id", containerID).Debug("StartShim")
+	log := log.G(shimCtx).WithField("id", containerID)
+	log.Debug("StartShim")
 
 	// If we are running a shim start routine, we can safely assume our current working
 	// directory is the bundle directory
@@ -319,6 +320,10 @@ func (s *service) StartShim(shimCtx context.Context, containerID, containerdBina
 		// task is deleted
 		containerCount = 1
 		exitAfterAllTasksDeleted = true
+
+		log.Infof("will start a single-task VM %s since no VMID has been provided", s.vmID)
+	} else {
+		log.Infof("will start a persistent VM %s", s.vmID)
 	}
 
 	client, err := ttrpcutil.NewClient(containerdAddress + ".ttrpc")
@@ -773,11 +778,34 @@ func (s *service) Start(requestCtx context.Context, req *taskAPI.StartRequest) (
 
 func (s *service) Delete(requestCtx context.Context, req *taskAPI.DeleteRequest) (*taskAPI.DeleteResponse, error) {
 	defer logPanicAndDie(log.G(requestCtx))
-	log.G(requestCtx).WithFields(logrus.Fields{"id": req.ID, "exec_id": req.ExecID}).Debug("delete")
+	logger := log.G(requestCtx).WithFields(logrus.Fields{"id": req.ID, "exec_id": req.ExecID})
+
+	logger.Debug("delete")
 
 	resp, err := s.taskManager.DeleteProcess(requestCtx, req, s.agentClient)
 	if err != nil {
 		return nil, err
+	}
+
+	// Only delete a process as like runc when there is ExecID
+	// https://github.com/containerd/containerd/blob/f3e148b1ccf268450c87427b5dbb6187db3d22f1/runtime/v2/runc/container.go#L320
+	if req.ExecID != "" {
+		return resp, nil
+	}
+
+	// Otherwise, delete the container
+	dir, err := s.shimDir.BundleLink(req.ID)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to find the bundle directory of the container: %s", req.ID)
+	}
+
+	_, err = os.Stat(dir.RootPath())
+	if os.IsNotExist(err) {
+		return nil, errors.Wrapf(err, "failed to find the bundle directory of the container: %s", dir.RootPath())
+	}
+
+	if err = os.Remove(dir.RootPath()); err != nil {
+		return nil, errors.Wrapf(err, "failed to remove the bundle directory of the container: %s", dir.RootPath())
 	}
 
 	return resp, nil
--- a/runtime/service_integ_test.go
+++ b/runtime/service_integ_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"strconv"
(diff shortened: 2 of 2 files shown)


runtime/service_integ_test.go
```diff
@@ -533,3 +533,98 @@ vdf  254:80   0  512B  0 disk`
...
+	client, err := containerd.New(containerdSockPath, containerd.WithDefaultRuntime(firecrackerRuntime))
+	require.NoError(t, err, "unable to create client to containerd service at %s, is containerd running?", containerdSockPath)
+	defer client.Close()
+
+	image, err := client.Pull(ctx, debianDockerImage, containerd.WithPullUnpack, containerd.WithPullSnapshotter(naiveSnapshotterName))
+	require.NoError(t, err, "failed to pull image %s, is the the %s snapshotter running?", debianDockerImage, naiveSnapshotterName)
+
+	// TODO once Noah's immutable rootfs change is merged, we can use that as our rootfs for all VMs instead of copying
+	// one per-VM
+	rootfsBytes, err := ioutil.ReadFile(defaultVMRootfsPath)
```
sipsma: nit: rewriting the rootfs file is only required if it's going to be used by multiple VMs concurrently, so I think it can be skipped for this test.

runtime/service_integ_test.go
```diff
@@ -533,3 +533,98 @@ vdf  254:80   0  512B  0 disk`
...
+	require.NoError(t, err, "failed to read rootfs file")
+	rootfsPath := fmt.Sprintf("%s.%d", defaultVMRootfsPath, vmID)
+	err = ioutil.WriteFile(rootfsPath, rootfsBytes, 0600)
+	require.NoError(t, err, "failed to copy vm rootfs to %s", rootfsPath)
+
+	tapName := fmt.Sprintf("tap%d", vmID)
```
sipsma: It doesn't look like the tap device is being provided to the VM (which is fine since it doesn't need network connectivity for this test), so we can skip creating it.

runtime/service_integ_test.go
```diff
@@ -533,3 +533,98 @@ vdf  254:80   0  512B  0 disk`
...
+		containerd.WithSnapshotter(naiveSnapshotterName),
+		containerd.WithNewSnapshot(snapshotName, image),
+		containerd.WithNewSpec(
+			oci.WithProcessArgs("echo", "hello"),
+			firecrackeroci.WithVMID(strconv.Itoa(vmID)),
+		),
+	)
+	require.NoError(t, err, "failed to create container %s", containerName)
+	require.Equal(t, "hello", startAndWaitTask(ctx, t, c1))
+	c1.Delete(ctx)
```
sipsma: Does this `Delete` call return an error? If so, we should check it. Same for `c2.Delete`

runtime/service_integ_test.go
```diff
@@ -533,3 +533,98 @@ vdf  254:80   0  512B  0 disk`
...
+		require.Equal(t, "", stderr.String())
+	case <-ctx.Done():
+		require.Fail(t, "context cancelled",
+			"context cancelled while waiting for container %s to exit, err: %v", c.ID(), ctx.Err())
+	}
+
+	return stdout.String()
+}
+
+func TestRemoveDir_Isolated(t *testing.T) {
```
sipsma: It looks like this test is verifying the problem mentioned in #65 about IO not working when creating a new container no longer exists, which is good. However, it doesn't appear to currently be testing that all files we create related to the container are removed, e.g. the symlinks here that are parameterized by `containerID`: https://github.com/firecracker-microvm/firecracker-containerd/blob/e84a435e52acb36157536b5f3ddba090ed0da789/internal/vm/dir.go#L100-L132

I think it's worth adding some assertions to this test that those symlinks are deleted too.

runtime/service_integ_test.go
kzys: Yes it does. Will fix

runtime/service_integ_test.go
kzys: Yes. It is currently only testing the equivalent of `ctr run ... test`, `ctr c rm ... test` and `ctr run ... test`. Let me check the symlinks.

runtime/service_integ_test.go
kzys: Will do.

runtime/service_integ_test.go
kzys: I saw that as more like a workaround for the lack of immutable rootfs. We are copying from defaultVMRootfsPath to rootfsPath, to make sure that the rootfs won't be compromised.

Do we already have the "Noah's immutable rootfs change"? If so, we can just use defaultVMRootfsPath and probably make the rootfs read-only.

sipsma: Your concern about making sure the rootfs isn't compromised is totally valid in general, however due to the particulars of this test we don't need to worry about it right now. These integ tests are all currently expected to be running inside isolated docker containers, so in the case that the rootfs was actually modified, the modification would be isolated to the container (thanks to overlayfs).

Noah's immutable rootfs is indeed merged, however our CI system has not yet been updated to use it. That's something we absolutely need to do, however IMO it's also a separate effort from what you're currently doing in this PR, so I think it should be addressed separately.

kzys:
I initially thought that the issue itself has been resolved, but it seems not.

https://buildkite.com/firecracker-microvm/firecracker-containerd/builds/621#dca1dcc9-eb49-456e-9d4e-7d25a09fa268

> failed to create VM dir bundle link: failed to create bundle symlink from "/var/run/firecracker-containerd/default/0/TestRemoveDir_Isolated-1563413455185085842"->"/run/containerd/io.containerd.runtime.v2.task/default/TestRemoveDir_Isolated-1563413455185085842": symlink /run/containerd/io.containerd.runtime.v2.task/default/TestRemoveDir_Isolated-1563413455185085842 /var/run/firecracker-containerd/default/0/TestRemoveDir_Isolated-1563413455185085842: file exists

kzys:
Now I think our code was fine but the tests were wrong. Let's see. I'm going to work on the comments after CIs are being green.

runtime/service_integ_test.go
kzys: Fixed

runtime/service_integ_test.go
kzys: Addressed

runtime/service_integ_test.go
kzys: Addressed

runtime/service_integ_test.go
kzys: The latest revision calls `os.Stat` to make sure that the associated directory of a container is being removed from the host OS correctly. Hard-coding the path is a bit annoying though.

runtime/service_integ_test.go
```diff
@@ -533,3 +534,80 @@ vdf  254:80   0  512B  0 disk`
...
+	require.NoError(t, err, "unable to create client to containerd service at %s, is containerd running?", containerdSockPath)
+	defer client.Close()
+
+	image, err := client.Pull(ctx, debianDockerImage, containerd.WithPullUnpack, containerd.WithPullSnapshotter(naiveSnapshotterName))
+	require.NoError(t, err, "failed to pull image %s, is the the %s snapshotter running?", debianDockerImage, naiveSnapshotterName)
+
+	containerName := fmt.Sprintf("%s-%d", t.Name(), time.Now().UnixNano())
+	snapshotName := fmt.Sprintf("%s-snapshot", containerName)
+
+	containerPath := fmt.Sprintf("/run/containerd/io.containerd.runtime.v2.task/default/%s", containerName)
```
sipsma: It's good to check this too, but what about the symlinks I mentioned last revision? It actually looks like we aren't deleting those at all right now AFAIK: https://github.com/firecracker-microvm/firecracker-containerd/blob/master/runtime/service.go#L774

I can see why that works out for the test case as is where you re-use a container name, but I think we should address removing those symlinks and have this test verify they are gone too. Otherwise, the shim's directory can fill up with symlinks for containers that have already been deleted. Let me know if I'm misunderstanding or missing something though.

runtime/service_integ_test.go
```diff
@@ -533,3 +534,80 @@ vdf  254:80   0  512B  0 disk`
...
+
+	task, err := c.NewTask(ctx, cio.NewCreator(cio.WithStreams(nil, &stdout, &stderr)))
+	require.NoError(t, err, "failed to create task for container %s", c.ID())
+
+	exitCh, err := task.Wait(ctx)
+	require.NoError(t, err, "failed to wait on task for container %s", c.ID())
+
+	err = task.Start(ctx)
+	require.NoError(t, err, "failed to start task for container %s", c.ID())
+	defer task.Delete(ctx)
```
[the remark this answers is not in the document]
kzys: Will do. Is there a way to check unchecked errors? I may be asking about checked exceptions though...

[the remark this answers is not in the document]
sipsma: I was just imagining something like:
```
defer func() {
    err := task.Delete(ctx)
    require.NoError(t, err, "...")
}()
```

Let me know if I'm misunderstanding the question though

runtime/service_integ_test.go
kzys: The shim itself is removing things from /var/run/firecracker-containerd/$namespace/$vm_id at `Shutdown()`.

https://github.com/firecracker-microvm/firecracker-containerd/blob/d9ebc8b24cac928fef30f302dae44f07489a3dbf/runtime/service.go#L999

When a VM only hosts a single task, what we have seems sufficient. When a VM hosts multiple tasks (in other words, the VM is persistent), we need to remove things at the end of the task, which we are not doing right now.

Probably I should have a different test for the persistent VM case.

sipsma: Ahhh right, that all adds up then. Yeah I agree having a test case for each codepath would be good

runtime/service_integ_test.go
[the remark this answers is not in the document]
kzys: Addressed at https://github.com/firecracker-microvm/firecracker-containerd/pull/225/commits/cdde8882e69ed92ce3148d5e7bccfa7effa7876c

kzys:
All tests are passing finally!

runtime/service_integ_test.go
kzys: Added /var/run check on https://github.com/firecracker-microvm/firecracker-containerd/pull/225/commits/31baf7a8fef63603023185d514d5da014420ce6c

runtime/service.go
```diff
@@ -773,13 +778,28 @@ func (s *service) Start(requestCtx context.Context, req *taskAPI.StartRequest) (
...
 
+	dir, err := s.shimDir.BundleLink(req.ID)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to find the bundle directory of the container: %s", req.ID)
+	}
+
+	_, err = os.Stat(dir.RootPath())
+	if os.IsNotExist(err) {
+		// is this really bad?
+		logger.Warnf("failed to delete %s: %v", dir.RootPath(), err)
```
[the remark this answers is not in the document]
kzys: Because Delete() can be called multiple times? I'm actually not so sure why TestMultipleVMs_Isolated triggers the warning.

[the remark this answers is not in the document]
mxpv: Even if it's called multiple times and we're attempting to delete a dir that is already deleted, `RemoveAll` just returns nil in this case. That should be fine for us.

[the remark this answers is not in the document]
kzys: We are deleting a process at https://github.com/firecracker-microvm/firecracker-containerd/blob/d2a7b17a0f8b89c90cff28b509fb051452a982e5/runtime/service_integ_test.go#L335 and then deleting its container at https://github.com/firecracker-microvm/firecracker-containerd/blob/d2a7b17a0f8b89c90cff28b509fb051452a982e5/runtime/service_integ_test.go#L380.

Is it safe to delete the container-wide files (e.g. /var/run/firecracker-containerd/$namespace/$vm_id/$container_id directory) at L335?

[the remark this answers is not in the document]
kzys: It seems ExecID = "" has a special meaning runc

https://github.com/containerd/containerd/blob/f3e148b1ccf268450c87427b5dbb6187db3d22f1/runtime/v2/runc/container.go#L328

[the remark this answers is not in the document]
kzys: Seems ExecID = "" means the container itself, according to https://github.com/containerd/containerd/blob/f3e148b1ccf268450c87427b5dbb6187db3d22f1/runtime/v2/runc/container.go#L320 and https://github.com/containerd/containerd/blob/f3e148b1ccf268450c87427b5dbb6187db3d22f1/runtime/v2/runc/container.go#L245

We probably should follow the convention on our Delete() implementation.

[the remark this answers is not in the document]
sipsma: Yeah, I agree we should only delete this bundle link when ExecID is "", as that is when the init process of the task is being deleted. We still need the bundle link if we are only deleting a non-init process of the task.

runtime/service_integ_test.go
```diff
@@ -533,3 +534,124 @@ vdf  254:80   0  512B  0 disk`
...
+}
+
+func testCreateContainerWithSameName(t *testing.T, vmID string) {
+	ctx := namespaces.WithNamespace(context.Background(), "default")
+
+	pluginClient, err := ttrpcutil.NewClient(containerdSockPath + ".ttrpc")
+	require.NoError(t, err, "failed to create ttrpc client")
+
+	// Explicitly specify Container Count = 2
+	// TODO: shouldn't it work without specifying the count?
```
sipsma: I think we should open an issue for this (being able to reuse a drive once a container is deleted) and then link to that issue here

kzys: Will link from the TOD comment #230

kzys: Done https://github.com/firecracker-microvm/firecracker-containerd/pull/225/commits/c68458acf62191dffee91eba01bc2ba2ccccbaad

runtime/service.go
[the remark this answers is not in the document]
kzys: Checking ExecID != "" on https://github.com/firecracker-microvm/firecracker-containerd/pull/225/commits/c68458acf62191dffee91eba01bc2ba2ccccbaad

runtime/service.go
```diff
@@ -773,13 +778,34 @@ func (s *service) Start(requestCtx context.Context, req *taskAPI.StartRequest) (
...
+	logger := log.G(requestCtx).WithFields(logrus.Fields{"id": req.ID, "exec_id": req.ExecID})
+
+	logger.Debug("delete")
 
 	resp, err := s.taskManager.DeleteProcess(requestCtx, req, s.agentClient)
 	if err != nil {
 		return nil, err
 	}
 
+	if req.ExecID != "" {
```
sipsma: nit: I'd add a comment explaining why we skip deletion in this case

PR #226, opened 2019-07-17
Sync quickstart and getting-started docs

sipsma:
The commits here:
* Build a `firecracker-ctr` binary from the same version of our `firecracker-containerd` binary to ensure the two are at compatible versions. The docs are updated to specify use of `firecracker-ctr` over plain `ctr`.
* Specify in the docs that Firecracker needs to be built from the 0.17.0 tag, which we recently updated to use.

As discussed elsewhere, I originally attempted to build the `firecracker-ctr` binary by adding `containerd` as a submodule and put a replace directive in our go.mod file pointing to the submodule. This actually worked but was extremely ugly as use of the replace directive requires that the replacement directory have a go.mod file, which in turn requires us to run `go mod init` in our containerd submodule, which itself had other side effects, etc. The solution instead is to just use `go install` to install the `ctr` binary using the version in our `go.mod` file and rename it to `firecracker-ctr`.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -66,7 +66,7 @@ optional `vsock` feature using the `--features vsock` flag.
 > seccomp, you must adjust your seccomp profile for these changes.
 
 ```bash
-git checkout v0.16.0 # latest released tag
+git checkout v0.17.0 # latest released tag
 cargo build --release --features vsock # --target x86_64-unknown-linux-gnu
 ```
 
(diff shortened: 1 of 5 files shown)


kzys:
Regarding the new ctr binary, do you want to update README.md files as well?

https://github.com/firecracker-microvm/firecracker-containerd/search?q=ctr&unscoped_q=ctr

sipsma:
> Regarding the new ctr binary, do you want to update README.md files as well?
> 
> https://github.com/firecracker-microvm/firecracker-containerd/search?q=ctr&unscoped_q=ctr

Ah good catch, I missed the mention of `ctr` in those other READMEs.

For the Runtime README, I think it may make the most sense to just have it point to our getting started guide as the current command doesn't really provide much context on how to use it anyways. I'll update with this.

For the snapshotter READMEs, the snapshotters can actually be used outside the context of firecracker-containerd, so I think it would make more sense to leave them as is. We may even remove them from this repository in the near-ish future anyways as Max merged them into containerd upstream.

PR #227, opened 2019-07-18
Run all tests as part of CI.

sipsma:
Before this change, there were a few set of tests that were accidentally not
being run during CI (including a unit test that failed, which was fixed here).

This change fixes that issue and also refactors the Makefile setup for running
tests to be more consistent and easy to understand. There's now top-level
targets in the main Makefile for "test" (unit tests) and "integ-test", which
recurse into each subdir Makefile to run the actual tests.

The top level "test" target is now used in CI, which should make it harder
for unit tests to be accidentally missed in the future. Unfortunately, right now
the top level "integ-test" target can't be used in CI because of the requirement
to limit integ test concurrency to 1. However, the top-level "integ-test"
target is still added here for developer convenience and in case it can be used
by CI in the future.


sipsma:
> Looks good to me! Can we close #114 now?

Thanks! And great point, yes we can

PR #232, opened 2019-07-24
Set log fields consistently

kzys:
Use "id", "exec_id" and "vmID" consistently.

*Issue #, if available:*

None

*Description of changes:*

Some methods used "containerID", "TaskID" and/or "ExecID"

### DIFF
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -315,6 +315,7 @@ func (s *service) StartShim(shimCtx context.Context, containerID, containerdBina
 		}
 
 		s.vmID = uuid.String()
+		s.logger = s.logger.WithField("vmID", s.vmID)
 
 		// If the client didn't specify a VMID, this is a single-task VM and should thus exit after this
 		// task is deleted
@@ -661,7 +662,7 @@ func (s *service) buildVMConfiguration(req *proto.CreateVMRequest) (*firecracker
 }
 
 func (s *service) Create(requestCtx context.Context, request *taskAPI.CreateTaskRequest) (*taskAPI.CreateTaskResponse, error) {
-	logger := s.logger.WithField("containerID", request.ID)
+	logger := s.logger.WithField("id", request.ID)
 	defer logPanicAndDie(logger)
 
 	err := s.waitVMReady()
@@ -814,7 +815,7 @@ func (s *service) Delete(requestCtx context.Context, req *taskAPI.DeleteRequest)
 // Exec an additional process inside the container
 func (s *service) Exec(requestCtx context.Context, req *taskAPI.ExecProcessRequest) (*ptypes.Empty, error) {
 	defer logPanicAndDie(log.G(requestCtx))
-	logger := s.logger.WithField("TaskID", req.ID).WithField("ExecID", req.ExecID)
+	logger := s.logger.WithField("id", req.ID).WithField("exec_id", req.ExecID)
 	logger.Debug("exec")
 
 	// no OCI config bytes or DriveID to provide for Exec, just leave those fields empty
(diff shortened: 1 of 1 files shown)


runtime/service.go
```diff
@@ -315,6 +315,7 @@ func (s *service) StartShim(shimCtx context.Context, containerID, containerdBina
 		}
 
 		s.vmID = uuid.String()
+		s.logger = s.logger.WithField("vmID", s.vmID)
```
kzys: BTW, we are mixing lowerCamelCase and snake_case.

runtime/service.go
```diff
@@ -814,7 +815,7 @@ func (s *service) Delete(requestCtx context.Context, req *taskAPI.DeleteRequest)
 // Exec an additional process inside the container
 func (s *service) Exec(requestCtx context.Context, req *taskAPI.ExecProcessRequest) (*ptypes.Empty, error) {
 	defer logPanicAndDie(log.G(requestCtx))
-	logger := s.logger.WithField("TaskID", req.ID).WithField("ExecID", req.ExecID)
+	logger := s.logger.WithField("id", req.ID).WithField("exec_id", req.ExecID)
```
sipsma: I'm good with this update in general as I agree the logs have grown inconsistent as the code evolved over time. But I would prefer that they consistently use `TaskID` and `ExecID` (or `task_id` and `exec_id`, don't have a preference on the casing style) as I think that's the most explicit names for the fields.

runtime/service.go
sipsma: Since we are adding this, we now longer need to include `s.vmID` in the logs written on lines 325 and 327 below.

runtime/service.go
sipsma: Also, it may be worth just defining these field name strings as constants somewhere so their definition can just be centralized. I don't think that's absolutely necessary though, just something to consider.

kzys: Thanks. Will use `task_id` instead of `id` then.

Regarding the constants, I would like to have helper methods. Maybe logForTask and logForExec?

sipsma: Yeah helper methods are a good idea too

kzys: Actually we are sometimes takes a logger from the context passed-in (`log.G(context)`) and sometimes takes a logger from the service instance. I'm going to rename "id" to "task_id" in this pull request, and will add helper methods to create a logger consistently.

runtime/service.go
```diff
@@ -315,15 +315,16 @@ func (s *service) StartShim(shimCtx context.Context, containerID, containerdBina
...
 		s.vmID = uuid.String()
+		s.logger = s.logger.WithField("vmID", s.vmID)
 
 		// If the client didn't specify a VMID, this is a single-task VM and should thus exit after this
 		// task is deleted
 		containerCount = 1
 		exitAfterAllTasksDeleted = true
 
-		log.Infof("will start a single-task VM %s since no VMID has been provided", s.vmID)
+		log.Info("will start a single-task VM since no VMID has been provided")
```
sipsma: I think this and below should be `s.logger`, right?

sipsma: Ah wait, after re-reading this I believe this line should actually stay as `log.Info` but you need to change line 318 to from `s.logger = s.logger.WithField(...)` to `log = log.WithField(...)`. 

`s.logger` is the shim's overall logger. However, this method `StartShim` actually only runs in a short lived process invoked by containerd to create the real, long-running shim process, so setting `s.logger` here won't have any effect (confusing, I know, [the containerd docs have more on this](https://github.com/containerd/containerd/blob/master/runtime/v2/README.md#start))

PR #233, opened 2019-07-25
Make integration tests distro-agnostic

### LINKED ISSUE #203: runtime/service_integ_test: Shouldn't use lsblk to validate
We currently use `lsblk` to validate that the stub drives have been created. However, we are `lsblk` is tied to the `Debian` image and the output, which we use for validation, is prone to change. A better and longer term solution would be write a piece of software that does something similarly to `lsblk` but has a strict output. This executable should be injected into the root image.

Something like:
```
Name        Size          StubDrive
vda         1M            N
vdb         512B          Y
```

kzys:
*Issue #, if available:*

#203

*Description of changes:*

This change introduces our own lsblk(8) equivalent, which makes
the integration test distro-agonostic and be able to run without
util-linux.

Fixes #203.

### DIFF
--- a/runtime/service_integ_test.go
+++ b/runtime/service_integ_test.go
@@ -434,7 +434,6 @@ func TestMultipleVMs_Isolated(t *testing.T) {
 
 	vmWg.Wait()
 }
-
 func TestStubBlockDevices_Isolated(t *testing.T) {
 	internal.RequiresIsolation(t)
 	const vmID = 0
@@ -490,8 +489,29 @@ func TestStubBlockDevices_Isolated(t *testing.T) {
 		containerd.WithSnapshotter(naiveSnapshotterName),
 		containerd.WithNewSnapshot(snapshotName, image),
 		containerd.WithNewSpec(
-			oci.WithProcessArgs("lsblk"),
 			firecrackeroci.WithVMID(strconv.Itoa(vmID)),
+			oci.WithProcessArgs("/bin/sh", "/var/firecracker-containerd-test/scripts/lsblk.sh"),
+
+			oci.WithMounts([]specs.Mount{
+				// Exposes the host kernel's /dev as /dev.
+				// By default, runc creates own /dev with a minimal set of pseudo devices such as /dev/null.
+				{
+					Type:        "bind",
+					Options:     []string{"bind"},
+					Destination: "/dev",
+					Source:      "/dev",
+				},
+
+				// Exposes test scripts from the host kernel
+				{
+					Type:        "bind",
+					Options:     []string{"bind"},
+					Destination: "/var/firecracker-containerd-test/scripts",
+					Source:      "/var/firecracker-containerd-test/scripts",
+				},
+			}),
+			// Make the host kernel's /dev readable
+			oci.WithParentCgroupDevices,
 		),
 	)
 	require.NoError(t, err, "failed to create container %s", containerName)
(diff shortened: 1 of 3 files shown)


kzys:
I'm still working on /dev issue but I'd like to make sure that the overall direction is fine.

runtime/Makefile
```diff
@@ -50,6 +50,7 @@ integ-test:
 			--volume /sys:/sys \
 			--volume /run/udev/control:/run/udev/control \
 			--volume $(CURDIR)/logs:/var/log/firecracker-containerd-test \
+			--volume $(CURDIR)/scripts:/tmp/scripts \
```
sipsma: I think it may be more consistent to `COPY` this into the Dockerfile we use for the test images, around here: https://github.com/firecracker-microvm/firecracker-containerd/blob/master/tools/docker/Dockerfile#L180

runtime/service_integ_test.go
```diff
@@ -485,12 +484,16 @@ func TestStubBlockDevices_Isolated(t *testing.T) {
...
+	script, err := ioutil.ReadFile("/tmp/scripts/lsblk.sh")
+	require.NoError(t, err)
+
 	newContainer, err := client.NewContainer(ctx,
 		containerName,
 		containerd.WithSnapshotter(naiveSnapshotterName),
 		containerd.WithNewSnapshot(snapshotName, image),
 		containerd.WithNewSpec(
-			oci.WithProcessArgs("lsblk"),
+			oci.WithEnv([]string{"SCRIPT=" + string(script)}),
```
sipsma: Do you mind trying out quickly adding `oci.WithParentCgroupDevices` and `oci.WithCgroup("")` here and seeing if that allows you to actually read the block devices? https://github.com/containerd/containerd/blob/41e172352c41b6ff9e3503241d1bb17bfbcc28a0/oci/spec_opts.go#L988

`WithParentCgroupDevices` is an option not available in the normal `ctr` command line and on the surface seems like it in combination with `WithCgroup` might do what we want. If it doesn't work, no worries, but I think it's worth a shot as the test is significantly improved if we can verify each stub drive is being written with the right contents

runtime/Makefile
kzys: Thanks. Will do.

kzys: Addressed https://github.com/firecracker-microvm/firecracker-containerd/pull/233/commits/c9f3074909d63e05909092f6f69923b89d2b3ce5

runtime/service_integ_test.go
kzys: Thanks. WithParentCgroupDevices works! https://github.com/firecracker-microvm/firecracker-containerd/pull/233/commits/569fea02d0d241a3f85e4ba661e484b0a2d21e1f

By default, a containerd client passes `LinuxDeviceCgroup { Allow:false, Access: rwm }` (ref https://github.com/containerd/containerd/blob/7ff23438f64cfa62651ee9beea83f4dddfcf5a1e/oci/spec.go#L231). `WithParentCgroupDevices` clears this thing which makes /dev accessible.

tools/docker/scripts/lsblk.sh
```diff
@@ -0,0 +1,25 @@
+#! /bin/sh
+# tiny lsblk(8) equivalent to make integration tests distro-agnostic
+set -eu
+
+echo 'NAME MAJ:MIN RM      SIZE RO | MAGIC'
+
+for name in $(ls /sys/block | grep -v loop)
```
sipsma: nit: Is it possible to avoid grep? Grep is also pretty universal, so if there's no other straightforward way that's alright, but it's technically not a coreutil so it would be nice to avoid

tools/docker/Dockerfile
```diff
@@ -177,6 +177,7 @@ COPY --from=firecracker-vm-root-builder /output/vm.ext4 /var/lib/firecracker-con
 COPY --from=firecracker-containerd-build /output/* /usr/local/bin/
 COPY _submodules/runc/runc /usr/local/bin
 COPY tools/docker/firecracker-runtime.json /etc/containerd/firecracker-runtime.json
+COPY tools/docker/scripts/* /tmp/scripts/
```
sipsma: nit: I think I'd prefer something like `/var/firecracker-containerd-test/scripts` over `/tmp/scripts`. It would just be a little more resilient to future changes as `/tmp` is somewhat common to override with tmpfs mounts or bind mounts from the host.

If there's something I'm not thinking of that makes that difficult, no worries, it's a minor concern.

kzys: No strong preferences on that. Will use /var in that case.

kzys: Or /usr/local/bin?

sipsma: Sure, /usr/local/bin/ is good too

kzys: Second thought. I think /var/firecracker-containerd-test/scripts is better. Then we can bind-mount the directory as is, which allows us to remove the hack below.

https://github.com/firecracker-microvm/firecracker-containerd/pull/233/files#diff-5e5cc717e2803524fb4d62e8faea3689R496

sipsma: I might be misunderstanding, but we won't be able to bind mount from the host to inside the microvm. If you create a bind mount for the container it will be a bind mount from the microvm's rootfs to inside the container, not from the actual host outside the microvm to inside the container.

tools/docker/scripts/lsblk.sh
kzys: Hmm, microVMs on my Debian EC2 box doesn't have loopX but it has on Buildkite...

https://buildkite.com/firecracker-microvm/firecracker-containerd/builds/653#846f0b05-81a2-441f-82fd-0f95b124fb5f

kzys: Used case...esac instead.

tools/docker/Dockerfile
kzys: Right. I ended-up adding them on Firecracker's host image first, and bind-mount the directory on the guest.

kzys:
If the whole diff is fine, I'm going to squash the branch.

PR #235, opened 2019-07-26
[RFC] Use Alpine Linux instead of Debian GNU/Linux

kzys:
*Issue #, if available:*

None

*Description of changes:*

Alpine is smaller than Debian that would be faster to pull from the internet.

Pros:
- It might make our tests faster, in theory.

Cons:
- Assuming developers use Debian (I use, because of https://github.com/firecracker-microvm/firecracker-containerd/blob/master/docs/quickstart.md), the subtle difference between Debian and Alpine would be sometimes annoying.

### DIFF
--- a/runtime/service_integ_test.go
+++ b/runtime/service_integ_test.go
@@ -53,7 +53,7 @@ const (
 	defaultNamespace = namespaces.Default
 
 	containerdSockPath = "/run/containerd/containerd.sock"
-	guestDockerImage   = "docker.io/library/alpine:latest"
+	guestDockerImage   = "docker.io/library/alpine:3.10.1"
 
 	firecrackerRuntime   = "aws.firecracker"
 	naiveSnapshotterName = "firecracker-naive"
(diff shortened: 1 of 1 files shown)


kzys:
We can, but not so sure this is really worth to do.

runtime/service_integ_test.go
```diff
@@ -53,7 +53,7 @@ const (
 	defaultNamespace = namespaces.Default
 
 	containerdSockPath = "/run/containerd/containerd.sock"
-	debianDockerImage  = "docker.io/library/debian:latest"
+	guestDockerImage   = "docker.io/library/alpine:latest"
```
sipsma: While we're at it, let's pin this to a specific version instead of `latest` (i.e. `3.10.1)

kzys: Will do.

kzys:
No, it haven't notice any speedup on my EC2 box. BuildKite is also really not stable in terms of performance. https://buildkite.com/firecracker-microvm/firecracker-containerd

sipsma:
> No, it haven't notice any speedup on my EC2 box. BuildKite is also really not stable in terms of performance. https://buildkite.com/firecracker-microvm/firecracker-containerd

Okay cool, was just curious, it may make more of a difference in the future if we start running tests with hundreds of containers or something like that.

kzys:
Ahh, I wanted to remove "[RFC]" from the commit...

Issue #217, opened 2019-06-28
Change Shim OOM score to +1 of containerd


sipsma:
[Containerd recently updated their reference shim implementations to set the OOM score of the shim to +1 that of the parent containerd process](https://github.com/containerd/containerd/pull/3380):
> This should allow the shim's to
be killed first in an OOM condition but leave the daemon alone for a bit
to help cleanup and manage the containers during this situation.

We should probably make the same change [here](https://github.com/firecracker-microvm/firecracker-containerd/blob/caa5ff0e0bca9b0327b88c328d72cbbc78e36c43/firecracker-control/local.go#L315) but can't quite re-use their code because the fc-control plugin actually runs in the same process as containerd, so it needs to set the shim score to +1 of its own (not of the parent process).

praveensastry:
@sipsma I'd like to contribute to this.

sipsma:
@praveensastry Sounds great, thanks! Please feel free to reach out on this issue or email if we can help clarify anything in the process

sipsma:
Closed by #238

PR #238, opened 2019-07-29
Set Shim OOM score to +1 of containerd

praveensastry:
*Issue #, if available:* #217 

*Description of changes:* Added a method to get the OOM score for containerd process and increment the score by 1 to set shim OOM score

### DIFF
--- a/firecracker-control/local.go
+++ b/firecracker-control/local.go
@@ -326,7 +326,7 @@ func (s *local) newShim(ns, vmID, containerdAddress string, shimSocket *net.Unix
 		logger.WithError(waitErr).Debug("completed waiting on shim process")
 	}()
 
-	err = setShimOOMScore(cmd.Process.Pid)
+	err = setShimOOMScore()
 	if err != nil {
 		logger.WithError(err).Error()
 		return nil, err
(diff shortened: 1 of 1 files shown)


praveensastry:
@sipsma Would be great if you can take a look and provide feedback.

firecracker-control/local.go
```diff
@@ -338,3 +338,15 @@ func (s *local) newShim(ns, vmID, containerdAddress string, shimSocket *net.Unix
 func isEADDRINUSE(err error) bool {
 	return err != nil && strings.Contains(err.Error(), "address already in use")
 }
+
+func setShimOOMScore(pid int) error {
+	score, err := sys.GetOOMScoreAdj(pid)
```
sipsma: Here we want to get the OOM score of the containerd process, but I think it's currently getting the OOM score of the shim child process instead.

This firecracker-control code runs within the containerd process itself, so to get the containerd pid all we have to do is call [`os.Getpid()`](https://golang.org/pkg/os/#Getpid). Then we can plug that into `sys.GetOOMScoreAdj`.

praveensastry: Thanks for clarifying that! I thought to do the same when I was looking at containerd's implementation but got confused when I was implementing it.

sipsma: Ah okay, one more clarification. Your update is correct in that we are now using containerd's pid to get the OOM score. However, we actually want to set the OOM score on the pid you were using previously (`cmd.Process.Pid`). The end result we want is that `cmd.Process.Pid` gets the OOM score +1 of containerd's OOM score.

This is a little different than what containerd's code does, so I understand the confusion :-)

praveensastry:
Travis build failed due to an environmental issue (I believe) with this error message `The command "travis_setup_go" failed and exited with 86 during .`

samuelkarp:
@praveensastry I've rerun the Travis job for you

praveensastry:
Thank you @sipsma for the opportunity. Looking forward to contributing more :)

PR #240, opened 2019-08-05
Don't patch non-stub drives

### LINKED ISSUE #230: Cannot reuse a Firecracker VM from multiple containers without specifying Container Count
(From #225)

Probably due to the way we handle stub drives, re-using a Firecracker VM from multiple containers doesn't work due to a mysterious

> Drive "root_drive" could not be found

error.

kzys:
*Issue #, if available:*

Related to, but may need more changes to actually fix #230

*Description of changes:*

Currently the drive handler knows non-stub drives (root drive). Because of that PatchStubDrive() replace non-stub drives, once all stub drives are patched.

### DIFF
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -588,6 +588,25 @@ func (s *service) SetVMMetadata(requestCtx context.Context, request *proto.SetVM
 	return &empty.Empty{}, nil
 }
 
+func (s *service) createStubDrives(stubDriveCount int) ([]models.Drive, error) {
+	paths, err := s.stubDriveHandler.StubDrivePaths(stubDriveCount)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to retrieve stub drive paths")
+	}
+
+	stubDrives := make([]models.Drive, 0)
+	for i, path := range paths {
+		stubDrives = append(stubDrives, models.Drive{
+			DriveID:      firecracker.String(fmt.Sprintf("stub%d", i)),
+			IsReadOnly:   firecracker.Bool(false),
+			PathOnHost:   firecracker.String(path),
+			IsRootDevice: firecracker.Bool(false),
+		})
+	}
+
+	return stubDrives, nil
+}
+
 func (s *service) buildVMConfiguration(req *proto.CreateVMRequest) (*firecracker.Config, error) {
 	logger := s.logger.WithField("cid", s.machineCID)
 
@@ -618,15 +637,6 @@ func (s *service) buildVMConfiguration(req *proto.CreateVMRequest) (*firecracker
 	}
 
 	// Drives configuration
-
-	var driveBuilder firecracker.DrivesBuilder
-	if root := req.RootDrive; root != nil {
-		driveBuilder = firecracker.NewDrivesBuilder(root.PathOnHost)
-	} else {
-		driveBuilder = firecracker.NewDrivesBuilder(s.config.RootDrive)
-	}
-
-	stubDriveIndex := int64(len(driveBuilder.Build()) - 1)
 	containerCount := int(req.ContainerCount)
 	if containerCount < 1 {
 		// containerCount should always be positive so that at least one container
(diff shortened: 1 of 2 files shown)


kzys:
Honestly speaking, I'm unsure this is the right place to address the issue. We could argue that the drive handler shouldn't know about a root drive and the caller of this handler should not pass the root drive.

runtime/drive_handler.go
```diff
@@ -125,10 +125,20 @@ func (h *stubDriveHandler) createStubDrive(driveID, path string) error {
 	return nil
 }
 
+func filterOnlyStubDrives(drives []models.Drive) []models.Drive {
```
xibz: Let's move this function out of drive handler and into the `service.go`. You can then call the method before [set drives](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/runtime/service.go#L651).

xibz: May be a good idea to document why this occurs which is due to the root drive being always being the last element when `Build` is called on the drive builder from the SDK.

runtime/drive_handler.go
```diff
@@ -125,10 +125,20 @@ func (h *stubDriveHandler) createStubDrive(driveID, path string) error {
...
+		if !*d.IsRootDevice {
+			stubDrives = append(stubDrives, d)
+		}
+	}
+	return stubDrives
+}
+
 // SetDrives will set the given drives and the offset to which the stub drives
 // start.
 func (h *stubDriveHandler) SetDrives(offset int64, d []models.Drive) {
```
xibz: This offset can be removed all together with the function provided above.

runtime/drive_handler.go
```diff
@@ -125,10 +125,20 @@ func (h *stubDriveHandler) createStubDrive(driveID, path string) error {
 	return nil
 }
 
+func filterOnlyStubDrives(drives []models.Drive) []models.Drive {
+	stubDrives := make([]models.Drive, 0)
```
mxpv: nit: I'd rather prefer preallocating here as we know the size in advance: `make([]models.Drive, 0, len(drives))`

runtime/drive_handler.go
kzys: You meant, this `Build`.

https://github.com/firecracker-microvm/firecracker-go-sdk/blob/29621e288fa1333809e5b43d867dbb73d82b5456/drives.go#L63

BTW, why do we want to place the root drive at the end?

runtime/drive_handler.go
kzys: Wouldn't it allocate more than what we need?

runtime/drive_handler.go
xibz: The location of root drive is arbitrary. End was chosen since you didn't need to create a `[]models.Drive` to append to.

xibz:
I wonder if it makes sense to just have an array of stub drives that is passed into `SetDrives`. We only use a single drive builder, but it doesn't mean we can use two. That may simpler than trying to filter, and if we prefer filtering we may want to filter on what is a stub drive as opposed to what is root.

kzys:
It does make sense. `runtime/service.go` knows about stub drives, but Firecracker doesn't have the notion of stub drives. Thus fiding stub-drives on Firecracker-origin structs (such as cfg.Drives) is not straightforward.

runtime/drive_handler.go
```diff
@@ -125,10 +125,20 @@ func (h *stubDriveHandler) createStubDrive(driveID, path string) error {
...
+		}
+	}
+	return stubDrives
+}
+
 // SetDrives will set the given drives and the offset to which the stub drives
 // start.
 func (h *stubDriveHandler) SetDrives(offset int64, d []models.Drive) {
-	h.drives = d
+	h.drives = filterOnlyStubDrives(d)
```
sipsma: What's the error now when you encounter the issue in #230 ?

The fix here is good either way and if the error is still unclear that can be addressed separately, just curious if it's changed to something more comprehensible.

runtime/drive_handler.go
mxpv: Should it be just `make([]models.Drive)` then?

runtime/drive_handler.go
kzys: Good point. Let me try that.

runtime/drive_handler.go
kzys: While the latest version doesn't have filterOnlyStubDrives() anymore (as @xibz suggested), all changes are in service.go now.

kzys: Thanks. The new revision has some comments around drives initialization. Would it be good enough?

runtime/drive_handler.go
kzys: Now offset is always zero. I will remove the parameter later.

runtime/drive_handler.go
kzys: We don't have the filter logic anymore, and all `make` calls explicitly specify the capacity.

runtime/drive_handler.go
kzys: Now the error mentions "failed to patch stub drive: There are no remaining drives to be used" which is better than what I had before

```
--- FAIL: TestCreateContainerWithSameName_Isolated (33.06s)
    require.go:794:
                Error Trace:    service_integ_test.go:565
                                                        service_integ_test.go:655
                                                        service_integ_test.go:676
                Error:          Received unexpected error:
                                unknown
                                github.com/containerd/containerd/errdefs.init.ializers
                                        /go/pkg/mod/github.com/containerd/containerd@v1.2.1-0.20190711201753-f2b6c31d0fa7/errdefs/errors.go:43
                                runtime.main
                                        /usr/local/go/src/runtime/proc.go:188
                                runtime.goexit
                                        /usr/local/go/src/runtime/asm_amd64.s:1337
                                failed to patch stub drive: There are no remaining drives to be used
                                github.com/containerd/containerd/errdefs.FromGRPC
                                        /go/pkg/mod/github.com/containerd/containerd@v1.2.1-0.20190711201753-f2b6c31d0fa7/errdefs/grpc.go:107
                                github.com/containerd/containerd.(*container).NewTask
                                        /go/pkg/mod/github.com/containerd/containerd@v1.2.1-0.20190711201753-f2b6c31d0fa7/container.go:286
                                github.com/firecracker-microvm/firecracker-containerd/runtime.startAndWaitTask
                                        /firecracker-containerd/runtime/service_integ_test.go:564
                                github.com/firecracker-microvm/firecracker-containerd/runtime.testCreateContainerWithSameName
                                        /firecracker-containerd/runtime/service_integ_test.go:655
                                github.com/firecracker-microvm/firecracker-containerd/runtime.TestCreateContainerWithSameName_Isolated
                                        /firecracker-containerd/runtime/service_integ_test.go:676
                                testing.tRunner
                                        /usr/local/go/src/testing/testing.go:865
                                runtime.goexit
                                        /usr/local/go/src/runtime/asm_amd64.s:1337
                Test:           TestCreateContainerWithSameName_Isolated
                Messages:       failed to create task for container TestCreateContainerWithSameName_Isolated-1565126341850109350
FAIL
exit status 1
```

sipsma: Nice, yeah that's much clearer than before

PR #241, opened 2019-08-05
Add log folder cleanup

xibz:
When running isolated tests a logs folder is created but is not properly
cleaned up. This fix alleviates that by removing the folder


xibz:
Going to close this PR. Originally I thought this is something we may want to clean up but after some thoughts, it may be best to let it be cleaned up manually.

PR #244, opened 2019-08-07
Make sure the error message regarding stub drives is clear

### LINKED ISSUE #243: Support using a stub drive of a deleted container
As part of #230 @kzys noticed that if a container is deleted, the stub drive for it cannot be re-used. This results in situations like the following:

1. Call CreateVM with ContainerCount set to 2
2. Start container A
3. Start container B, let it exit and then Delete it.
3. Try to start container C, it will fail due to being out of stub drives

In theory, once container B is deleted, the stub drive is no longer needed and should be able to be used by container C. The implementation would need an update to allow this. To be clear, this is just re-use of stub-drives within the context of a single VM (not across different VMs, which creates a bunch of other concurrency related issues).

kzys:

*Issue #, if available:*

Related, but doesn't fix #243

*Description of changes:*

While we don't support the use-case right now, the error message
regarding the lack of free stub drives should be clear for customers.

This change also cleans up drive_handler.go.

### DIFF
--- a/runtime/drive_handler.go
+++ b/runtime/drive_handler.go
@@ -51,16 +51,41 @@ type stubDriveHandler struct {
 	mutex          sync.Mutex
 }
 
-func newStubDriveHandler(path string, logger *logrus.Entry) stubDriveHandler {
-	return stubDriveHandler{
+func newStubDriveHandler(path string, logger *logrus.Entry, count int) (*stubDriveHandler, error) {
+	h := stubDriveHandler{
 		RootPath: path,
 		logger:   logger,
 	}
+	drives, err := h.createStubDrives(count)
+	if err != nil {
+		return nil, err
+	}
+	h.drives = drives
+	return &h, nil
 }
 
-// StubDrivePaths will create stub drives and return the paths associated with
+func (h *stubDriveHandler) createStubDrives(stubDriveCount int) ([]models.Drive, error) {
+	paths, err := h.stubDrivePaths(stubDriveCount)
+	if err != nil {
+		return nil, err
+	}
+
+	stubDrives := make([]models.Drive, 0, stubDriveCount)
+	for i, path := range paths {
+		stubDrives = append(stubDrives, models.Drive{
+			DriveID:      firecracker.String(fmt.Sprintf("stub%d", i)),
+			IsReadOnly:   firecracker.Bool(false),
+			PathOnHost:   firecracker.String(path),
+			IsRootDevice: firecracker.Bool(false),
+		})
+	}
+
+	return stubDrives, nil
+}
+
+// stubDrivePaths will create stub drives and return the paths associated with
 // the stub drives.
-func (h *stubDriveHandler) StubDrivePaths(count int) ([]string, error) {
+func (h *stubDriveHandler) stubDrivePaths(count int) ([]string, error) {
 	paths := []string{}
 	for i := 0; i < count; i++ {
 		driveID := fmt.Sprintf("stub%d", i)
@@ -123,13 +148,6 @@ func (h *stubDriveHandler) createStubDrive(driveID, path string) error {
 	}
 
 	return nil
-}
-
-// SetDrives will set the given drives and the offset to which the stub drives
-// start.
-func (h *stubDriveHandler) SetDrives(d []models.Drive) {
-	h.drives = d
-	h.stubDriveIndex = 0
 }
 
 // GetDrives returns the associated stub drives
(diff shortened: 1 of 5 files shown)


runtime/service_integ_test.go
```diff
@@ -676,3 +677,54 @@ func TestCreateContainerWithSameName_Isolated(t *testing.T) {
...
+		err := c2.Delete(ctx, containerd.WithSnapshotCleanup)
+		require.NoError(t, err, "failed to delete a container")
+	}()
+
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+
+	// When we reuse a VM explicitly, we cannot start multiple containers unless we pre-allocate stub drives.
+	_, err = c2.NewTask(ctx, cio.NewCreator(cio.WithStreams(nil, &stdout, &stderr)))
+	assert.Equal("failed to patch stub drive: There are no remaining drives to be used: unknown", err.Error())
```
kzys: This `unknown` is a bit annoying. Should we return one of containerd's pre-defined errors such as ErrUnavailable?

https://github.com/containerd/containerd/blob/03d934adc4fb41763364b643e03399fdfcdcb20b/errdefs/errors.go#L48

runtime/drive_handler.go
```diff
@@ -127,9 +127,9 @@ func (h *stubDriveHandler) createStubDrive(driveID, path string) error {
 
 // SetDrives will set the given drives and the offset to which the stub drives
 // start.
-func (h *stubDriveHandler) SetDrives(offset int64, d []models.Drive) {
+func (h *stubDriveHandler) SetDrives(d []models.Drive) {
 	h.drives = d
-	h.stubDriveIndex = offset
+	h.stubDriveIndex = 0
```
sipsma: nit: I think we can just remove setting `h.stubDriveIndex` now since the struct will default the value to initially be 0 anyways.

runtime/service_integ_test.go
sipsma: Yeah I think returning `ErrUnavailable` is a great idea

runtime/drive_handler.go
kzys: `SetDrives()` can be called anytime during the handler's lifecycle. How about returning an error from this function, since we assume that SetDrives should be called before calling PatchStubDrive().

Another type-safe way of doing that is getting rid of SetDrives() and initializing the handler with stub drives. Making things immutable is good, while it may make this PR a bit bigger though.

sipsma: One thing to keep in mind is that this is a `main` package so I don't think any customers would be able to use the code. I get what you're saying though in that we'd be relying on our other code in this package to only call `SetDrives` once. That's technically enforced right now because you can only call CreateVM once (due to a `s.vmStartOnce` in `service.go`), but it does still all feel a bit fragile.

I'd be good with either approach you suggested. Initializing the stub drive handler w/ a container count only when the VM has started would be most in line with the rest of the code though (i.e. we currently initialize `s.agentClient`, `s.eventBridgeClient` and `s.exitAfterAllTasksDeleted` only when the VM has started and guard their initialization with `s.vmStartOnce`).

runtime/service_integ_test.go
```diff
@@ -676,3 +677,54 @@ func TestCreateContainerWithSameName_Isolated(t *testing.T) {
...
+		err := c2.Delete(ctx, containerd.WithSnapshotCleanup)
+		require.NoError(t, err, "failed to delete a container")
+	}()
+
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+
+	// When we reuse a VM explicitly, we cannot start multiple containers unless we pre-allocate stub drives.
+	_, err = c2.NewTask(ctx, cio.NewCreator(cio.WithStreams(nil, &stdout, &stderr)))
+	assert.Equal("no remaining stub drives to be used: unavailable: unknown", err.Error())
```
sipsma: Did changing to use `ErrUnavailable` not result in this error message being changed? I wonder if `ErrUnavailable` is getting lost somewhere in the call chain (as it goes client->ctrd->shimstart->fc-control->shim and back).

The updates in this PR LGTM anyways, the error message is still more clear and the code is more clear now too. I don't personally mind if dealing with `unknown` is a separate, not high-priority issue.

kzys: > no remaining stub drives to be used: **unavailable:** unknown

This part is coming from https://github.com/containerd/containerd/blob/ea13c9fe9941b0714630aa614ef5a0038c0083a3/errdefs/errors.go#L48, but maybe I should check the type rather than its string representation coming from `Error()` to make the intention more clear.

sipsma: Ah okay, totally missed the "unavailable"! Makes sense

PR #252, opened 2019-08-27
Upgrade containerd to 1.3.0-beta.2

kzys:
*Issue #, if available:* NA

*Description of changes:*

1.3.x is still beta, but

- 1.3.x is the first version that has Device Mapper snapshotter
- 1.2.x doesn't have
  github.com/containerd/containerd/runtime/v2/runc/v2

So, it wouldn't be worth to rewrite our agent to support 1.2.x at this
stage.


kzys:
Seems we could revert the PATH change, because of https://github.com/containerd/containerd/issues/3597.

Issue #88, opened 2019-02-07
CRI conformance tests


xibz:
The runtime should be CRI conformant. To verify that the runtime is we need to run a set of tests to ensure this provided by containerd, ie [CRI validation tests](https://github.com/containerd/cri/blob/master/docs/testing.md#cri-validation-test), [CRI e2e (end to end) tests](https://github.com/containerd/cri/blob/master/docs/testing.md#node-e2e-test).

The validation tests should be implemented first due to e2e tests taking a long amount of time due to if something doesn't work, it'll timeout in 2 minutes for each test.

kzys:
Note that firecracker-contained binary we built is [currently not having cri plugin](https://github.com/firecracker-microvm/firecracker-containerd/blob/8bfaf4fb7ba619ca92e6e3b83cd15bcf568809d8/firecracker-control/cmd/containerd/main.go#L16), whereas [containerd includes the plugin by default](https://github.com/containerd/containerd/blob/95301fee3a39ade71aff0df0dcbce0548d9a6798/cmd/containerd/builtins_cri_linux.go#L21).

Building cri with firecracker-containerd should be possible, but it would add a bunch of "replace" directives on our go.mod, [due to the way Kubernetes manages Go packages](https://github.com/kubernetes/kubernetes/issues/79384).

alexandrevilain:
Hi @kzys !

Containerd is now supporting go modules (https://github.com/containerd/containerd/pull/4760) and cri has been moved to containerd's main repo (https://github.com/containerd/containerd/tree/master/pkg/cri).

It think that we can work back on this issue.

Would you like to update #255 ? Or I can create a new one if you don't have time to work on it :)

FFMMM:
@kzys is there an opportunity for us to resume work here? Or is this issue complete via #255 ?

PR #255, opened 2019-09-03
Include CRI support

kzys:
*Issue #, if available:*

#88

*Description of changes:*

This change adds CRI support on firecracker-containerd, by including the CRI plugin.

Since Go's plugin system is pretty picky about the versions of the dependencies, it is hard to build either CRI support or Firecracker support as a plugin, separately from containerd itself.

This change adds both of them in firecracker-containerd, to unblock CRI conformance verification work.


go.mod
```diff
@@ -23,29 +26,84 @@ require (
...
-	gotest.tools v2.2.0+incompatible // indirect
+	k8s.io/klog v0.4.0 // indirect
+	k8s.io/kubernetes v1.15.3 // indirect
+	k8s.io/utils v0.0.0-20190829053155-3a4a5477acf8 // indirect
 )
 
 // Workaround for github.com/containerd/containerd issue #3031
 replace github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
+
+replace k8s.io/api => k8s.io/api v0.0.0-20190620084959-7cf5895f2711
```
sipsma: Just to make sure I'm understanding correctly, we need these replace directives because they are indirect dependencies and we thus can't specify specific versions in our go.mod? Kind of unfortunate if so and seems like a major deficiency in go modules, but if there's no better way than it is what it is.

May be worth including a comment in the go.mod file explaining this and that they come from the update-gomod.sh script

update-gomod.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
```
sipsma: nit: maybe `update-cri-gomod.sh` or similar would be a slightly better name so it's apparent this just handles cri-related dependencies

go.mod
kzys: The reality is a bit sad than that. We cannot even build our binaries without having a bunch of replace directives due to https://github.com/kubernetes/kubernetes/issues/79384. 

From the issue above;

> k8s.io/kubernetes is not primarily intended to be consumed as a module. Only the published subcomponents are (and go get works properly with those).

Ideally speaking, the CRI plugin should depend on the published subcomponents, but it may not be possible right now. I haven't looked how hard it would be though.

update-gomod.sh
kzys: Okay. Will rename the script.

kzys:
I'm going to update containerd's dependencies first.

kzys:
Let me close the PR for now, since I wouldn't be working on that right now and containerd's go mod migration (which is in progress) may make the situation better in future.

Issue #100, opened 2019-02-07
Runtime should support networking


xibz:
- [ ] runtime should support DNS config
- [ ] runtime should support port mapping with host port and container port
- [ ] runtime should support port mapping with only container port

sipsma:
Some preliminary thoughts on this in relation to the CNI support being added to Firecracker-containerd:

The CNI support for Firecracker-containerd allows VMs as a whole to be spun up in a CNI-configured network namespace, so containers that want access to that CNI-configured network need to use their "host" netns (which is actually the Firecracker VM guest's netns).

For CRI-containerd specifically, one valid approach may be to specify that the "sandbox" container use the "host" netns, which will then result in all other containers in the pod using that same netns. This in theory should have an end effect of all containers in the pod have network access to the VM's CNI-configured network.

PR #257, opened 2019-09-12
Add support for CNI-configured VM network interfaces.

### LINKED ISSUE #14: Implement networking
Integrate CNI into the appropriate components such that MicroVM-enclosed containers have network access. 

Because Firecracker requires the use of Linux tap devices, many existing CNI plugins will not work. We will need to design one or more plugins to facilitate various network configurations.

sipsma:

Closes #14 


internal/netTestutils.go
```diff
@@ -0,0 +1,144 @@
+// Copyright 2018-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
samuelkarp: nit: this filename seems a bit...weird. `net_test_utils.go`?  Or maybe `net_utils_test.go` so it only gets compiled into the test binary and not by default?

proto/types.proto
```diff
@@ -16,11 +16,83 @@ message ExtraData {
 
 // Message to specify network config for a Firecracker VM
 message FirecrackerNetworkInterface {
+	bool AllowMMDS = 1; // Specifies if metadata service should be available on this network interface
```
samuelkarp: So, uh, generally renumbering fields isn't good here as it's backwards incompatible.  But so is removing fields, and we're doing that because this is a backwards-incompatible change anyway.

I don't think you need to do anything here, but I'm calling this out because it should be abnormal.

proto/types.proto
```diff
@@ -16,11 +16,83 @@ message ExtraData {
...
+// generate a network interface for a Firecracker VM
+message CNIConfiguration {
+	// NetworkName is the name of a CNI network (as found in CNI
+	// configuration files) that will be used to generate the
+	// network interface.
+	string NetworkName = 1;
+
+	// IfName corresponds to the CNI_IFNAME parameter that will be
+	// provided to CNI plugins during invocation.
+	string IfName = 2;
```
samuelkarp: :( I still don't like this name.  Do you still prefer to leave it as `IfName` instead of `InterfaceName`?

runtime/cni_integ_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	localServices, err := internal.NewLocalNetworkServices(webpages)
+	require.NoError(t, err, "failed to create local network test services")
+
+	cniNetworkName := "fcnet-test"
+	err = writeCNIConf("/etc/cni/conf.d/fcnet-test.conflist", cniNetworkName, localServices.DNSServerIP())
+	require.NoError(t, err, "failed to write test cni conf")
+
+	go func() {
+		require.NoError(t, localServices.Serve(ctx), "failed serving local dns and http")
```
samuelkarp: This `require.NoError` might not actually catch an error that occurs later.  You might want to use a waitgroup and cancellable context here too.

```go
serveCtx, serveCancel := context.WithCancel(ctx)
var serveWait sync.WaitGroup
serveWait.Add(1)
go func() {
	defer serveWait.Done()
	require.NoError(t, localServices.Serve(ctx), "failed serving local dns and http")
}()

// ...

serveCancel()
serveWait.Wait()
```

runtime/cni_integ_test.go
```diff
@@ -0,0 +1,212 @@
...
+				containerName,
+				containerd.WithSnapshotter(naiveSnapshotterName),
+				containerd.WithNewSnapshot(snapshotName, image),
+				containerd.WithNewSpec(
+					oci.WithProcessArgs("/usr/bin/wget",
+						"-q",      // don't print to stderr unless an error occurs
+						"-O", "-", // write to stdout
+						localServices.URL(vmID)),
+					firecrackeroci.WithVMID(vmID),
+					firecrackeroci.WithVMNetwork,
```
samuelkarp: How come this is specified here but not in `taskworkflow.go`?

I'm wondering if this is something we really want the client to have to specify or not.  We might want to default to using the VM network when nothing is specified and to allow opting out instead.

internal/netTestutils.go
```diff
@@ -0,0 +1,144 @@
...
+// paths defined in the keys of "webpages", with the content set
+// in the values of that map.
+func NewLocalNetworkServices(webpages map[string]string) (LocalNetworkServices, error) {
+	testDevName := "testdev0"
+	ipAddr := "10.0.0.1"
+	ipCidr := fmt.Sprintf("%s/32", ipAddr)
+
+	output, err := exec.Command("ip", "tuntap", "add", testDevName, "mode", "tun").CombinedOutput()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to add tun dev, \"ip\" output: %s", string(output))
```
samuelkarp: nit: you can use \` instead of " for strings where you want to embed more "s.  Also, the `%q` format verb can be useful in quoting/escaping a string:

```
errors.Wrapf(err, `failed to add tun dev, "ip" output: %q`, string(output))
```

internal/netTestutils.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+type dnsHandler struct {
+	records map[string]string
+}
+
+func (h dnsHandler) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
+	msg := dns.Msg{}
+	msg.SetReply(r)
+	if r.Question[0].Qtype == dns.TypeA {
```
samuelkarp: Does this return `NXDOMAIN` if it's not a question for an `A` record?

internal/netTestutils.go
```diff
@@ -0,0 +1,144 @@
...
+		return nil, errors.Wrapf(err, "failed to assign ip to tun dev, \"ip\" output: %s", string(output))
+	}
+
+	output, err = exec.Command("ip", "link", "set", "dev", testDevName, "up").CombinedOutput()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to set tun dev up, \"ip\" output: %s", string(output))
+	}
+
+	return &localNetworkServices{
+		domain:   "phony.domain.test",
```
samuelkarp: Do you want to embed this in a constant, and then have the DNS server only serve records for this name?

docs/quickstart.md
```diff
@@ -80,11 +80,14 @@ cd ~
 #   overlay
 # * firecracker-containerd, an alternative containerd binary that includes the
 #   firecracker VM lifecycle plugin and API
+# * tc-redirect-tap and other CNI dependencies that enable VMs to start with
```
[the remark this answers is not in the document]
sipsma: Well, I'm glad I did this (tried just running it on a fresh i3.metal w/ Debian AMI) because it revealed a bunch of issues, some pre-existing, some new:
1. The pre-existing line `exec newgrp docker` appears to just end the script (I guess since it's exec'ing to a different process?). 
   * I got rid of this line and use `sg` to run with the docker group when it's needed, let me know if there's a better way.
1. I forgot during my previous runs that I was using my fixed `ptp` plugin that didn't have the DNS issue I found.
   * [I have the PR out to fix that now](https://github.com/containernetworking/plugins/pull/388), but for now I added a note to the getting-started guide that DNS resolution may not work until that is fixed in `ptp`.
   * I will create an issue to follow up on updating the getting-started guide if my `ptp` fix is not merged by the time this commit is merged.
1. I had not tested with the image-builder rootfs (just the default one), which revealed the need for adding an empty `/etc/hosts` file to the image-builder rootfs (though Noah coincidentally just added that to his unrelated PR)
   * I added that to this PR too
1. The PTP plugin seems to fail to add iptables rules that allow the VM outbound internet access, though the VM is capable of pinging the IPs on the host itself. I highly suspect this is due to the fact that I ran this on the host (previously just inside a container) where there are all sorts of iptables rules added automatically by docker.
   * I added a note to the getting-started guide that while the demo setup should give the VM access to host IPs, it may or may not grant the VM outbound internet access depending on the details of the user's host network.

I don't think I can do it in this PR, but I think we should probably add an example test that just runs the quickstart script so we don't get into this situation again. I'll add an issue for following up on that.

internal/netTestutils.go
samuelkarp: `dig` should give you more clear answers for DNS-specific questions than `wget`

docs/quickstart.md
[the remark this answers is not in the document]
samuelkarp: > The pre-existing line exec newgrp docker appears to just end the script (I guess since it's exec'ing to a different process?).

Are you running this as a script or as an interactive session?  When `exec newgrp docker` is run in an interactive session, it spawns a new shell that you can continue to use for commands.

I think `sg` is fine too.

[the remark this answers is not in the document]
samuelkarp: > I added a note to the getting-started guide that while the demo setup should give the VM access to host IPs, it may or may not grant the VM outbound internet access depending on the details of the user's host network.

Is this something we can fix for the quickstart since we're specifying exactly what should be run on the host?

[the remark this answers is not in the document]
sipsma: > Are you running this as a script or as an interactive session?

I pasted the script into a file and just did`./quickstart.sh`

> Is this something we can fix for the quickstart since we're specifying exactly what should be run on the host?

Not exactly, the issue appears to be that the ptp plugin sometimes succeeds in setting up iptables rules and sometimes doesn't. I haven't dug into the exact details as to why. My suspicion is that it's because when I ran previously, it was inside a container with just a single veth endpoint to forward to, whereas when I ran on the host i3.metal instance, there were multiple interfaces with IPs and a ton of iptables rules setup by docker for its default bridged network.

Given this is just a demo a users would inevitably need to customize for their own network/use-case anyways, I think I'm okay with it, but I'll dig a bit more to see if I can find exactly where the ptp plugin is failing and there's anything else we can do to make it more consistent (I might try appending the `firewall` cni plugin, which claims to setup host network forwarding).

internal/netTestutils.go
sipsma: I'll give that a try

docs/quickstart.md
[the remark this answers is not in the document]
samuelkarp: > Given this is just a demo a users would inevitably need to customize for their own network/use-case anyways [...]

It is, but it's also meant to be as much of a working setup as we can make it.  If it's not something that you think you can address right now, can you open an issue to track it?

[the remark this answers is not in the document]
sipsma: My concern is that it's probably just impossible for this to *universally* work; there's always going to be some particular network setup on their host that's going to result in the VM not having outbound access. But I agree we should try to minimize those cases as much as possible; we should just also leave the disclaimer in too I think.

I just tried out adding the CNI `firewall` plugin. I think it actually uncovered a bug in `firewall` because at first I was not able to use it due to the fact that its delete method doesn't conform to the CNI spec; it returns an error when there was no previous result (which the CNI spec explicitly says it shouldn't do) which causes us to hit [the error on this line](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/e4d1f825a38ade07bd1e9dc2bfd64abb7f0331f2/network.go#L324-L336).

We can probably fix that issue upstream, but in the meantime I tried with a forked go-sdk that just logs that error instead of returning it. Then, using `firewall` actually does allow the VM to get outbound access as expected even on the more complicated network present on the i3.metal host.

Given all of the above, I'll follow up by:
1. Trying to address the issue with the `firewall` plugin upstream
2. Adding an issue to include the `firewall` plugin in this demo-setup once it has been fixed.
3. Opening an issue in the Go SDK to add an optional "Force" parameter or something that allows users to ignore errors returned by delete methods (obviously it will default to false), since apparently it's not that uncommon to find plugins that don't perfectly conform to the spec.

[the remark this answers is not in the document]
samuelkarp: > My concern is that it's probably just impossible for this to *universally* work; there's always going to be some particular network setup on some host that's going to result in the VM not having outbound access.

Right, I'm not saying that the quickstart needs to account for every situation, but rather that it accounts for itself.  Assuming someone follows the quickstart guide exactly, it should work.

docs/networking.md
```diff
@@ -117,51 +118,45 @@ VMs will execute in.
...
 }
 
-message StaticIPConfiguration {
+message IPConfiguration {
     // Network configuration that will be applied to a network interface in a
     // Guest VM on boot.
-    string IP;
-    string SubnetMask;
-    string DefaultGateway;
+    string PrimaryAddr;
```
samuelkarp: Any particular reason why we have "PrimaryAddr" and "MacAddress" (i.e., abbreviated in one place and spelled out in another)?

runtime/cni_integ_test.go
samuelkarp: I think we can have users specify the option for now.  I'm trying to anticipate what it would look like to try and use the existing cri-containerd plugin with firecracker-containerd, and seeing whether we can avoid requiring changes.  Might be worth opening an issue for this so we can follow up later on.

tools/docker/Dockerfile
```diff
@@ -183,9 +187,16 @@ COPY _submodules/runc/runc /usr/local/bin
 COPY --from=firecracker-vm-root-builder /output/vm.ext4 /var/lib/firecracker-containerd/runtime/default-rootfs.img
 COPY tools/docker/firecracker-runtime.json /etc/containerd/firecracker-runtime.json
 
+# pull the images the tests need into the content store so we don't need internet
+# access during the tests themselves
 COPY tools/docker/naive-snapshotter/config.toml /etc/containerd/config.toml
-COPY tools/docker/naive-snapshotter/entrypoint.sh /entrypoint
+RUN containerd 2>/dev/null & \
```
samuelkarp: I hate it but I don't have any suggestions that don't involve a lot more work than we want to do right now...

docs/networking.md
sipsma: Nope, changed to `Address`

docs/quickstart.md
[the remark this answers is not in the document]
sipsma: Yep, that's why I updated the getting-started guide to specify that the expected end result is that the host IPs are accessible and that the outside internet *may* be accessible but might not depending on the details of the host network. I've added a similar note to the quickstart guide too now.

Once this PR is merged I'll create an issue for updating to use the `firewall` plugin, then once I fix the plugin upstream we can reduce the number of cases where outbound internet access doesn't work (but will still need to leave the qualification anyways).

I also created an issue to run the quickstart as a test case, which we should do no matter what: https://github.com/firecracker-microvm/firecracker-containerd/issues/263

internal/netTestutils.go
sipsma: Tried it, it was not failing with `NXDOMAIN` specifically, so I updated the code to now return `NXDOMAIN` if the name is not found. I don't think it makes a difference at the moment but good catch either way, it may matter in the future.

runtime/cni_integ_test.go
samuelkarp: > What do you think we should open an issue about? Finding a way to set these options by default?  Or figuring out how this will interact with CRI?

I think "figuring out how this will interact with CRI" is the right answer right now.  It *might* also be reasonable to set these options by default.  Either way, I think it'd be good to move this conversation to an issue.

> My impression was that CRI is making containerd CreateTask calls itself

So I'd like to disambiguate a bit.  CRI is a [spec](https://kubernetes.io/blog/2016/12/container-runtime-interface-cri-in-kubernetes/) with multiple implementations.  [cri-containerd](https://github.com/containerd/cri) is a containerd plugin that implements CRI and is bundled by default into containerd.

> We'd instead want just CRI to create its sandbox container using "host" netns (which would actually be the Firecracker VM's).

Yeah, this is probably a reasonable approach as well.

sipsma: Commented on this issue, can continue discussion there: https://github.com/firecracker-microvm/firecracker-containerd/issues/100

docs/quickstart.md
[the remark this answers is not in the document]
sipsma: Created issue for the firewall plugin here: https://github.com/firecracker-microvm/firecracker-containerd/issues/265

Issue #259, opened 2019-09-16
Upgrade to Firecracker v0.18


sipsma:
We need to upgrade to the latest Firecracker release: https://github.com/firecracker-microvm/firecracker/releases/tag/v0.18.0

The largest change here is going to be using the new production vsock support, but will also need to deal with any other miscellaneous changes and updating the getting-started/quickstart guides.

PR #260, opened 2019-09-16
Build firecracker from a submodule.

sipsma:
This is intended to ease development with Firecracker and version upgrades.

Note this does not upgrade Firecracker to a new version ( ref #259 ), it keeps the version at 0.17 for now, but it should make moving to 0.18 easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,9 +23,18 @@ export DOCKER_IMAGE_TAG?=latest
 GOPATH:=$(shell go env GOPATH)
 BINPATH:=$(abspath ./bin)
 SUBMODULES=_submodules
+UID:=$(shell id -u)
+
+FIRECRACKER_DIR=$(SUBMODULES)/firecracker
+FIRECRACKER_TARGET?=x86_64-unknown-linux-musl
+FIRECRACKER_BIN=$(FIRECRACKER_DIR)/target/$(FIRECRACKER_TARGET)/release/firecracker
+JAILER_BIN=$(FIRECRACKER_DIR)/target/$(FIRECRACKER_TARGET)/release/jailer
+FIRECRACKER_BUILDER_NAME?=firecracker-builder
+CARGO_CACHE_VOLUME_NAME?=cargocache
+
 RUNC_DIR=$(SUBMODULES)/runc
 RUNC_BIN=$(RUNC_DIR)/runc
-UID:=$(shell id -u)
+RUNC_BUILDER_NAME?=runc-builder
 
 # Set this to pass additional commandline flags to the go compiler, e.g. "make test EXTRAGOARGS=-v"
 export EXTRAGOARGS?=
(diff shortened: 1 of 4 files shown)


tools/docker/Dockerfile.firecracker-builder
```diff
@@ -0,0 +1,25 @@
...
+# License is located at
+#
+# 	http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+FROM rust:1.32-stretch
```
samuelkarp: Any particular reason you're using `-stretch` instead of the default or instead of `-buster`?

sipsma: I did not see a buster image for rust 1.32 on dockerhub (only for newer rust versions), so I don't think they ever released one.

mxpv: 2. We probably should target same version of Rust as Firecracker:
https://github.com/firecracker-microvm/firecracker/blob/master/tools/devctr/Dockerfile.x86_64#L7
(or even we may consider using the Dockerfile from their repo as base to keep up)

sipsma: That Dockerfile is for Firecracker v0.18 (which updates to rust 1.35). We are still building Firecracker v0.17 for now, which uses an older rust toolchain.

I considered using their Dockerfile, but it's doing a lot more than we need here and not necessarily compatible with the way we structure our CI system, so I felt it would be simpler overall for us to just have this tiny Dockerfile of our own. We can reconsider in the future if maintaining this Dockerfile becomes meaningful overhead.

Makefile
```diff
@@ -23,9 +23,17 @@ export DOCKER_IMAGE_TAG?=latest
 GOPATH:=$(shell go env GOPATH)
 BINPATH:=$(abspath ./bin)
 SUBMODULES=_submodules
+UID:=$(shell id -u)
+
+FIRECRACKER_DIR=$(SUBMODULES)/firecracker
+FIRECRACKER_BIN=$(FIRECRACKER_DIR)/target/x86_64-unknown-linux-musl/release/firecracker
+JAILER_BIN=$(FIRECRACKER_DIR)/target/x86_64-unknown-linux-musl/release/jailer
```
nmeyerhans: Can we make these paths architecture independent without too much work? At some point we'll need arm64, at least.

Makefile
```diff
@@ -42,11 +50,14 @@ clean:
...
 	$(MAKE) -C $(RUNC_DIR) clean
+	$(MAKE) firecracker-clean
 	rm -f *stamp
 	$(MAKE) -C tools/image-builder clean-in-docker
 
 distclean: clean
-	docker rmi localhost/runc-builder:latest
+	docker rmi localhost/$(RUNC_BUILDER_NAME):$(DOCKER_IMAGE_TAG)
+	docker rmi localhost/$(FIRECRACKER_BUILDER_NAME):$(DOCKER_IMAGE_TAG)
+	docker volume rm -f $(CARGO_CACHE_VOLUME_NAME)
```
nmeyerhans: these docker commands should be prefixed with '-' so they don't count as a failure if a given image doesn't exist.

Raise the bar by providing a make function that checks to see if an image exists, exits true if not, tries to delete it if so (at that point you can fail meaningfully if it doesn't exist)

tools/docker/Dockerfile
```diff
@@ -179,11 +152,12 @@ RUN mkdir -p /var/lib/firecracker-containerd/runtime \
 	&& echo "882fa465c43ab7d92e31bd4167da3ad6a82cb9230f9b0016176df597c6014cef default-vmlinux.bin" | sha256sum -c - \
 	&& mv default-vmlinux.bin /var/lib/firecracker-containerd/runtime/default-vmlinux.bin
 
+COPY _submodules/firecracker/target/x86_64-unknown-linux-musl/release/firecracker /usr/local/bin/
+COPY _submodules/firecracker/target/x86_64-unknown-linux-musl/release/jailer /usr/local/bin/
```
nmeyerhans: Same question as above regarding architecture independent paths.

Makefile
sipsma: I learned some more Make and added a macro for conditionally deleting the images. For the docker volume, the "-f" gets us that effect already (there is also a -f flag for rmi but it does much more that we don't want).

Makefile
sipsma: Yeah, for now I just updated to use an explicit `FIRECRACKER_TARGET` var that is used in the Makefile and also passed to the Dockerfile as a build-arg.

PR #262, opened 2019-09-18
Use image-builder during test execution

nmeyerhans:
This change updates the test configuration to use a root filesystem image generated by the `tools/image-builder` component. This serves 3 purposes:

1. It gets us test coverage of the image-builder, which was previously lacking
2. It eliminates duplicate functionality (we had two different mechanisms for building root images)
3. image-builder images use read/write layers on top of read-only images, meaning they're safe for concurrent use by multiple VMs.


tools/image-builder/Makefile
```diff
@@ -87,11 +88,17 @@ builder_stamp:
...
+	   else \
+	     $(MAKE) clean-in-docker ;\
+	   fi ; \
+	fi
 
-distclean: clean-in-docker
+distclean: clean
 	rm -rf files_ephemeral
-	docker rmi fc-image-builder:latest
+	- docker rmi fc-image-builder:latest
```
sipsma: If the `latest` tag is hardcoded, then I think parallel runs of the CI system will end up referencing the same image and probably interfere with each other. 

There an env var `DOCKER_IMAGE_TAG` that is set in in buildkite runs to a unique ID per-run, so I think we want to use that for the image-builder too now that it's being used in the CI.

tools/image-builder/Makefile
```diff
@@ -56,6 +56,7 @@ endif
 		--include=udev,systemd,systemd-sysv,procps,libseccomp2 \
 		stretch \
 		"$(WORKDIR)" $(DEBMIRROR)
+	touch "$(WORKDIR)/etc/hosts"
```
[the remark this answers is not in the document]
nmeyerhans: Nope, no reason to do it this way. Will fix.

tools/image-builder/Makefile
nmeyerhans: Good point, thanks. Will fix.

tools/image-builder/Makefile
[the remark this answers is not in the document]
nmeyerhans: Fixed. Hosts file contains meaningful content, too, in case it gets used as-is instead of as a target for a bind mount.

tools/image-builder/Makefile
nmeyerhans: Fixed.

nmeyerhans:
Because of the way buildkit's agent resets the environment between steps, the `root.img` file generated by the `test-images` target wasn't available in subsequent steps when it was needed. I've updated the PR to explicitly preserve the file across steps...

PR #266, opened 2019-09-24
Upgrade to Firecracker v0.18.0.

### LINKED ISSUE #259: Upgrade to Firecracker v0.18

sipsma:
This consists of:
* Upgrading to the new Go SDK version.
* Using the new vsock interface with unix-domain socket backend on the host.

Closes #259 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,6 +165,11 @@ demo-network: $(PTP_BIN) $(HOSTLOCAL_BIN) $(TC_REDIRECT_TAP_BIN) $(FCNET_CONFIG)
 .PHONY: firecracker
 firecracker: $(FIRECRACKER_BIN) $(JAILER_BIN)
 
+.PHONY: install-firecracker
+install-firecracker: firecracker
+	install -D -o root -g root -m755 -t $(INSTALLROOT)/bin $(FIRECRACKER_BIN)
+	install -D -o root -g root -m755 -t $(INSTALLROOT)/bin $(JAILER_BIN)
+
 $(FIRECRACKER_DIR)/Cargo.toml:
 	git submodule update --init --recursive $(FIRECRACKER_DIR)
 
@@ -182,7 +187,7 @@ $(FIRECRACKER_BIN) $(JAILER_BIN): $(FIRECRACKER_DIR)/Cargo.toml tools/firecracke
 		-e HOME=/tmp \
 		--workdir /src \
 		localhost/$(FIRECRACKER_BUILDER_NAME):$(DOCKER_IMAGE_TAG) \
-		cargo build --release --features vsock --target $(FIRECRACKER_TARGET)
+		cargo build --release --target $(FIRECRACKER_TARGET)
 
 .PHONY: firecracker-clean
 firecracker-clean:
--- a/internal/vm/vsock.go
+++ b/internal/vm/vsock.go
@@ -15,125 +15,127 @@ package vm
 
 import (
 	"context"
+	"fmt"
 	"net"
-	"strings"
 	"time"
 
 	"github.com/mdlayher/vsock"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
 const (
-	vsockConnectTimeout = 20 * time.Second
+	vsockRetryTimeout      = 20 * time.Second
+	vsockRetryInterval     = 100 * time.Millisecond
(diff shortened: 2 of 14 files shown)


Makefile
```diff
@@ -175,14 +180,14 @@ tools/firecracker-builder-stamp: tools/docker/Dockerfile.firecracker-builder
 		tools/docker
 	touch $@
 
-$(FIRECRACKER_BIN) $(JAILER_BIN): $(FIRECRACKER_DIR)/Cargo.toml tools/firecracker-builder-stamp
+$(FIRECRACKER_BIN) $(JAILER_BIN): $(FIRECRACKER_DIR)/Cargo.toml $(FIRECRACKER_DIR) tools/firecracker-builder-stamp
```
mxpv: Do we need both `$(FIRECRACKER_DIR)/Cargo.toml` and `$(FIRECRACKER_DIR)` here ? As we pull submodule, `Cargo.toml` will be already there, so we can have just one dependency.

sipsma: No that has no effect right now, this is leftover from something else I was trying but didn't include here, will remove.

internal/vm/vsock.go
```diff
@@ -144,23 +228,93 @@ func VSockAcceptConnector(port uint32) IOConnector {
...
+	return true
 }
 
-func isECONNRESET(err error) bool {
-	return strings.HasSuffix(err.Error(), "connection reset by peer")
+// isTemporaryNetErr returns whether the provided error is a retriable
+// error, according to the interface defined here:
+// https://golang.org/pkg/net/#Error
+func isTemporaryNetErr(err error) bool {
+	terr, ok := err.(interface {
```
samuelkarp: Any reason not to do this:

```go
type tempErr interface {
	Temporary() bool
}

func isTemporaryNetErr(err error) bool {
	terr, ok := err.(tempErr)
	return err != nil && ok && terr.Temporary()
}
```

internal/vm/vsock.go
```diff
@@ -15,125 +15,209 @@ package vm
...
+}
 
-	// Retries occur every 10ms up to vsockConnectTimeout
-	const retryInterval = 10 * time.Millisecond
-	ctx, cancel := context.WithTimeout(reqCtx, vsockConnectTimeout)
+func (l vsockListener) Accept() (net.Conn, error) {
+	ctx, cancel := context.WithTimeout(l.ctx, vsockRetryTimeout)
 	defer cancel()
 
+	tryAccept := func() (net.Conn, error) {
```
samuelkarp: Same question here: does this function need to be or benefit from being defined inline?

runtime/service_integ_test.go
```diff
@@ -551,7 +551,7 @@ func TestStubBlockDevices_Isolated(t *testing.T) {
 
 		const expectedOutput = `
 NAME MAJ:MIN RM      SIZE RO | MAGIC
-vda  254:0    0 53633024B  0 |  104 115 113 115 128  28   0   0
+vda  254:0    0 53641216B  0 |  104 115 113 115 128  28   0   0
```
samuelkarp: How come this changed?

internal/vm/vsock.go
```diff
@@ -144,23 +228,93 @@ func VSockAcceptConnector(port uint32) IOConnector {
...
+// socket deadlines to none after returning.
+func tryConnWrite(conn net.Conn, expectedWrite string, timeout time.Duration) error {
+	conn.SetDeadline(time.Now().Add(timeout))
+	defer conn.SetDeadline(time.Time{})
+
+	bytesWritten, err := conn.Write([]byte(expectedWrite))
+	if err != nil {
+		return err
+	}
+	if bytesWritten != len(expectedWrite) {
```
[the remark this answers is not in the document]
samuelkarp: As discussed offline, since this isn't used as the main read/write implementation but is rather used only for the CONNECT and ack messages, I think that's fine.

internal/vm/vsock.go
samuelkarp: Didn't realize it was pre-existing, feel free to ignore.  And yeah, I don't think it's massively different except for avoiding an inline interface and having a named interface instead.

Issue #247, opened 2019-08-22
Feature request for StopVM API


haikuoliu:
Hi,

Currently the “StopVM” API does the following in order:

1. Shutdown fc-agent
2. Send a SIGTERM to VM
3. Remove shim dir

Any of them could produce an error, the error message is good but the error type is not well defined, we don't know if the failure is from shim dir removal or stopping VM. To integrate with firecracker-containerd, we want to take different actions on different errors, if the error comes from removing shim dir, that's probably fine and we can continue, but if the VM itself is not stopped, we may want to trigger an alarm or something because it's a more critical failure. So is it possible for firecracker-containerd to define error type for "StopVM" API?

Also, reading from code, I see the SIGTERM is sent to VM to gracefully stop it, is it possible to set a timeout for SIGTERM, and SIGKILL (not sure if you are already doing it) will be sent by firecracker-containerd in case SIGTERM is not working?

Let me know if it makes sense.
Thanks,
Haikuo

nmeyerhans:
I agree with the suggestion to add more specific error types.

I think the timeout + SIGKILL logic should be left to the application. Based on the type of error returned upon failure, you may want to respond in various ways. Falling back to a call to `StopVM` with a "force" flag set, or passing a custom signal (e.g. KILL) might be reasonable ways to implement this from the SDK side.

haikuoliu:
> I think the timeout + SIGKILL logic should be left to the application. Based on the type of error returned upon failure, you may want to respond in various ways. Falling back to a call to StopVM with a "force" flag set, or passing a custom signal (e.g. KILL) might be reasonable ways to implement this from the SDK side.

Yeah it would be good as long as there is a parameter in the "StopVM" API to stop VM forcefully (send SIGKILL).

sipsma:
Also agree with using more specific error types.

For the SIGTERM vs. SIGKILL issue, the current behavior actually should be that the VM gets a SIGKILL in the even that the graceful shutdown fails. This is due to the fact that the VM process's context get cancelled, [see this line](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/runtime/service.go#L1004) and the [`cmd.Exec` docs](https://golang.org/pkg/os/exec/#CommandContext)

haikuoliu:
@sipsma Thanks for the link, but I think it's still worthy providing options to caller. Let's say I define a 1 min timeout for `StopVM`, when the timeout happens, there might be two cases:
[...]

PR #272, opened 2019-10-02
Added timeout for VMshutdown

Zyqsempai:
Issue #247

Added timeout for VM Shutdown
Added special type of error for case when timeout exceeded 


Zyqsempai:
@sipsma PTAL

runtime/service.go
```diff
@@ -67,6 +67,7 @@ const (
 	defaultVsockPort        = 10789
 	minVsockIOPort          = uint32(11000)
 	firecrackerStartTimeout = 5 * time.Second
+	stopVMTimeoutSeconds    = 5
```
sipsma: nit: For this variable, I think it should be a `time.Duration` (so just `5 * time.Second`) so it can stay agnostic of particular units.

Also, can this be named more like `defaultStopVMTimeout`? Just to make it clear it's a default value that can be optionally overridden.

runtime/service.go
```diff
@@ -514,14 +521,20 @@ func (s *service) StopVM(requestCtx context.Context, request *proto.StopVMReques
 		return nil, err
 	}
 
-	// The graceful shutdown logic, including stopping the VM, is centralized in Shutdown. We set "Now" to true
-	// to ensure that the service will actually shutdown even if there are still containers being managed.
-	_, err = s.Shutdown(requestCtx, &taskAPI.ShutdownRequest{Now: true})
-	if err != nil {
-		return nil, err
+	for {
```
sipsma: I'm not sure we need the `for` statement wrapping the `select`, unless I'm missing something we should be good with just the `select` here

proto/firecracker.proto
```diff
@@ -36,6 +36,7 @@ message CreateVMRequest {
 
 message StopVMRequest {
     string VMID = 1;
+    uint32 VMShutDownTimeOut = 2;
```
sipsma: For this variable, units matter, so can we name it more like `VMShutdownTimeoutSeconds`?

sipsma:
Oh also, the tests are failing just because your commit `Added timeout for VMshutdown` doesn't have a DCO (your other commit does), so please just amend that commit with a signoff when updating.

runtime/service.go
sipsma: Actually, I took a second look and think we will need a little bit more than just removing the `for`. We want to make sure that if we hit the timeout while blocked on the `Shutdown` call, the `StopVM` method doesn't stay blocked. I think the simplest way to do this will be running `Shutdown` in its own goroutine. Roughly sketched out:
```
shutdownCh := make(chan error)
go func() {
    defer close(shutdownCh)
    _, err = s.Shutdown(...)
    shutdownCh <- err
}()

select {
    case <- ctx.Done():
        ...
    case err = <-shutdownCh:
        if err != nil {
            return nil, err
        }
        return Empty{}, nil
}
```

sipsma:
@Zyqsempai Saw in the test runs of your latest commits that the `DeadlineExceeded` error was being returned unexpectedly in the test case, which didn't make any sense since the code you wrote looked good. It took a while, but I think I tracked down why. It's really really tricky (but fortunately easy to fix I think).

The fix should be to not create a new `ctx` derived from the `requestCtx` in `StopVM`. Instead, just use the existing `requestCtx` unmodified. Then, use a`time.Timer` to enforce the timeout, something along the lines of:
```
func (s *service) StopVM(...) {
    ...
    timer := time.NewTimer(timeout)
    ...
    select {
        case <-timer.C:
            ...
        case err = <-shutdownCh:
            ...
    }
}
```

To explain what I think the actual problem is, containerd gives our code [a context object during initialization in the NewService method, which we decided to call `shimCtx`](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/runtime/service.go#L140). This context represents the execution of the shim process as a whole; when it is cancelled that means the process is shutting down (and thus all the services we run need to shut down too).

One place we give the `shimCtx` to is [the server for the firecracker-control API](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/runtime/service.go#L260), which is what serves the `StopVM` API (among others). The TTRPC server implementation uses [that context](https://github.com/containerd/ttrpc/blob/master/server.go#L73) to decide when to shutdown, but it also derives [the `requestCtx` objects for all its methods](https://github.com/containerd/ttrpc/blob/master/server.go#L420) from that context.

As a result, [the `requestCtx` provided to `StopVM`](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/runtime/service.go#L511) is a child context of `shimCtx`. This by itself is reasonable; if the shim is shutting down then ongoing requests should get cancelled too. However, inside [the `Shutdown` method, we cancel the `shimCtx`](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/runtime/service.go#L1029) in order to actually start the full shutdown process, which is also reasonable by itself.

Putting all the above together, canceling `shimCtx` cancels `requestCtx`, which in turn cancels the new `ctx` your code added, so then when we hit the select statement in the new code we always hit the `<-ctx.Done()` case first and return `DeadlineExceeded`. Getting rid of that new extra `ctx` should thus fix the problem (and keep things from getting even more complicated than they already are). Let me know if this makes sense, it's obviously not straightforward at all.

Zyqsempai:
@sipsma Yep, it's really tricky, but totally make sense.

runtime/service.go
```diff
@@ -514,14 +522,22 @@ func (s *service) StopVM(requestCtx context.Context, request *proto.StopVMReques
...
-	// The graceful shutdown logic, including stopping the VM, is centralized in Shutdown. We set "Now" to true
-	// to ensure that the service will actually shutdown even if there are still containers being managed.
-	_, err = s.Shutdown(requestCtx, &taskAPI.ShutdownRequest{Now: true})
-	if err != nil {
-		return nil, err
-	}
+	shutdownCh := make(chan error)
+	go func() {
+		defer close(shutdownCh)
+		_, err = s.Shutdown(requestCtx, &taskAPI.ShutdownRequest{Now: true})
```
sipsma: just noticed this, let's use `_, err := s.Shutdown(...)` (`:=` instead of `=`), just to avoid the confusion that can result from overwriting a named return val from within a separate goroutine.

Zyqsempai:
@sipsma Done, thanks for you comments. PTAL.

kzys:
Two (a bit bikeshed-y) questions:

- If a customer do want to shutdown the VM regardless of the time it takes, what would be the right way? Giving large enough timeout?
- Do we want to have server-side timeout for all of operations? It is not clear why StopVM can have the timeout, but others don't.

sipsma:
>     * If a customer do want to shutdown the VM regardless of the time it takes, what would be the right way? Giving large enough timeout?

Yes, this change essentially prevents StopVM from ever being blocked forever waiting for the shutdown, which I think is preferable as I can't think of any situation in which being blocked *literally* forever is the desired behavior for stopping the VM.

>     * Do we want to have server-side timeout for all of operations? It is not clear why StopVM can have the timeout, but others don't.

I think adding it to StopVM is particularly valuable because it ensures that users don't get in a situation where they are unable cleanup resources on their machine (once the timeout is hit, we just send SIGKILL).

The cases where other operations besides StopVM get blocked forever are still not good, but the worst case scenario is the user doesn't create a VM or get information about their VM, which is somewhat less harmful than being unable to remove resources from their machine. That being said, I agree that adding timeouts to those other operations could still be valuable and is something to consider in the future (perhaps just not in this PR).

kzys:
Thanks. I agree that blocking literally forever won't be what customers want.

How about renaming the field from `VMShutdownTimeoutSeconds` to `TimeoutSeconds` or simply `Timeout`, if we may add the field in other requests?

runtime/service.go
```diff
@@ -514,14 +522,22 @@ func (s *service) StopVM(requestCtx context.Context, request *proto.StopVMReques
...
+	go func() {
+		defer close(shutdownCh)
+		_, err := s.Shutdown(requestCtx, &taskAPI.ShutdownRequest{Now: true})
+		shutdownCh <- err
+	}()
 
-	return &empty.Empty{}, nil
+	select {
+	case <-timer.C:
+		return nil, status.Error(codes.DeadlineExceeded, "timed out waiting for VM shutdown")
```
IRCody: In this case do we need err to be set to != nil so that shimCancel is called in the deferred func above?

xibz:
I wonder if it may be best to a grpc.ClientConn constructor function. You can set a timeout for the individual methods there. I wonder if that is preferable rather than adding a timeout field, and this would allow us to make use of the context that is being passed in. Thoughts?

runtime/service.go
sipsma: It's not straightforward, but that is what's happening implicitly here: https://play.golang.org/p/xqGlTUUX8jq

IRCody: ah I missed that it had a named return for err. Thanks!

sipsma:
> Can this not be handled by setting a context deadline when making the request? Rather than an API change?

I think having the client specify the timeout in the `context` of their requests raises a bunch of questions that make the behavior harder to reason about: if the timeout is hit, then will the client be unblocked right away before any response is received (as opposed to receiving a timeout error from the server)? What if the context timeout is applied to other parts of the client connection, could that mean that the user is unable to distinguish between a dial timeout and a timeout shutting down the VM? How would the user determine if SIGKILL was sent in that case?

[Also see this comment](https://github.com/firecracker-microvm/firecracker-containerd/pull/272#issuecomment-538180715); from a purely internal code perspective, using context timeouts here creates more confusion than it's worth IMO.

I think in practice it ends up being simpler to just have it be part of the API.

sipsma:
> Thanks. I agree that blocking literally forever won't be what customers want.
> 
> How about renaming the field from `VMShutdownTimeoutSeconds` to `TimeoutSeconds` or simply `Timeout`, if we may add the field in other requests?

Agree, @Zyqsempai I think that's worth making before we merge here.

runtime/service.go
```diff
@@ -499,6 +500,13 @@ func (s *service) createVM(requestCtx context.Context, request *proto.CreateVMRe
 // created yet and the timeout is hit waiting for it to exist, an error will be returned but the shim will
 // continue to shutdown.
 func (s *service) StopVM(requestCtx context.Context, request *proto.StopVMRequest) (_ *empty.Empty, err error) {
+	var timeout time.Duration
```
xibz: One minor nit commit, but otherwise looks good.

```go
timeout := defaultStopVMTimeout
if request.VMShutdownTimeoutSeconds > 0 {
    timeout = time.Duration(request.VMShutdownTimeoutSeconds) * time.Second
}
```

Zyqsempai:
Done!

PR #274, opened 2019-10-02
Use CURDIR instead of PWD in Makefile.

sipsma:
PWD is an env var, which can be unset in many scenarios (such as when a command
is prefixed with sudo without "-E"), which in turn can lead to confusing
failures when running make commands. CURDIR is set by make and avoids those
issues.


sipsma:
> This sounds [familiar](https://github.com/firecracker-microvm/firectl/commit/fc574786869aba500fe5f3a68e32cf3c56ef3230)

Apparently! This is just something we need to look out for during PRs going forward I guess

PR #282, opened 2019-10-09
Add doc proposing support for mounting drives inside VM.

sipsma:

For those who have reviewed previous versions of this doc, I've made some changes to the API in order to replace the pre-existing `AdditionalDrives` field rather than supplement it. This seems to result in a cleaner overall API at the expense of backwards-incompatibility (though it should be trivial for consumers to update). The actual substance of the proposal+implementation is the same though.

### DIFF
--- a/docs/drive-mounts-proposal.md
+++ b/docs/drive-mounts-proposal.md
@@ -1,0 +1,115 @@
+# Proposal - Drive Mount Support in Firecracker-containerd
+
+# Problem
+
+* Firecracker-containerd users may need to attach extra drives containing filesystems (i.e. those besides the rootfs and the container images) and mount them inside their VM. This can enable, for example, extra persistent storage that's bind mounted into container rootfs dirs.
+* However, they will run into some of the [same issues](https://github.com/firecracker-microvm/firecracker-containerd/blob/36b47bf0a988083d3b5b15502ce61d5909b8a798/docs/design-approaches.md#block-devices) that Firecracker-containerd did when implementing support for multiple container rootfs drives, namely that there is no way to deterministically map a drive ID outside the VM to a block device within the VM guest.
+* This leaves users with a very complicated problem to solve.
+    * One (terrible) option would be for them to add custom code to their VM rootfs that will, on VM boot, somehow figure out which block device is which, being very careful to not interfere with any stub drives that Firecracker-containerd plans on using for containers later.
+    * This essentially exposes internal implementation details (stub drives) to users to have to deal with
+
+# Potential Solutions
+
+## Option A) Extend CreateVM API (Preferred) 
+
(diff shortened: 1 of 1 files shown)


docs/drive-mounts.md
```diff
@@ -0,0 +1,115 @@
...
+    // (Optional) RateLimiter that will be applied to the backing-drive.
+    FirecrackerRateLimiter RateLimiter;
+}
+```
+
+When `DriveMounts` are specified as part of a `CreateVM` request, Firecracker-containerd will internally take care of loading the filesystem image on the host into a Firecracker drive and then mounting the filesystem on that drive inside the VM guest before the `CreateVM` call returns (i.e. as part of the VM+Agent boot process).
+
+The API update is backwards incompatible, but in such a way that upgrading to the new schema is straightforward. It results in a cleaner interface than trying to adapt the previous schema due to:
+
+1. Removing the need to either support or reject rootfs drives that have the new `VMPath`, `FilesystemType` and `Options` field specified
```
kzys: The root drive can be non-EXT2 and may have options. Isn't it?

https://github.com/firecracker-microvm/firecracker/blob/80cddef3969e97285418c494ef489fa404e9d859/docs/rootfs-and-kernel-setup.md#creating-a-rootfs-image

> For instance, our getting started guide uses an EXT4 FS image with OpenRC as an init system. Note that, whichever file system you choose to use, support for it will have to be compiled into the kernel, so it can be mounted at boot time.

I cannot see EXT4 dependency, while I haven't tested other filesystems.

sipsma: The root drive specifically is configured entirely by the Firecracker VMM and their API does not support passing filesystem type or mount options: https://github.com/firecracker-microvm/firecracker/blob/master/api_server/swagger/firecracker.yaml#L407-L431

The `FirecrackerDriveMount` objects, on the other hand, are mounted in userspace by our firecracker-containerd-specific agent. When agent does the [mount, it needs to provide the filesystem type and options](http://man7.org/linux/man-pages/man2/mount.2.html).

I believe it's *technically* possible for agent to autodetect the filesystem type before doing the mount (I think the command line `mount` utility does this), but I haven't found a pure-golang library for doing so and don't see that introducing that extra complexity would actually provide much benefit as we'd still want to support passing this `FilesystemType` field even if we managed to make it optional.

kzys: You are right. Firecracker is doing some special Linux-specific stuff for an root drive.

https://github.com/firecracker-microvm/firecracker/blob/0982209eac8a1bb3eb5f79a762a1413aaabc776c/vmm/src/lib.rs#L825

kzys:
LGTM

sipsma:
> Is docs/ the right location for this?

I think that's a good point. We have just been putting all proposals in here previously, but I could see how that would be confusing for someone looking just for docs. I will update the name of the file and the title to clarify this is a proposal like you suggested.

PR #283, opened 2019-10-10
The rootfs of a microVM should be read-only

kzys:
*Issue #, if available:*

NA

*Description of changes:*

While customers won't be able to access the rootfs, making it read-only
will reduce its attack surface.


agent/main.go
```diff
@@ -71,6 +71,11 @@ func main() {
 		log.G(shimCtx).WithError(err).Fatal("failed to set shim as subreaper")
 	}
 
+	_, err := os.Create("/verify-rootfs")
```
samuelkarp: The agent will likely have a writable view of the root filesystem (at least in the image built from our image-builder) as it runs in a writable portion of overlayfs.  I don't think we need to verify whether it's writable from within the agent, as that's already part of the system that runs in a potentially-hostile environment.  I think it should be sufficient to ensure that we attach the drive as read-only when we configure the VM.

runtime/service.go
```diff
@@ -647,12 +648,16 @@ func (s *service) buildVMConfiguration(req *proto.CreateVMRequest) (*firecracker
 
 	var driveBuilder firecracker.DrivesBuilder
 	// Create non-stub drives
+	var path = s.config.RootDrive
```
xibz: Do we want the configuration `RootDrive` field to be read only? They may want a read/write root drive. Either we make it read/write or introduce a `RootDriveIsReadOnly` or something in the configuration.

runtime/service.go
```diff
@@ -647,12 +648,16 @@ func (s *service) buildVMConfiguration(req *proto.CreateVMRequest) (*firecracker
...
 	if root := req.RootDrive; root != nil {
-		driveBuilder = firecracker.NewDrivesBuilder(root.PathOnHost)
-	} else {
-		driveBuilder = firecracker.NewDrivesBuilder(s.config.RootDrive)
```
samuelkarp: Should we remove this constructor if we're not going to use it?

runtime/service.go
samuelkarp: We likely want the root drive to be read-only by default.  Running multiple VMs from the same root drive where it's mutable can corrupt the drive; multiple writers to the same disk isn't really supported by common filesystems.  The image builder that we have in this repo will generate a VM that runs from a read-only root filesystem, but then mounts a read-write filesystem on top with overlay so that the VM thinks it's writable.

agent/main.go
kzys: Discussed offline: To automatically and continuously verify the protection, we can put a malicious agent, rather than let the production agent do things that should fail.

runtime/service.go
kzys: Customers can have different files by utilizing the image builder tool. There might be a customer who want to put files at runtime, but supporting that is possible by what @xibz suggested (adding that in a request and/or the configuration file).

However, since we don't have the customer request, we can start from the safest option -- always read-only.

runtime/service.go
kzys: I'd like to keep the constructor to be backward-compatible. The SDK is pre 1.0, but keeping the constructor doesn't cost us much.

runtime/service.go
```diff
@@ -647,12 +648,16 @@ func (s *service) buildVMConfiguration(req *proto.CreateVMRequest) (*firecracker
...
+		path = root.PathOnHost
 	}
 
+	driveBuilder = firecracker.DrivesBuilder{}.
+		WithRootDrive(path, func(d *models.Drive) {
+			d.IsReadOnly = firecracker.Bool(true)
+		})
```
sipsma: Users are able to specify that their [`RootDrive`](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/proto/firecracker.proto#L22) should be [read-only or read-write](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/proto/types.proto#L111), so don't we want to only set the default of `IsReadOnly = true` when the user didn't actually provide a root drive config? Unless I'm misunderstanding this will always set the root drive to be readonly regardless of the configuration provided in the CreateVM request.

kzys: Right. We are discarding the most of the RootDrive field on a request, but it shouldn't be.

runtime/service.go
kzys: Actually, as @sipsma pointed out, CrateVMRequest is already having a field to make the rootfs read-only, which we are discarding even before my change.

I've fixed that in the last revision. So the default is read-only but customers can override that through CrateVMRequest. The JSON configuration doesn't have the way to make the rootfs writable though.

agent/main.go
kzys: Removed on the last revision.

runtime/service.go
kzys: The last revision honors the fields from the request.

runtime/service.go
```diff
@@ -681,6 +668,33 @@ func (s *service) buildVMConfiguration(req *proto.CreateVMRequest) (*firecracker
 	return &cfg, nil
 }
 
+func (s *service) buildNonStubDrives(req *proto.CreateVMRequest) []models.Drive {
+	var builder firecracker.DrivesBuilder
+
+	if input := req.RootDrive; input != nil {
+		builder = firecracker.DrivesBuilder{}.
+			WithRootDrive(input.PathOnHost, func(d *models.Drive) {
+				d.IsReadOnly = firecracker.Bool(input.IsReadOnly)
```
samuelkarp: If `input.IsReadOnly` is unset in the request, does this evaluate to true or false?

runtime/service.go
```diff
@@ -681,6 +668,33 @@ func (s *service) buildVMConfiguration(req *proto.CreateVMRequest) (*firecracker
 	return &cfg, nil
 }
 
+func (s *service) buildNonStubDrives(req *proto.CreateVMRequest) []models.Drive {
+	var builder firecracker.DrivesBuilder
+
+	if input := req.RootDrive; input != nil {
+		builder = firecracker.DrivesBuilder{}.
+			WithRootDrive(input.PathOnHost, func(d *models.Drive) {
```
samuelkarp: Would it be worth creating functions like `WithReadOnly(bool)`, `WithPartuuid(string)`, etc?

runtime/service_test.go
```diff
@@ -74,12 +74,13 @@ func TestBuildVMConfiguration(t *testing.T) {
 			expectedStubDriveCount: 1,
 		},
 		{
-			name: "Input",
+			name: "Only Request",
```
samuelkarp: Can we have a testcase where the `IsReadOnly` flag is unset?

runtime/service.go
kzys: Currently `false` which is inconsistent with the configuration file parsing.

While proto2 is having a way to [specify per-field default values](https://developers.google.com/protocol-buffers/docs/proto#optional), proto3 [doesn't have that anymore](https://developers.google.com/protocol-buffers/docs/proto3#default).

> Note that for scalar message fields, once a message is parsed there's no way of telling whether a field was explicitly set to the default value (for example whether a boolean was set to false) or just not set at all: you should bear this in mind when defining your message types. For example, don't have a boolean that switches on some behaviour when set to false if you don't want that behaviour to also happen by default.

runtime/service_test.go
kzys: Like the above comment, `IsReadOnly` is either true or false.

runtime/service.go
kzys: Maybe yes. Would it be better to upgrade Go SDK first then?

runtime/service.go
xibz: Perhaps we can change the name of the field, so it follows the inverse, like `IsWritable`? That way it defaults to false. I don't know if I like that, but I think it beats using a string or enumerables.

runtime/service.go
kzys: Opened as https://github.com/firecracker-microvm/firecracker-go-sdk/pull/140. Will update this branch after merging the change.

runtime/service.go
kzys: Sounds good. Will do.

kzys: Fixed on 91938abf83568d853544632e213b73c12d1a1b70

However, can't we just make IsReadOnly required?

kzys: Another option is making the root drive always read-only, since https://github.com/firecracker-microvm/firecracker-containerd/blob/master/docs/drive-mounts-proposal.md will separate root drives from non-root drives.

xibz: @kzys - You can't force the requirement though. How do you know if the user meant to set `IsReadOnly` to false versus not setting it at all?

> Another option is making the root drive always read-only, since https://github.com/firecracker-microvm/firecracker-containerd/blob/master/docs/drive-mounts-proposal.md will separate root drives from non-root drives.

I'd be more in favor of this if we want to get rid of the field. However, if someone ever wanted a writable root drive, then the `IsWritable` would be the implementation preferred. So you either can not do the work now, but we will probably need to do it in the future.

kzys: I meant, `IsReadOnly` required on Protobuf, instead of making the field optional. The change I'm making is small, but the way we workaround the mismatch between JSON vs. Protobuf seems hacky.

runtime/service.go
kzys: Fixed on 0e858869869d6f4bf1b366adbd232d2437c60f36

runtime/service.go
xibz: proto3 does not have the concept of optional versus required. proto2 does, but we don't use that.

kzys: Wow. Sorry. I haven't noticed that proto3 removed that aw well.

runtime/service.go
```diff
@@ -681,6 +668,33 @@ func (s *service) buildVMConfiguration(req *proto.CreateVMRequest) (*firecracker
...
+				firecracker.WithPartuuid(input.Partuuid),
+				firecracker.WithReadOnly(!input.IsWritable),
+				func(d *models.Drive) {
+					if limiter := input.RateLimiter; limiter != nil {
+						d.RateLimiter = rateLimiterFromProto(limiter)
+					}
+				})
+	} else {
+		builder = firecracker.DrivesBuilder{}.
+			WithRootDrive(s.config.RootDrive, func(d *models.Drive) {
```
xibz: `WithReadOnly` can be used here

runtime/service.go
```diff
@@ -681,6 +668,33 @@ func (s *service) buildVMConfiguration(req *proto.CreateVMRequest) (*firecracker
 	return &cfg, nil
 }
 
+func (s *service) buildNonStubDrives(req *proto.CreateVMRequest) []models.Drive {
+	var builder firecracker.DrivesBuilder
+
+	if input := req.RootDrive; input != nil {
+		builder = firecracker.DrivesBuilder{}.
```
xibz: We don't need to create a new builder since we could use the one above.

```go
builder = builder.WithRootDrive(...)
```

runtime/service.go
```diff
@@ -681,6 +668,33 @@ func (s *service) buildVMConfiguration(req *proto.CreateVMRequest) (*firecracker
...
+			WithRootDrive(input.PathOnHost,
+				firecracker.WithPartuuid(input.Partuuid),
+				firecracker.WithReadOnly(!input.IsWritable),
+				func(d *models.Drive) {
+					if limiter := input.RateLimiter; limiter != nil {
+						d.RateLimiter = rateLimiterFromProto(limiter)
+					}
+				})
+	} else {
+		builder = firecracker.DrivesBuilder{}.
```
xibz: Same here.

runtime/service.go
kzys: 6690f80e0383c6061290e93ac461489860565a20 renames IsReadOnly to IsWritable

runtime/service.go
kzys: https://github.com/firecracker-microvm/firecracker-containerd/pull/283/commits/96c7417c7fe06a4d8f2c25150c83010d9e6e400c upgrades the SDK and uses the newly-added helpers.

runtime/service_test.go
kzys: https://github.com/firecracker-microvm/firecracker-containerd/pull/283/commits/6690f80e0383c6061290e93ac461489860565a20 adds a new test case that doesn't specify IsWritable.

runtime/service.go
kzys: Ah, right. Still doing Java here. Addressed at https://github.com/firecracker-microvm/firecracker-containerd/pull/283/commits/96c7417c7fe06a4d8f2c25150c83010d9e6e400c

runtime/service.go
kzys: Addressed at https://github.com/firecracker-microvm/firecracker-containerd/pull/283/commits/96c7417c7fe06a4d8f2c25150c83010d9e6e400c

runtime/service.go
kzys: Good catch! Addressed at https://github.com/firecracker-microvm/firecracker-containerd/pull/283/commits/96c7417c7fe06a4d8f2c25150c83010d9e6e400c.

PR #284, opened 2019-10-11
Disable setting the shim as a subreaper.

sipsma:
Setting the shim process as a subreaper appears to enable race conditions
when using the Go stdlib "os/exec" library to start child processes, resulting
in occasional "wait: no child processes" errors when subprocesses are run. There
are similar reports online of these race conditions, though I have not found
any open issues with golang at this time.

The issue is very inconsistent. I first encountered it randomly when manually
testing with `ctr` on an instance, where it seemed to happen about half the time
(specifically being triggered when the shim spawned CNI plugin child processes
via the Firecracker Go SDK). I managed to see it again in an automated test
case, but only once. However, when making this change to disable setting the
shim as a subreaper, I no longer ever got the "wait: no child processes" error
from the manual ctr commands where I could reproduce it more consistently, so I
have some degree of confidence that it fixes the issue at this time.

It seems safe to disable this as it only makes sense for the shim to be a
subreaper if it also actually runs a reaper loop (where it calls Wait with -1,
or handles SIGCHLD signals), but our shim doesn't do that, so even if it
inherited a child process from a descendant, it would not actually wait on it.
There also doesn't seem to be any reason for us to add a reaper loop; child
processes can just be reparented to init (or the next subreaper), which can
handle waiting on them without any harm.

The likely reason containerd sets this to true by default is that a caller of
runc actually would want to be a subreaper (due to the fact that the final
container process created by runc is several descendants below the caller).
However, we do not invoke runc from the shim, only from the VM agent (where we
actually do run a reaper loop).

### DIFF
--- a/runtime/main.go
+++ b/runtime/main.go
@@ -35,5 +35,6 @@ func init() {
 func main() {
 	shim.Run(shimID, NewService, func(cfg *shim.Config) {
 		cfg.NoSetupLogger = true
+		cfg.NoSubreaper = true
 	})
 }


sipsma:
I just remembered that very soon we will actually be invoking runc from the shim once the jailer changes are merged... This change may still be okay but needs some more consideration, marking it as a draft until then.

sipsma:
Update in regards to whether this will affect runc, I think this change should still be fine. The jailer change calls ‘runc run ...’ as a child process, which will [set itself up as a subreaper unless specifically configured not to](https://github.com/opencontainers/runc/blob/e7a87dd240043ae17a4bb84f779e40e0ab1ee2b9/man/runc-run.8.md#options). I think being a subreaper matters if you use libcontainer directly, which we do not here. So this change should still be good to merge I believe.

sipsma:
Turns out the change to not set a subreaper made the error more rare, but I did actually manage to see the error again when running with `NoSubreaper` set. So this issue needs a lot more investigation. Closing this PR in favor of this issue: https://github.com/firecracker-microvm/firecracker-containerd/issues/285

PR #289, opened 2019-10-14
Fix root tests to be ran as isolated tests

xibz:
Prior to this fix the snapshotter root tests were not being run. This
change changes the root tests to be isolated tests and which allows them
to be ran in our CI pipeline.


xibz:
Closing as it looks like the root tests are actually ran. See in buildkite under `snapshotter "root" tests`

PR #293, opened 2019-10-17
Upgrade golangci-lint and enable errcheck

kzys:
*Issue #, if available:*

NA

*Description of changes:*

This branch upgrades golangci-lint and enable errcheck. Compared to gosec which we've been disabling. errcheck with [golangci-lint's default regexp](https://github.com/golangci/golangci-lint#command-line-options) results better signal-to-noise ratio.


internal/network_test_utils.go
```diff
@@ -115,7 +115,7 @@ func (l localNetworkServices) Serve(ctx context.Context) error {
 		for path, contents := range l.webpages {
 			webpage := contents
 			http.HandleFunc("/"+path, func(w http.ResponseWriter, r *http.Request) {
-				fmt.Fprintf(w, webpage)
+				fmt.Print(w, webpage)
```
sipsma: I don't think this is correct, [`Print` writes to stdout](https://golang.org/pkg/fmt/#Print) but we want to write to `w` here

internal/vm/task.go
```diff
@@ -123,6 +123,13 @@ func (m *taskManager) newProc(taskID, execID string) (*vmProc, error) {
 	return proc, nil
 }
 
+func (m *taskManager) cleanupProc(taskID, execID string) {
+	_, err := m.deleteProc(taskID, execID)
+	if err != nil {
+		m.logger.WithField("TaskID", taskID).WithField("ExecID", execID).WithError(err).Error("failed to delete proc")
```
sipsma: I think the log line here should be more like "failed to cleanup proc after create or exec failure". It, in theory, should never return an error because it's only called after a successful call to `newProc`, so when it happens in practice it's a bug in some invariant we were trying to hold and should be distinguished from a normal failure to delete it.

Also, `m.logger` already includes the `TaskID` and `ExecID` fields, so we don't need to add them again.

internal/vm/vsock.go
```diff
@@ -254,11 +254,16 @@ func tryAccept(logger *logrus.Entry, listener net.Listener, port uint32) (net.Co
...
 func tryConnRead(conn net.Conn, expectedRead string, timeout time.Duration) error {
-	conn.SetDeadline(time.Now().Add(timeout))
-	defer conn.SetDeadline(time.Time{})
+	err := conn.SetDeadline(time.Now().Add(timeout))
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = conn.SetDeadline(time.Time{})
+	}()
```
sipsma: It appears `SetDeadline` probably only fails if the socket itself is invalid (in which case we'll just get errors when trying to do the read/write), but it's still good to check these for extra safety. But in that case we should also return an error if the `SetDeadline` call in the defer fails. If that did fail but the socket was somehow still usable, then the caller would be getting a socket with an unexpected deadline.

This may be an okay use case for using a named return, but we'll also need to make sure we don't overwrite any other error already being returned.

internal/network_test_utils.go
samuelkarp: `fmt.Fprint` would be the right function

internal/vm/vsock.go
```diff
@@ -277,8 +282,13 @@ func tryConnRead(conn net.Conn, expectedRead string, timeout time.Duration) erro
...
 // connection is established fully.
 func tryConnWrite(conn net.Conn, expectedWrite string, timeout time.Duration) error {
-	conn.SetDeadline(time.Now().Add(timeout))
-	defer conn.SetDeadline(time.Time{})
+	err := conn.SetDeadline(time.Now().Add(timeout))
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = conn.SetDeadline(time.Time{})
```
xibz: I really don't like wrapping these in defer statements to just ignore the error anyways. Prefer `defer conn.SetDeadline(time.Time{})` instead.

firecracker-control/local.go
```diff
@@ -167,7 +167,10 @@ func (s *local) CreateVM(requestCtx context.Context, req *proto.CreateVMRequest)
 
 	defer func() {
 		if err != nil {
-			cmd.Process.Kill()
+			err = cmd.Process.Kill()
+			if err != nil {
```
xibz: nit:

```go
if err := cmd.Process.Kill(); err != nil {
}
```

This also doesn't stomp over the `err` variable in the function scope.

internal/vm/task_test.go
```diff
@@ -328,50 +332,52 @@ func TestTaskManager_CreateExecDeleteTask(t *testing.T) {
 	tm := NewTaskManager(shimCtx, logger.WithField("test", t.Name()))
 	ts := &mockTaskService{}
 
-	mockTask := newMockProc("fakeTask", "", mockIOConnector, mockIOConnector, mockIOConnector)
+	require := require.New(t)
```
mxpv: Out of curiosity. Why do we need this? As far as I see `require.New` just wraps `t`, so why not to call `require.X` directly?

internal/vm/vsock.go
xibz: Perhaps adding //nolint:errcheck above the defer may be preferable in cases where we are ignoring the error anyways.

internal/network_test_utils.go
kzys: Oh, right. Thanks. Why didn't it fail? Let me check.

internal/vm/task.go
kzys: For things that shouldn't happen, can we just `panic`?

internal/vm/task_test.go
kzys: Passing testing.T to newMockProc() is another way. I'm fine with both ways, while having a contextual require is slightly shorter though.

internal/vm/vsock.go
kzys: I somehow thought that inside defer is too late to return `err`. Will update.

internal/network_test_utils.go
sipsma: The CNI test failed in buildkite due to this

internal/vm/task.go
sipsma: We can, but I really prefer that we only panic if there is absolutely no other choice. While this shouldn't happen in theory, obviously there could be a bug now or in the future that causes it to arise. In that scenario, we absolutely should log it but I don't think this error specifically means we need to immediately crash everything. We can continue to operate.

internal/vm/vsock.go
kzys: Let me return the err instead of ignoring.

samuelkarp: > Perhaps adding `//nolint:errcheck` above the defer may be preferable in cases where we are ignoring the error anyways.

Unless a linter provides a ton of value (and I'm not convinced that `errcheck` does) my general preference is to disable a faulty linter rather than trying to hold its hand.  There are a number of valid reasons to ignore a returned error and I think this type of linting is too aggressive.

kzys: If so, I will exclude errcheck and just upgrade golangci-lint.

internal/network_test_utils.go
```diff
@@ -115,7 +116,7 @@ func (l localNetworkServices) Serve(ctx context.Context) error {
 		for path, contents := range l.webpages {
 			webpage := contents
 			http.HandleFunc("/"+path, func(w http.ResponseWriter, r *http.Request) {
-				fmt.Fprintf(w, webpage)
+				io.WriteString(w, webpage)
```
xibz: out of curiosity, why change this?

kzys: golangci-lint is not happy about using `*f` function for just writing a string.

PR #295, opened 2019-10-17
Makes use of ioutil.TempDir in stub drive tests

xibz:
This commit changes to use ioutil.TempDir as opposed to creating a
static directory when running the TestStubDriveHandler.


runtime/drive_handler_test.go
```diff
@@ -32,8 +32,7 @@ import (
 )
 
 func TestStubDriveHandler(t *testing.T) {
-	const tempPath = "test"
-	err := os.Mkdir(tempPath, os.ModePerm)
+	tempPath, err := ioutil.TempDir("./", "TestStubDriveHandler")
 	assert.NoError(t, err)
```
sipsma: nit: this should probably be `require.NoError` instead of `assert` because the rest of the test is destined to fail if the temp dir fails to create.

xibz: Fixed

Issue #296, opened 2019-10-17
Support RateLimiters and ReadOnly for DriveMounts


sipsma:
The first iteration of support for DriveMounts leaves out RateLimiter and IsReadOnly settings for the drives due to the fact that doing so requires significant refactoring of the stubdrive code (as RateLimiter and IsReadOnly settings cannot be updated via a Patch call).

Right now, users can still specify that a filesystem be bind-mounted as read-only into their container, but they don't have any options for the rate-limiter currently.

PR #297, opened 2019-10-17
Support DriveMount API in CreateVM.

sipsma:
This implements the proposal found in docs/drive-mounts-proposal.md, with the
exception for supporting RateLimiters and IsReadOnly in the DriveMount objects,
due to the need for further refactoring of the internal stub drive code (will
be followed up in #296).

This change will require a fair bit of non-trivial rebasing with the jailer PR, so I'd like to let that PR get merged first and then rebase this one on top if at all possible.


agent/drive_handler.go
```diff
@@ -146,3 +164,87 @@ func isStubDrive(d drive) bool {
...
+
+	return nil, errors.Errorf("exhausted retries mounting drive from %q to %q",
+		drive.Path(), req.DestinationPath)
+}
+
+// evalAnySymlinks is similar to filepath.EvalSymlinks, except it will not return an error if part of the
+// provided path does not exist. It will evaluate symlinks present in the path up to a component that doesn't
+// exist, at which point it will just append the rest of the provided path to what has been resolved so far.
+func evalAnySymlinks(path string) (string, error) {
+	curPath := "/"
```
xibz: Can we add a comment here saying that we validate that the curPath must be at `/`?

agent/drive_handler.go
```diff
@@ -146,3 +164,87 @@ func isStubDrive(d drive) bool {
...
+
+	// Do a basic check that we won't be mounting over any important system directories
+	resolvedDest, err := evalAnySymlinks(req.DestinationPath)
+	if err != nil {
+		return nil, errors.Wrapf(err,
+			"failed to evaluate any symlinks in drive mount destination %q", req.DestinationPath)
+	}
+
+	for _, systemDir := range bannedSystemDirs {
+		if strings.HasPrefix(filepath.Clean(resolvedDest), filepath.Clean(systemDir)) {
```
kzys: - You don't have to call filepath.Clean() on systemDir, while it is harmless
- Wouldn't it match a path that technically contain systemDir such as /system?

agent/service.go
```diff
@@ -113,19 +115,42 @@ func unmarshalExtraData(marshalled *types.Any) (*proto.ExtraData, error) {
...
 	execID := "" // the exec ID of the initial process in a task is an empty string by containerd convention
 
 	logger := log.G(requestCtx).WithField("TaskID", taskID).WithField("ExecID", execID)
 	logger.Info("create")
 
+	defer func() {
+		if err != nil {
+			cleanupErr := ts.doCleanup(taskID, execID)
+			if cleanupErr != nil {
+				logger.WithError(err).Error("failed to cleanup task")
```
xibz: Do we not want to log the cleanup error as well?

agent/drive_handler.go
```diff
@@ -146,3 +164,87 @@ func isStubDrive(d drive) bool {
 
 	return internal.IsStubDrive(f)
 }
+
+func (dh driveHandler) MountDrive(ctx context.Context, req *drivemount.MountDriveRequest) (*empty.Empty, error) {
+	logger := log.G(ctx).WithField("MountDriveRequest", req.String())
+	logger.Debug()
+
+	driveID := strings.TrimSpace(req.DriveID)
```
kzys: Can't we just use req.DriveID as is?

agent/service.go
```diff
@@ -375,6 +386,15 @@ func (ts *TaskService) Exec(requestCtx context.Context, req *taskAPI.ExecProcess
 	logger := log.G(requestCtx).WithField("TaskID", taskID).WithField("ExecID", execID)
 	logger.Debug("exec")
 
+	defer func() {
+		if err != nil {
+			cleanupErr := ts.doCleanup(taskID, execID)
+			if cleanupErr != nil {
+				logger.WithError(err).Error("failed to cleanup task")
```
xibz: Same comment here regarding not logging `cleanupErr`

agent/drive_handler.go
```diff
@@ -146,3 +164,87 @@ func isStubDrive(d drive) bool {
 
 	return internal.IsStubDrive(f)
 }
+
+func (dh driveHandler) MountDrive(ctx context.Context, req *drivemount.MountDriveRequest) (*empty.Empty, error) {
+	logger := log.G(ctx).WithField("MountDriveRequest", req.String())
```
mxpv: `MountDriveRequest` has lots of fields. I dont think we want to repeat them in every log message below. Would it make sense to log those fields once in `Debugf`?

agent/drive_handler.go
```diff
@@ -146,3 +164,87 @@ func isStubDrive(d drive) bool {
...
+	logger.Debug()
+
+	driveID := strings.TrimSpace(req.DriveID)
+	drive, ok := dh.GetDrive(driveID)
+	if !ok {
+		return nil, fmt.Errorf("Drive %q could not be found", driveID)
```
mxpv: nit: lower case

internal/fsutil.go
```diff
@@ -0,0 +1,113 @@
...
+// into MountInfo objects holding the parsed data about each mount.
+func ParseProcMountLines(lines ...string) ([]MountInfo, error) {
+	mountInfos := []MountInfo{}
+	for _, line := range lines {
+		if line == "" {
+			continue
+		}
+
+		// see man 5 fstab for the format of /proc/mounts
+		var source string
```
xibz: would prefer
```go
var (
    source string
    dest string
    ...
)
```

agent/drive_handler.go
sipsma: Good point on it matching `/system`. Will fix

agent/drive_handler.go
sipsma: Yeah I think so, this was just copied from the previous implementation, but unless I'm missing something `TrimSpace` shouldn't be needed, will update

runtime/service.go
```diff
@@ -500,10 +501,35 @@ func (s *service) createVM(requestCtx context.Context, request *proto.CreateVMRe
...
 
+func (s *service) mountDrives(requestCtx context.Context, driveMounts []*proto.FirecrackerDriveMount) error {
+	for _, driveMount := range driveMounts {
+		driveID, err := s.stubDriveHandler.PatchStubDrive(requestCtx, s.machine, driveMount.HostPath)
+		if err != nil {
+			return errors.Wrapf(err, "failed to patch drive mount stub")
+		}
+
+		_, err = s.driveMountClient.MountDrive(requestCtx, &drivemount.MountDriveRequest{
+			DriveID:         *driveID,
```
xibz: Should use `firecracker.StringValue` here to prevent panics

agent/service.go
mxpv: can we include `cleanupErr` as well ? like `failed to cleanup task: %v`

agent/drive_handler.go
```diff
@@ -146,3 +164,87 @@ func isStubDrive(d drive) bool {
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create drive mount destination %q", req.DestinationPath)
+	}
+
+	const (
+		maxRetries = 100
+		retryDelay = 10 * time.Millisecond
+	)
+
+	for i := 0; i < maxRetries; i++ {
```
mxpv: In which cases drive might not be mountable?

agent/drive_handler.go
sipsma: Yeah good point, will do

agent/drive_handler.go
sipsma: This can happen when we've patched a stub drive but the OS has not yet realized a change was made, which is rare but did happen in practice occasionally. We needed these retries in the previous implementation too, I just migrated the code as part of this change.

agent/service.go
sipsma: We do, this was an oversight, will fix

agent/service.go
sipsma: Yep this was an oversight, will fix

runtime/service.go
```diff
@@ -500,10 +501,35 @@ func (s *service) createVM(requestCtx context.Context, request *proto.CreateVMRe
...
+		return err
+	}
+
 	s.logger.Info("successfully started the VM")
+	return nil
+}
 
+func (s *service) mountDrives(requestCtx context.Context, driveMounts []*proto.FirecrackerDriveMount) error {
+	for _, driveMount := range driveMounts {
+		driveID, err := s.stubDriveHandler.PatchStubDrive(requestCtx, s.machine, driveMount.HostPath)
```
xibz: `PatchStubDrive` can return a `string` now for ease of use.

sipsma: Done

agent/drive_handler.go
mxpv: nit: I think this is useful context and it worth to leave comments explaining the motivation behind retries.

sipsma: Yeah good call, added a comment w/ link to original issue

sipsma:
Have the approvals now, but will be waiting until the Jailer PR is merged and do a rebase on top of that before merging this one. Any more comments in the meantime are welcome though of course.

agent/drive_handler.go
nmeyerhans: Yeah, not only do we validate it, but we override it if it isn't.

IMO this function should be given a name that more clearly indicates what it does. Converting a relative path to an absolute path doesn't have anything to do with symlinks.

sipsma: The point of this function is not to convert a relative path into an absolute path, it's to see if any part of the provided path already exists and, if so, evaluate any symlinks in the part of the path that exists. I needed to add it because the standard `EvalSymlinks` function in the standard lib will always return an error if any part of the provided path doesn't exist yet.

I.e. if you are given `/foo/bar/baz/` and `/foo` is a symlink to `/other`, but `/other/bar/baz` doesn't exist yet, then `EvalSymlinks` will return an error. `evalAnySymlinks`, on the other hand, will return `/other/bar/baz`.

I updated the comment on the function to also state that it's assumed `path` is an absolute path, which is validation we do earlier. Given this is just a private, one-off util function only used in this package, I'm okay with that, but let me know if there's disagreement.

sipsma:
Rebased on the Jailer PR. It ended up being straightforward to resolve conflicts, nothing substantive had to change in the drive mount logic or the jailer logic.

sipsma:
I just realized that because the jailer config is not enabled by default there currently aren't integ-tests that use both drive mounts and the jailer, I'm gonna update with that before merging too. Looking at the jailer code again I think this will require some updates to the "ExposeDeviceToJail" method in order to handle exposing flat files in addition to exposing block devices.

PR #298, opened 2019-10-21
Test that the size of a root drive is capped 

kzys:
*Issue #, if available:*

NA

*Description of changes:*

This change adds integration tests that verify in-container workloads
are restricted correctly.

### DIFF
--- a/runtime/limits_integ_test.go
+++ b/runtime/limits_integ_test.go
@@ -1,0 +1,66 @@
+// Copyright 2018-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+package main
+
+import (
+	"context"
+	"testing"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/namespaces"
+	"github.com/containerd/containerd/oci"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDiskLimit_Isolated(t *testing.T) {
+	prepareIntegTest(t)
+
+	assert := assert.New(t)
+	require := require.New(t)
+
+	ctx := namespaces.WithNamespace(context.Background(), "default")
+
+	client, err := containerd.New(containerdSockPath, containerd.WithDefaultRuntime(firecrackerRuntime))
+	require.NoError(err, "unable to create client to containerd service at %s, is containerd running?", containerdSockPath)
+	defer client.Close()
+
+	image, err := alpineImage(ctx, client, defaultSnapshotterName())
+	require.NoError(err, "failed to get alpine image")
+
+	// Right now, both naive snapshotter and devmapper snapshotter are configured to have 1024MB image size.
+	// The former is hard-coded since the snapshotter is not for production. The latter is configured in tools/docker/entrypoint.sh.
+	sh := containerd.WithNewSpec(
+		oci.WithProcessArgs("dd", "if=/dev/zero", "of=/tmp/fill", "bs=1M", "count=2000"),
+		oci.WithDefaultPathEnv,
+	)
+
+	container, err := client.NewContainer(ctx,
+		"container",
+		containerd.WithSnapshotter(defaultSnapshotterName()),
+		containerd.WithNewSnapshot("snapshot", image),
+		sh,
+	)
+	defer func() {
+		err = container.Delete(ctx, containerd.WithSnapshotCleanup)
+		require.NoError(err, "failed to delete a container")
+	}()
+
+	result, err := runTask(ctx, container)
+	require.NoError(err, "failed to create a container")
+
+	assert.Equal(uint32(1), result.exitCode, "writing 2GB must fail")
+	assert.Equal(`952+0 records in
+951+0 records out
+`, result.stderr, "but it must be able to write ~1024MB")
+}
--- a/runtime/integ_test.go
+++ b/runtime/integ_test.go
@@ -13,12 +13,17 @@
 package main
 
 import (
+	"bytes"
+	"context"
 	"encoding/json"
 	"os"
 	"strings"
 	"testing"
 
+	"github.com/containerd/containerd"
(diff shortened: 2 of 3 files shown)


runtime/limits_integ_test.go
```diff
@@ -0,0 +1,174 @@
...
+	result, err := runTask(ctx, container)
+	require.NoError(err, "failed to create a container")
+
+	assert.Equal(uint32(1), result.exitCode, "exitCode")
+
+	assert.Equal(`Filesystem                Size      Used Available Use% Mounted on
+/dev/vdb                975.9M      8.4M    900.3M   1% /
+`, result.stdout, "stdout")
+
+	// TODO: Doesn't it have only 900MB? The written records are higher than expected.
```
samuelkarp: Where did the 975.9M disk come from?  It doesn't look like that's defined as part of the test, but rather an implementation detail of the naive snapshotter.  Is that correct?

runtime/limits_integ_test.go
```diff
@@ -0,0 +1,174 @@
...
+		err = container.Delete(ctx, containerd.WithSnapshotCleanup)
+		require.NoError(err, "failed to delete a container")
+	}()
+
+	result, err := runTask(ctx, container)
+	require.NoError(err, "failed to create a container")
+
+	assert.Equal(uint32(1), result.exitCode)
+	assert.Equal("", result.stdout, "stdout")
+	assert.Equal("sh: out of memory\n", result.stderr, "stderr")
```
samuelkarp: I don't see a memory limit specified anywhere.  Where are you getting a limit from?

kzys: This limit is coming from Firecracker's default memory limit. I would have an explicit Create VM Request.

runtime/limits_integ_test.go
kzys: Yes. Coming from https://github.com/firecracker-microvm/firecracker-containerd/blob/6218f7e43acfa439b5ac299677aa7a1458796470/snapshotter/naive/naive.go#L38, I think.

Let me check whether we have a way to have an explicit limit or not.

runtime/limits_integ_test.go
samuelkarp: Putting it in the CreateVM request would help a lot.  I think another useful thing would be to assert on the amount of memory the `firecracker` process is using (as a proxy for the amount of memory used by the VM).  There would need to be some amount of buffer for the overhead of Firecracker itself, but that way we'd be testing the VMM and our configuration of it rather than the behavior of the Linux kernel inside the VM.

sipsma: I agree with Sam but would go further and say that all firecracker-containerd code does is translate CreateVM API parameters into Firecracker API calls, so I feel like an e2e test of just *our* code would make a CreateVM call with a given memory limit and then open a client directly with the VMM and verify it reports it was actually configured with the expected memory limit. Verifying the memory limit was passed through as expected is what we actually care about testing, not that the Firecracker implementation is also honoring that memory limit (presumably they already have tests for that in their own repo).

Now, if it turns out that for some reason it's actually harder to ask the VMM directly for what memory limit it was configured with compared to asserting on the VMM process's memory usage, then I'd be fine with asserting on the VMM process's memory usage. I just highly suspect that the issue Sam brought up with needing to deal with the VMM's own memory usage introduces enough complication to that approach that it's probably easier to write a non-flaky test that just talks to the VMM API endpoint to make assertions on the outcome of a CreateVM call.

kzys: Yeah. I agree that this test is doing what Firecracker supposed to do. Firecracker has some tests like https://github.com/firecracker-microvm/firecracker/commit/36583c62714c5a5d5637ea2df0cce6624d0d7b46 to confirm the behavior.

Checking only the data transform layer would not cover handling of stdout, stderr and/or exit code, but those are technically covered by the disk usage test I have in this PR.

runtime/limits_integ_test.go
kzys: Yes. With devmapper, the size is much much smaller.

```
--- FAIL: TestDiskLimit_Isolated (6.98s)
    limits_integ_test.go:118:
                Error Trace:    limits_integ_test.go:118
                Error:          Not equal:
                                expected: "Filesystem                Size      Used Available Use% Mounted on\n/dev/vdb                975.9M      8.4M    900.3M   1% /\n"
                                actual  : "Filesystem                Size      Used Available Use% Mounted on\n/dev/vdb                120.0M      7.0M    104.0M   6% /\n"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,3 +1,3 @@
                                 Filesystem                Size      Used Available Use% Mounted on
                                -/dev/vdb                975.9M      8.4M    900.3M   1% /
                                +/dev/vdb                120.0M      7.0M    104.0M   6% /

                Test:           TestDiskLimit_Isolated
                Messages:       stdout
    limits_integ_test.go:123:
                Error Trace:    limits_integ_test.go:123
                Error:          Not equal:
                                expected: "952+0 records in\n951+0 records out\n"
                                actual  : "111+0 records in\n110+0 records out\n"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,3 +1,3 @@
                                -952+0 records in
                                -951+0 records out
                                +111+0 records in
                                +110+0 records out

                Test:           TestDiskLimit_Isolated
                Messages:       stderr
FAIL
```

kzys: Because I'm specifying https://github.com/firecracker-microvm/firecracker-containerd/blob/38742eeb4fb5d549b31727a57b8c79e88b204b59/tools/docker/entrypoint.sh#L29 🤦‍♂

kzys: Commented about where the size is coming from. We can make naive snapshotter more configurable, but the snapshotter is not for production. So, not so sure the value it brings vs. complexity.

runtime/limits_integ_test.go
kzys: I have removed that from this PR. Still not so sure what would be the best way to confirm the behavior.

We may also set the limit through https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#memory.

runtime/limits_integ_test.go
```diff
@@ -0,0 +1,115 @@
...
+			stdout:   stdout.String(),
+			stderr:   stderr.String(),
+			exitCode: exitStatus.ExitCode(),
+		}, nil
+	case <-ctx.Done():
+		return nil, errors.New("context cancelled")
+	}
+}
+
+func TestDiskLimit_Isolated(t *testing.T) {
```
samuelkarp: nit: I'd move `TestDiskLimit_Isolated` up to the top of the file since this is actually the entry point and the other functions are helpers.

kzys: Okay. Will do.

kzys: Let me move runTask to integ_test.go since it can be used from other tests.

PR #299, opened 2019-10-21
Add pre-exit hook to cleanup after job

xibz:
This commit removes the cleanup step from the pipeline.yml file and
instead adds a hook that allows for cleanup to always occur even in the
event of a cancel.

The `pipeline.yml` cleanup only occurs if you let the whole job finish, which is not what we want.

### DIFF
--- a/.buildkite/hooks/pre-exit
+++ b/.buildkite/hooks/pre-exit
@@ -1,0 +1,4 @@
+#!/bin/bash
+set -eu
+
+make -C "$BUILDKITE_BUILD_CHECKOUT_PATH" clean
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -86,11 +86,3 @@ steps:
     command: 'make -C examples integ-test'
     concurrency: 1
     concurrency_group: 'loop-device test'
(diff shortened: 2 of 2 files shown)


.buildkite/hooks/pre-exit
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -eu
+CODEPATH=$BUILDKITE_BUILD_CHECKOUT_PATH
+cd $CODEPATH && make clean
```
nmeyerhans: Why introduce this second variable? Why not just cd to `$BUILDKITE_BUILD_CHECKOUT_PATH`?

Variables pointing to path names should always be enclosed in double quotes in shell scripts, especially when the path name comes from somewhere else. Consider what happens in the event that `$BUILDKITE_BUILD_CHECKOUT_PATH` contains spaces or other special characters.

A better way to write this would be `make -C "$BUILDKITE_BUILD_CHECKOUT_PATH" clean`

xibz: Fixed!

The reason for the second variable was clarity on what the path was, but I removed it

Issue #190, opened 2019-05-21
Loop device cleanup in CI


xibz:
In reference to #184, our CI system will only run one job at a time, which doesn't utilize the full capacity of the CI system. The reason this was done as loop devices kept getting created this would slow down the system and in addition parallel tests would cause each other to time out when trying to utilize the same loop device.

sipsma:
We run in parallel now thanks to #300 , closing as I don't think there's further action for this issue

PR #300, opened 2019-10-21
Run devmapper snapshotter on BuildKite

kzys:

*Issue #, if available:*

This might address #190 by not using "naive" device mapper.

*Description of changes:*

Using Device Mapper instead of "naive" one to

- Be more like production
- Remove the concurrency limit to make testing faster

But I'm not so sure what would be the best way to integrate this into other integration tests...


tools/docker/Dockerfile
```diff
@@ -168,13 +168,31 @@ COPY tools/docker/naive-snapshotter/entrypoint.sh /entrypoint
 ENTRYPOINT ["/entrypoint"]
 CMD ["exec /bin/bash"]
 
+FROM firecracker-containerd-test as firecracker-containerd-devmapper-integ-test
```
sipsma: I think I'd prefer we take a different approach than adding a new test image for this. We could update to not start or configure containerd as part of the test image at all and instead have each test case configure+start containerd itself (as part of the `prepareTest` util you recently added).

That approach will give each test case a lot more flexibility and keep our test-image infra much simpler (we can continue just using one single test image instead of managing multiples). 

I think there is more effort required upfront in order to do this, but I suspect it's worth it for longer-term maintenance. However, if it turns out for some reason to be an enormous effort way out of the scope of just using the devmapper, then I'm fine with it just being a backlog issue.

kzys:
The last commit didn't solve the issue, but `sudo lvremove -f /dev/mapper/fcci--vg-fcci--thin` on the CI host did.

tools/docker/Dockerfile
kzys: While I agree that we don't have to have an additional Docker image for running the tests, I'm unsure that starting containerd itself from prepareTest is a good way.

Currently integration tests are act as clients, but starting containerd from there makes the tests as servers.

Let me take a look how hard it would be.

sipsma: I was imagining the tests start containerd by just shelling out, so the tests themselves would not be servers. They would start a server and then use a client to interact with/test that server. But yeah, let's see how much effort it would be.

examples/prepare-config-toml.sh
```diff
@@ -0,0 +1,28 @@
...
+snapshotter="$1"
+pool_name="$2"
+
+case "$snapshotter" in
+    'naive')
+        cat > "$config_toml" <<EOF
+[proxy_plugins]
+  [proxy_plugins.firecracker-naive]
+    type = "snapshot"
+    address = "/var/run/firecracker-containerd/naive-snapshotter.sock"
```
mxpv: It might be somewhat easier to use config imports here.
https://github.com/containerd/containerd/pull/3310#issuecomment-504182696
https://github.com/containerd/containerd/pull/3574

tools/docker/devmapper-snapshotter/config.toml
```diff
@@ -0,0 +1,6 @@
+disabled_plugins = ["cri"]
```
mxpv: nit: I think we don't need this line, I'm not sure why we cary this everywhere.

.buildkite/pipeline.yml
```diff
@@ -87,6 +87,18 @@ steps:
...
+    env:
+      DOCKER_IMAGE_TAG: "$BUILDKITE_BUILD_NUMBER"
+      EXTRAGOARGS: "-v -count=1"
+    artifact_paths:
+      - "examples/logs/*"
+    command:
+      - sudo lvcreate --type thin-pool --poolmetadatasize 16GiB --extents '50%FREE' -n build$BUILDKITE_BUILD_NUMBER fcci-vg
+      - make -C examples integ-test TEST_SS=devmapper TEST_POOL=build$BUILDKITE_BUILD_NUMBER
+      - sudo dmsetup remove_all
+      - sudo lvremove -f /dev/mapper/fcci--vg-build$BUILDKITE_BUILD_NUMBER
```
mxpv: Should we move these 2 commands to `cleanup` step below? Will they run if make fails ?

tools/docker/devmapper-snapshotter/config.toml
kzys: Thanks. Will remove.

examples/prepare-config-toml.sh
kzys: Nice! Then I think we can have imports = ["/etc/containerd/snapshotter/*.toml"] and bind-mount the files.

.buildkite/pipeline.yml
kzys: Maybe  .buildkite/hooks/pre-exit?

mxpv: Should do the job too.

tools/docker/Dockerfile
kzys: Okay. Renaming the Docker image to be snapshotter-agonostic is in #301.

.buildkite/pipeline.yml
kzys: BTW 'sudo dmsetup remove_all' would remove devices that are in-use by other tests. I may need to find a better alternative.

examples/prepare-config-toml.sh
kzys: Seems import needs containerd upgrade #302

.buildkite/pipeline.yml
mxpv: As an option: devices are prefixed with pool name ([1](https://github.com/containerd/containerd/blob/master/snapshots/devmapper/snapshotter.go#L420)), so you can try removing all devices that belong to a pool. Something like
 `dmsetup list | grep pool-prefix- |  dmsetup remove`

tools/docker/devmapper-snapshotter/config.toml
kzys: Removed the line from tools/docker/config.toml, which is shared by all integration tests.

examples/prepare-config-toml.sh
kzys: import is now working!

.buildkite/pipeline.yml
kzys: Updated the cleanup process. However .buildkite/hooks/pre-exit is running at every step which doesn't work well for this usecase, since I'd like to reuse the thin-pool while executing the steps.

tools/docker/Dockerfile
kzys: Unified the Docker image

kzys:
Right now, BuildKite has "example tests" and "example tests on devmapper" to run two snapshotters separately.

I'm unsure whether we want to remove "naive" snapshotter entirely, or duplicate all tests with two snapshotters. Any thoughts?

sipsma:
> Right now, BuildKite has "example tests" and "example tests on devmapper" to run two snapshotters separately.
> 
> I'm unsure whether we want to remove "naive" snapshotter entirely, or duplicate all tests with two snapshotters. Any thoughts?

I'd prefer that the tests in the `runtime` package just run with the devmapper going forward because they will be much, much faster and will be testing what we actually want users to run with in production.

However, when we've discussed this in the past, it was brought up that removing the naive snapshotter can't be done until we also update the getting-started/quick-start guides to setup the pre-reqs for the devmapper snapshotter (which starts to expand the scope of the changes here pretty significantly).

What if for now we actually leave the example test as using the `naive` snapshotter so we still have some test coverage for `naive` but update all the `runtime` tests to use `devmapper`?

kzys:
The changes are ready for review. Also the tests are a bit faster than master ([7m41s](https://buildkite.com/firecracker-microvm/firecracker-containerd/builds/969) vs [9m56s](https://buildkite.com/firecracker-microvm/firecracker-containerd/builds/963)), mainly because now we can run more tests in parallel. Only one datapoint though :)

examples/taskworkflow.go
```diff
@@ -115,11 +122,11 @@ func taskWorkflow(containerCIDR string, gateway string) (err error) {
 		}
 	}()
 
-	log.Printf("Successfully pulled %s image\n", image.Name())
+	log.Printf("Successfully pulled %s image from %s\n", image.Name(), snapshotter)
```
samuelkarp: s/from/with/

kzys: Thanks. Fixed on https://github.com/firecracker-microvm/firecracker-containerd/pull/300/commits/cf050b10ec05105122509892206000107247d0fe and since the commit is the first commit in this PR, I have to force-push...

tools/thinpool.sh
```diff
@@ -0,0 +1,46 @@
+#! /bin/sh
```
nmeyerhans: This file should have a copyright header.

tools/thinpool.sh
```diff
@@ -0,0 +1,46 @@
...
+
+case "$subcommand" in
+    'create')
+        pool_create
+        ;;
+    'remove')
+        pool_remove
+        ;;
+    'reset')
+        if [ -e ${dm_device} ]; then
```
nmeyerhans: Please add double-quotes around `${dm_device}`

tools/thinpool.sh
kzys: Oh right. Will do.

tools/thinpool.sh
kzys: Will do.

PR #305, opened 2019-10-25
Adding netns to jailer 

xibz:
This adds netns specification per create vm request. In the event that
no netns was found in the runc config, we will then make sure of what
was passed into the create vm request


tools/docker/Dockerfile
```diff
@@ -160,7 +160,6 @@ RUN containerd 2>/dev/null & \
 	ctr content fetch docker.io/library/alpine:3.10.1 >/dev/null && \
 	ctr content fetch docker.io/mlabbe/iperf3:3.6-r0 >/dev/null
 
-COPY tools/docker/naive-snapshotter/entrypoint.sh /entrypoint
```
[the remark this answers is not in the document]
xibz: Thanks! Fixed

runtime/firecracker-runc-config.json.example
```diff
@@ -96,9 +96,6 @@
             {
                 "type": "pid"
             },
-            {
-                "type": "network"
-            },
```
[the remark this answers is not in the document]
xibz: Yea, I was questioning which is preferable. The current functionality is that if there is no netns in the runc config, we will insert the one that was given to us in the `CreateVMRequest`. But after your explanation, I agree with going with keeping the netns in the config, and always overwrite if there is a netns specified in the request.

xibz:
> 
    Is it possible to have firecracker-containerd actually set the netns field in the go sdk's machine.JailerConfig when it was set in the CreateVM's JailerConfig?
    If the above isn't possible, could we change the Go SDK to move the existing machine.JailerConfig.NetNS field to just be directly in the machine object? Then we can conditionally set that field in firecracker-containerd w/out creating a whole machine.JailerConfig object.

I think what I can do is modify the SDK code to create a netns handler by passing in the netns path. This will use a default netns in the default handler, then during jailing it'll create and swap handlers with its netns. Then in the firecracker-containerd code I just need to inject the correct handlers.

xibz:
Waiting on https://github.com/firecracker-microvm/firecracker-go-sdk/pull/155 to be merged

runtime/firecracker-runc-config.json.example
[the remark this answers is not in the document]
sipsma: Can we make this update to set a network namespace in the config file by default?

tools/docker/Dockerfile
```diff
@@ -151,7 +151,7 @@ COPY _submodules/firecracker/target/$FIRECRACKER_TARGET/release/firecracker /usr
 COPY _submodules/firecracker/target/$FIRECRACKER_TARGET/release/jailer /usr/local/bin/
 COPY _submodules/runc/runc /usr/local/bin
 COPY tools/image-builder/rootfs.img /var/lib/firecracker-containerd/runtime/default-rootfs.img
-COPY runtime/firecracker-runc-config.json.example /etc/containerd/firecracker-runc-config.json
+COPY tools/docker//firecracker-runc-config.json /etc/containerd/firecracker-runc-config.json
```
sipsma: did we remove the other file at `runtime/firecracker-runc-config.json.example`?  I think we should only have one for simplicity

runtime/firecracker-runc-config.json.example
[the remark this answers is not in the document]
xibz: Done!

tools/docker/Dockerfile
xibz: Done

go.sum
```diff
@@ -61,6 +61,8 @@ github.com/docker/go-units v0.3.3 h1:Xk8S3Xj5sLGlG5g67hJmYMmUgXv5N4PhkjJHHqrwnTk
 github.com/docker/go-units v0.3.3/go.mod h1:fgPhTUdO+D/Jk86RDLlptpiXQzgHJF7gydDDbaIK4Dk=
 github.com/firecracker-microvm/firecracker-go-sdk v0.17.1-0.20191029213755-dbf9a1e05f09 h1:JDfRpK+V2J1Es+Xm6aMJjCqvA4xv1kuWnJfeSopyDwo=
 github.com/firecracker-microvm/firecracker-go-sdk v0.17.1-0.20191029213755-dbf9a1e05f09/go.mod h1:tVXziw7GjioCKVjI5/agymYxUaqJM6q7cp9e6kwjo8Q=
```
kzys: `go mod tidy` may remove the lines.

runtime/runc_jailer.go
```diff
@@ -353,36 +367,68 @@ func (j runcJailer) jailerCommand(containerName string, isDebug bool) *exec.Cmd
...
+		return err
+	}
+
+	spec := *configSpec
 	spec.Process.User.UID = j.uid
 	spec.Process.User.GID = j.gid
 
-	configBytes, err = json.Marshal(&spec)
+	// remove the network namespace if there exists a CNI
+	if hasCNI(machineConfig.NetworkInterfaces) {
```
sipsma: Another thought on all this, `overwriteConfig` runs as part of a handler, which means that it runs after [`NewMachine`](https://github.com/firecracker-microvm/firecracker-containerd/blob/5f3e0c74a7c707ac7e0cd0a8f1466abff75d47de/runtime/service.go#L496) is called, right?

If so, `NewMachine` is where the SDK [sets the network namespace if there is a CNI interface](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/eec83f3dcb7b4eb4eeb1e2c5f975715a25f79e96/machine.go#L304-L306). So then because it's in a handler (with signature `func(ctx context.Context, m *firecracker.Machine)`), we can access that netns value at `m.Cfg.NetNS`, right? If so, then `overwriteConfig` shouldn't actually need to run `hasCNI`, it should be able to just read `m.Cfg.NetNS` and fill in the runc config if a netns is specified there.

I still think the better long-term cleanup of this is probably to move all this jailing to the go-sdk, but if the above works then it should simplify a little bit in the short term at least.

go.sum
xibz: Good catch! Fixed

runtime/runc_jailer.go
```diff
@@ -353,36 +367,68 @@ func (j runcJailer) jailerCommand(containerName string, isDebug bool) *exec.Cmd
...
-// TODO: Add netns
-func (j runcJailer) overwriteConfig(cfg *Config, socketPath, configPath string) error {
-	spec := specs.Spec{}
-	configBytes, err := ioutil.ReadFile(configPath)
-	if err != nil {
-		return err
-	}
+func (j *runcJailer) overwriteConfig(cfg *Config, machineConfig *firecracker.Config, socketPath, configPath string) error {
+	var err error
+	j.once.Do(func() {
```
kzys: What is the purpose of the once here? The all operations inside the closure look have no side-effects.

xibz: caches the config so multiple reads on the file do not occur

runtime/runc_jailer.go
xibz: Fixed!

Yea, I agree a better solution would be to have a jailer interface in the SDK. However, the logic would probably still live here, but would instead just satisfy the interface and pass that into the SDK's config.

runtime/runc_jailer.go
```diff
@@ -136,7 +150,7 @@ func (j *runcJailer) BuildJailedRootHandler(cfg *Config, socketPath *string, vmI
 			}
 
 			j.logger.Debug("Overwritting process args of config")
-			if err := j.overwriteConfig(cfg, filepath.Base(m.Cfg.SocketPath), rootPathToConfig); err != nil {
+			if err := j.overwriteConfig(cfg, machineConfig, filepath.Base(m.Cfg.SocketPath), rootPathToConfig); err != nil {
```
sipsma: I'm not positive but I would have thought this would need to be `m.Cfg` as opposed to `machineConfig`. The reason being that `machineConfig` is the one [provided during `buildVMConfiguration` to `BuildJailedRootHandler` and `NewMachine`](https://github.com/firecracker-microvm/firecracker-containerd/blob/5f3e0c74a7c707ac7e0cd0a8f1466abff75d47de/runtime/service.go#L486), but `NewMachine` doesn't modify that `machineConfig` object as its [signature has just `Config`, not `*Config`](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/eec83f3dcb7b4eb4eeb1e2c5f975715a25f79e96/machine.go#L257).

`m.Cfg` should have the fully updated config though if I'm understanding correctly (which I may not be).

xibz: Whoops, yea, I meant `m.Cfg`. Too many configs lol

xibz: Fixed!

runtime/runc_jailer.go
kzys: But the function is called by BuildJailedMachine() -> BuildJailedRootHandler() (this one cloud be private) -> overwriteConfig(). Can we remove this one for now? I don't think this file read would be a hot spot.

runtime/cni_integ_test.go
```diff
@@ -39,8 +39,7 @@ import (
 )
 
 func TestCNISupport_Isolated(t *testing.T) {
-	prepareIntegTest(t)
-
+	prepareIntegTest(t, withJailer())
```
sipsma: I just looked in buildkite, it seems like this isn't working as expected, `TestCNISupport_Isolated` is still running w/ the noop jailer according to the logs, do we need to update the `CreateVM` request in here to specify a non-nil jailer?

runtime/runc_jailer.go
xibz: Was worried about the file system call, but if you think it isn't going to be a hot spot I can remove it.

runtime/cni_integ_test.go
xibz: Doh, fixed.

runtime/runc_jailer.go
```diff
@@ -353,19 +379,11 @@ func (j runcJailer) jailerCommand(containerName string, isDebug bool) *exec.Cmd
...
-	}
-
-	if err := json.Unmarshal(configBytes, &spec); err != nil {
-		return err
-	}
-
+func (j *runcJailer) overwriteConfig(cfg *Config, machineConfig *firecracker.Config, socketPath, configPath string) error {
+	var err error
+	// here we attempt to cache the runc config. If the config has already been
+	// cached, we will return immediately
```
sipsma: Is this comment still relevant? Otherwise LGTM

xibz: Good catch, fixed!

runtime/runc_jailer.go
```diff
@@ -63,6 +68,19 @@ func newRuncJailer(ctx context.Context, logger *logrus.Entry, ociBundlePath, run
...
+	spec := specs.Spec{}
+	var configBytes []byte
+	configBytes, err := ioutil.ReadFile(runcConfigPath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read firecracker-runc-config.json")
+	}
+
+	if err = json.Unmarshal(configBytes, &spec); err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal firecracker-runc-config.json")
+	}
```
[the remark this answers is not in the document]
xibz: Fixed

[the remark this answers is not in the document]
kzys: You don't have to use Sprintf inside Wrapf :)

xibz:
> You don't have to use Sprintf inside Wrapf :)

lol, fixed!

PR #306, opened 2019-10-25
Adds cgroup path to CreateVMResponse

xibz:
This commit adds a response to CreateVM. The CreateVMResponse contains
the CgroupPath field that will be returned when jailing is enabled.


runtime/jailer.go
```diff
@@ -72,5 +76,5 @@ func newJailer(
 	}
 
 	l := logger.WithField("jailer", "runc")
-	return newRuncJailer(ctx, l, ociBundlePath, service.config.JailerConfig.RuncBinaryPath, jailerUID, jailerGID)
+	return newRuncJailer(ctx, l, ociBundlePath, service, jailerUID, jailerGID)
```
IRCody: Not sure if I like passing in the service. We are doing this to get the VMID?

runtime/service_integ_test.go
```diff
@@ -277,9 +277,10 @@ func TestMultipleVMs_Isolated(t *testing.T) {
 
 			rootfsPath := defaultVMRootfsPath
 
+			vmIDStr := strconv.Itoa(vmID)
 			fcClient := fccontrol.NewFirecrackerClient(pluginClient.Client())
 			req := &proto.CreateVMRequest{
-				VMID: strconv.Itoa(vmID),
+				VMID: vmIDStr,
```
IRCody: Since CreateVM now returns a response with data is it worth adding tests around that in this file?

firecracker-control/local.go
```diff
@@ -78,7 +78,7 @@ func newLocal(ic *plugin.InitContext) (*local, error) {
 // CreateVM creates new Firecracker VM instance. It creates a runtime shim for the VM and the forwards
 // the CreateVM request to that shim. If there is already a VM created with the provided VMID, then
 // AlreadyExists is returned.
-func (s *local) CreateVM(requestCtx context.Context, req *proto.CreateVMRequest) (*empty.Empty, error) {
+func (s *local) CreateVM(requestCtx context.Context, req *proto.CreateVMRequest) (*proto.CreateVMResponse, error) {
```
sipsma: I like the idea of returning metadata from `CreateVM` but feel like if we are going to include the cgroup path in the `CreateVM` response it should also be included in the `GetVMInfoResponse` object.

That in turn makes me wonder if we should just have `CreateVM` return everything that's currently in `GetVMInfoResponse`, something like:
```
message VMInfo {
    string VMID = 1;
    uint32 ContextID = 2;
    string SocketPath = 3;
    string LogFifoPath = 4;
    string MetricsFifoPath = 5;
    string CgroupPath = 6;
}

message CreateVMResponse {
    VMInfo VMInfo = 1;
}

service Firecracker {
    rpc CreateVM(CreateVMRequest) returns (CreateVMResponse);

    rpc GetVMInfo(GetVMInfoRequest) returns (VMInfo);

    ...
}
```

runtime/jailer.go
xibz: We need the config and the vmID, so I figured since we already have a lot of parameters, it made sense to just pass in the service.

runtime/service_integ_test.go
xibz: Ah good call, I changed the `TestMultipleVMs_Isolated` to check the cgroup path in the response instead of one the jailer.

firecracker-control/local.go
xibz: @sipsma - I like the idea, but I don't know if I like the example as much. I am on the fence on whether we should use a response object as another object in another response. I would be more in favor of something like this,

```
message GetVMInfoResponse {
    string VMID = 1;
    uint32 ContextID = 2;
    string SocketPath = 3;
    string LogFifoPath = 4;
    string MetricsFifoPath = 5;
    string CgroupPath = 6;
}

message CreateVMResponse {
    string VMID = 1;
    uint32 ContextID = 2;
    string SocketPath = 3;
    string LogFifoPath = 4;
    string MetricsFifoPath = 5;
    string CgroupPath = 6;
}

service Firecracker {
    rpc CreateVM(CreateVMRequest) returns (CreateVMResponse);

    rpc GetVMInfo(GetVMInfoRequest) returns (GetVMInfoResponse);

    ...
}
```

This makes the request and response names consistent and better for our users as they can easily predict what the response or request names are going to be. I don't mind the duplication here as there is a clear separation between the APIs. However, if that is something a lot of people disagree with I'd be okay with introducing the `VMInfo` structure, and have that returned in both as a field.

sipsma: Sure, I think that's fine too, the duplication is pretty minor maintenance overhead and would possibly be desirable anyways if we ever end up needing to diverge in the future.

xibz: Awesome, fixed!

proto/firecracker.proto
```diff
@@ -36,6 +36,15 @@ message CreateVMRequest {
     JailerConfig JailerConfig = 10;
 }
 
+message CreateVMResponse {
+    string VMID = 1;
+    uint32 ContextID = 2;
```
sipsma: nit: `ContextID` is no longer set or used anywhere after the recent vsock changes, we may as well remove it from both CreateVMResp and GetVMInfoResp

runtime/jailer.go
```diff
@@ -58,6 +58,10 @@ type jailer interface {
 	StubDrivesOptions() []stubDrivesOpt
 }
 
+type cgroupPather interface {
+	CgroupPath() string
+}
```
sipsma: Why not just add `CgroupPath` to the `jailer` interface? It seems a lot simpler to just put it there given both of our implementations have methods for it anyways.

runtime/runc_jailer.go
```diff
@@ -412,6 +418,8 @@ func (j runcJailer) setDefaultConfigValues(cfg *Config, socketPath string, spec
 		spec.Process.Args = cmd.Args
 	}
 
+	spec.Linux.CgroupsPath = j.CgroupPath()
```
sipsma: nit: can we add a debug log saying `using cgroup path xyz` around here?

runtime/service_integ_test.go
```diff
@@ -415,9 +416,12 @@ func TestMultipleVMs_Isolated(t *testing.T) {
 
 						jailer := &runcJailer{
 							ociBundlePath: string(shimDir),
+							vmID:          vmIDStr,
 						}
 						_, err = os.Stat(jailer.RootPath())
 						require.NoError(t, err, "failed to stat root path of jailer")
+						_, err = os.Stat(filepath.Join("/sys/fs/cgroup/cpu", resp.CgroupPath))
```
sipsma: Can we add an assertion that `resp.CgroupPath` is the expected value too? Otherwise I feel like this stat could succeed even if the cgroup path was unexpectedly `""`, `"/"` or something like that

runtime/runc_jailer.go
```diff
@@ -46,21 +46,23 @@ type runcJailer struct {
 	runcBinaryPath string
 	uid            uint32
 	gid            uint32
+	vmID           string
 }
 
 const firecrackerFileName = "firecracker"
 
-func newRuncJailer(ctx context.Context, logger *logrus.Entry, ociBundlePath, runcBinPath string, uid, gid uint32) (*runcJailer, error) {
+func newRuncJailer(ctx context.Context, logger *logrus.Entry, ociBundlePath string, service *service, uid, gid uint32) (*runcJailer, error) {
```
sipsma: nit: I'd personally prefer we just separately provide `runcBinaryPath` and `vmID` given that `service` is a pretty massive object, it just makes it easier to know+control what the runc jailer is actually depending on

runtime/jailer.go
xibz: cgroup path doesn't feel like it belongs in a jailer interface and seems to be more of a implementation detail which is why I made separate. The `noop` jailer doesn't implement it, and if we decided to use Firecracker's jailer, it would also be absent.

proto/firecracker.proto
xibz: Done!

runtime/runc_jailer.go
xibz: Yep, fixed!

runtime/service_integ_test.go
xibz: Fixed!

runtime/runc_jailer.go
xibz: Fixed

runtime/jailer.go
kzys: If so, shouldn't we remove CgroupPath from noopJailer?

xibz: Thought I removed it, but I guess not o.O

Anyways, fixed!

sipsma: I'm confused why `CgroupPath` wouldn't belong in the jailer, it is a piece of information that the jailer configures as part of creating the jail and that we expose to users by sending it w/ the CreateVMResp and GetVMInfoResp (which makes it more than just an implementation detail). If we created a common jailing interface, then I would just imagine that Firecracker's jailer implementation would just return `""`, like the noop jailer, since it doesn't configure them. It seems like exactly what should be in a jailer interface.

xibz: It's just that. Only the runc jailer sets the value. Why put it in a **common** interface, if it is only used for one of the jailer types?

sipsma: Because the other types are explicitly not setting it, which is just as relevant as it being set by the runc implementation. It's something that can be configured by the general concept of a "jailer" (and is by many), so it is confusing for it not to be part of a "jailer" interface.

xibz: They aren't setting it because it doesn't have a notion of a cgroup path which I have hard time following to why it should be in the `jailer` interface.

xibz: @samuelkarp @nmeyerhans - any preference on this?

kzys: The lack of cgroups can be represented by either not implementing an interface (cgroupPather) or returning a special value (""). Using an interface may be more scalable if we have a lot of differences between jailers, where implementing all of required methods to just say "no" is hassle.

But, we haven't reached that stage, and we may not. So I'm fine with both ways :)

sipsma: If no one else has any feelings either way on this, then given this is a subjective issue I think the rule of thumb should be to go with the person who actually wrote the code :-)

The only other thing I'd add is that part of the reason this causes me deep confusion is that it seems to negate the existence of the jailer interface. If we want to have separate interfaces for each possible feature of a jailer, then why would we even have a jailer interface at all, especially given we have a no-op jailer implementation? I personally greatly prefer the simplicity and explicitness of just maintaining a single interface here (even if it's slightly more verbose), but if we want to split up features into separate interfaces then I think we should at least be consistent in doing so, which would suggest that we get rid of the no-op jailer and replace it with interface tests that have a default code path in case the method doesn't exist.

xibz: @sipsma - I am thinking if there is a solution that could make both of us happy. So I was thinking perhaps adding something like `SetCreateVMResponse`. The only issue I have with this is you'd have to look at what the jailer is overwriting when setting any of the responses fields, but I think that may be fine. What are your thoughts on this approach?

sipsma: Honestly I like that option less than using `cgroupPather` because it ties the jailer directly into firecracker-containerd's api, whereas the eventual end goal in my mind is to have a generic jailer interface that's independent of firecracker-containerd and lives in the go sdk. But like I said before, these are stylistic concerns and they can be modified in the future pretty easily if we agree to, so I don't want all of this to block the PR if there are no feelings either way from anyone else right now. We can go with your original option or the CreateVMResponse-based option for now; when we refactor the jailer some more in the future would be a good time to discuss all this further.

Issue #248, opened 2019-08-22
Send PATCH request in SetVMMetadata API


haikuoliu:
Hi,

Currently the "SetVMMetadata" API sends PUT request to update the entire MMDS JSON, but we have use cases where we only want to update part of the MMDS, so it would be nice to have PATCH request support in "SetVMMetadata" API.

Let me know if it makes sense,
Thank you 
Haikuo

IRCody:
Closed by #311

PR #311, opened 2019-10-26
Add {Get,Update}VMMetadata calls to the fc-control api.

IRCody:
This enables callers of the fc-control api  to patch
metadata instead of requiring a user to submit all the
metadata everytime.

*Issue #, if available:* #248


runtime/service_integ_test.go
```diff
@@ -1003,3 +1003,58 @@ func TestDriveMountFails_Isolated(t *testing.T) {
...
+
+	rootfsPath := defaultVMRootfsPath
+	_, err = fcClient.CreateVM(ctx, &proto.CreateVMRequest{
+		VMID: "1",
+		MachineCfg: &proto.FirecrackerMachineConfiguration{
+			MemSizeMib: 512,
+		},
+		RootDrive: &proto.FirecrackerRootDrive{
+			HostPath: rootfsPath,
+		},
```
sipsma: nit: no need to specify this, the runtime's default is the same value

runtime/service_integ_test.go
```diff
@@ -1003,3 +1003,58 @@ func TestDriveMountFails_Isolated(t *testing.T) {
 		assert.Error(t, err, "unexpectedly succeeded in creating a VM with a drive mount under banned path")
 	}
 }
+
+func TestUpdateVMMetadata_Isolated(t *testing.T) {
```
sipsma: I think there would be some value in adding a container to this test that gets the VM metadata too so we can verify that we are able to spawn containers that have access to it too. There's an existing test case that uses `wget` in the alpine container image which could probably be adapted to hit the mmds endpoint instead: https://github.com/firecracker-microvm/firecracker-containerd/blob/38742eeb4fb5d549b31727a57b8c79e88b204b59/runtime/cni_integ_test.go#L112

IRCody: That's a good idea, I'll try to add it.

runtime/service_integ_test.go
IRCody: I tried removing this but it fails to create the vm if I do, saying it tries to stat the hostpath and it's "". I changed it to not use rootfsPath and just use the defaultVMRootFSPath directly though.

runtime/service_integ_test.go
IRCody: Added this based on the example you linked.

PR #312, opened 2019-10-28
Run "make demo-network" earlier to cache the result

kzys:
*Issue #, if available:*

NA

*Description of changes:*

The target installs CNI-related binaries through "go get" and "go install".
Having the step before copying the entire firecracker-containerd directory
would allow us to cache the result, to make increment build faster.

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -112,11 +112,7 @@ RUN mkdir -p /var/lib/firecracker-containerd/runtime \
 	&& chmod 0444 default-vmlinux.bin \
 	&& mv default-vmlinux.bin /var/lib/firecracker-containerd/runtime/default-vmlinux.bin
 
-COPY --from=firecracker-containerd-build /home/builder/firecracker-containerd /firecracker-containerd
-COPY tools/image-builder/rootfs.img /var/lib/firecracker-containerd/runtime/default-rootfs.img
-COPY --from=firecracker-containerd-build /output/* /usr/local/bin/
-
-RUN --mount=type=cache,from=build-base,source=/home/builder/go/pkg/mod,target=/tmp/go/pkg/mod,readonly \
+RUN --mount=type=cache,from=build-base,source=/home/builder/go/pkg/mod,target=/tmp/go/pkg/mod \
 	mkdir -p ${GOPATH}/pkg/mod \
 	&& cp -R /tmp/go/pkg/mod/* ${GOPATH}/pkg/mod \
 	&& cp -R /tmp/go/pkg/mod/* /home/builder/go/pkg/mod \
@@ -148,7 +144,6 @@ FROM firecracker-containerd-test as firecracker-containerd-integ-test
 ARG FIRECRACKER_TARGET=x86_64-unknown-linux-musl
 
 COPY _submodules/firecracker/target/$FIRECRACKER_TARGET/release/firecracker /usr/local/bin/
-COPY _submodules/firecracker/target/$FIRECRACKER_TARGET/release/jailer /usr/local/bin/
 COPY _submodules/runc/runc /usr/local/bin
 COPY tools/image-builder/rootfs.img /var/lib/firecracker-containerd/runtime/default-rootfs.img
 COPY runtime/firecracker-runc-config.json.example /etc/containerd/firecracker-runc-config.json


tools/docker/Dockerfile
```diff
@@ -112,17 +112,29 @@ RUN mkdir -p /var/lib/firecracker-containerd/runtime \
...
+COPY --from=firecracker-containerd-build \
+    /home/builder/firecracker-containerd/Makefile \
+    /home/builder/firecracker-containerd/go.mod \
+    /home/builder/firecracker-containerd/go.sum \
+    /home/builder/firecracker-containerd/tools \
+    /firecracker-containerd/
+COPY --from=firecracker-containerd-build \
+    /home/builder/firecracker-containerd/tools \
+    /firecracker-containerd/tools
+RUN make -C /firecracker-containerd demo-network
```
sipsma: We are still running demo-network below on line 176, which shouldn't be needed

tools/docker/Dockerfile
```diff
@@ -112,17 +112,29 @@ RUN mkdir -p /var/lib/firecracker-containerd/runtime \
...
+
+# Copying the entire directory likely invalidate the build cache.
+# Time-consuming operations should be happening before this step.
 COPY --from=firecracker-containerd-build /home/builder/firecracker-containerd /firecracker-containerd
```
sipsma: Can you clarify the comment here? The binaries installed via demo-network shouldn't be in the build cache from previous steps as they aren't referenced before demo-network is run. Unless I'm missing something I think it would be simpler to just move demo-network to run after this line

tools/docker/Dockerfile
kzys: Ah right. I'd like to remove that from firecracker-containerd-integ-test.

tools/docker/Dockerfile
kzys: I wanted to skip running `RUN make -C /firecracker-containerd demo-network` when we don't have to. But it would be faster if we have cache under GOPATH. Let me double-check.

kzys: Even without downloading, demo-network took 3 seconds

```
#34 [firecracker-containerd-integ-test 8/10] RUN make -C /firecracker-contai...
#34 0.503 make: Entering directory '/firecracker-containerd'
#34 0.530 GOBIN=/opt/cni/bin GO111MODULE=off go get -u github.com/containernetworking/plugins/plugins/main/ptp
#34 1.440 GOBIN=/opt/cni/bin GO111MODULE=off go get -u github.com/containernetworking/plugins/plugins/ipam/host-local
#34 2.318 GOBIN=/opt/cni/bin GO111MODULE=off go get -u github.com/containernetworking/plugins/plugins/meta/firewall
#34 3.078 make: Leaving directory '/firecracker-containerd'
#34 DONE 3.2s
```

So I'd like to skip this step when we can.

kzys: In addition to that, after having https://github.com/golang/go/commit/ab724d43efe7e1a7516c1d13e40b55dca26a61b4, we probably can move "go get -d ./..." right after copying go.sum and go.mod.

sipsma: Oh okay I think I understand the original intention better, is it so we only re-download the binaries when there was an actual change to the Makefile's target for those binaries as opposed to when literally anything changes in the entirety of firecracker-containerd?

Maybe there's a simpler way I'm not thinking of, but I'm not sure we will be able to do that without larger refactoring of this whole dockerfile and the makefile of the repo. The problem is that because we copy the directory in early on line 74, everything after that will be invalidated whenever anything under the firecracker-containerd dir changes.

I think the better way would be to use the approach of the image-builder and firecracker-builder, which is to just have a simple Dockerfile with the host-dependencies needed to perform the build (i.e. golang at some version and few other system libraries) and then define all the logic of the build as steps in the main Makefile that run the build via `docker run` with the directory of the host bind mounted in. That allows the host filesystem to serve as the "cache" and normal Makefile logic to determine when to re-run a step. For the demo-network binaries specifically I think we'd also need to first download to `BINPATH` and then install to `CNI_BIN_PATH` so we can use the `BINPATH` as the cache and only copy binaries to the image's `/opt/cni/bin`

Obviously this is a much larger change, what if for now we just move the `demo-network` step earlier and I create an issue for the larger change I was describing?

kzys: Maybe I should say "intermediate layers" instead of "build cache"?

kzys: Yeah. That would work. Let me simply this branch in that case.

sipsma: Cool, sounds good, I opened an issue here for the larger-scale changes: https://github.com/firecracker-microvm/firecracker-containerd/issues/313

I personally won't have time in the next few weeks for that, but would be happy to make the changes once I have time as it will definitely result in a much faster dev-test loop for everyone.

kzys: Actually moving that doesn't result any performance gain. Will revisit as #313.

PR #316, opened 2019-10-30
Fine-tune runc-related permissions

kzys:
*Issue #, if available:*

*Description of changes:*

While most of them are harmless to have, it may be better to keep files less readable/writable when we can.

### DIFF
--- a/runtime/runc_jailer.go
+++ b/runtime/runc_jailer.go
@@ -183,7 +183,13 @@ func (j *runcJailer) BuildJailedRootHandler(cfg *Config, socketPath *string, vmI
 						return errors.Wrapf(err, "failed to stat drive %q", drivePath)
 					}
 
-					if err := copyFile(drivePath, newDrivePath, info.Mode()); err != nil {
+					// If the file will be mounted as read-only, drop write permissions
+					mode := info.Mode()
+					if *d.IsReadOnly {
+						mode &= 0555
+					}
+
+					if err := copyFile(drivePath, newDrivePath, mode); err != nil {
 						return errors.Wrapf(err, "failed to copy drive %v", drivePath)
 					}
 				}
--- a/runtime/runc_jailer.go
+++ b/runtime/runc_jailer.go
@@ -183,18 +183,12 @@ func (j *runcJailer) BuildJailedRootHandler(cfg *Config, socketPath *string, vmI
 				defer f.Close()
 
 				if !internal.IsStubDrive(f) {
-					info, err := os.Stat(drivePath)
-					if err != nil {
-						return errors.Wrapf(err, "failed to stat drive %q", drivePath)
+					mode := 0600
+					if firecracker.BoolValue(d.IsReadOnly) {
+						mode = 0400
 					}
 
-					// If the file will be mounted as read-only, drop write permissions
-					mode := info.Mode()
-					if *d.IsReadOnly {
-						mode &= 0555
-					}
-
-					if err := copyFile(drivePath, newDrivePath, mode); err != nil {
+					if err := copyFile(drivePath, newDrivePath, os.FileMode(mode)); err != nil {
 						return errors.Wrapf(err, "failed to copy drive %v", drivePath)
 					}
 				}
--- a/runtime/runc_jailer.go
+++ b/runtime/runc_jailer.go
@@ -156,8 +156,11 @@ func (j *runcJailer) BuildJailedRootHandler(cfg *Config, socketPath *string, vmI
 			newKernelImagePath := filepath.Join(rootPath, kernelImageFileName)
 			j.logger.WithField("newKernelImagePath", newKernelImagePath).Debug("copying kernel image")
 
-			if err := copyFile(m.Cfg.KernelImagePath, newKernelImagePath, 0444); err != nil {
+			if err := copyFile(m.Cfg.KernelImagePath, newKernelImagePath, 0400); err != nil {
 				return errors.Wrap(err, "failed to mount kernel image")
+			}
+			if err := os.Chown(newKernelImagePath, int(j.uid), int(j.gid)); err != nil {
+				return err
 			}
 
 			m.Cfg.KernelImagePath = kernelImageFileName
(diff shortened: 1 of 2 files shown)


runtime/runc_jailer.go
```diff
@@ -156,9 +158,12 @@ func (j *runcJailer) BuildJailedRootHandler(cfg *Config, socketPath *string, vmI
 			newKernelImagePath := filepath.Join(rootPath, kernelImageFileName)
 			j.logger.WithField("newKernelImagePath", newKernelImagePath).Debug("copying kernel image")
 
-			if err := copyFile(m.Cfg.KernelImagePath, newKernelImagePath, 0444); err != nil {
+			if err := copyFile(m.Cfg.KernelImagePath, newKernelImagePath, 0400); err != nil {
 				return errors.Wrap(err, "failed to mount kernel image")
 			}
+			if err := os.Chown(newKernelImagePath, int(j.uid), int(j.gid)); err != nil {
```
samuelkarp: Do we want to chown it?  I think leaving at 0444 with ownership different to the user inside the jail is probably sufficient.

kzys: We chown a lot of stuff, including a copied firecracker binary. Would it be better to chown only $OCI_BUNDLE_PATH/rootfs/ where we have firecracker, Linux kernel and drive files?

runtime/runc_jailer.go
```diff
@@ -183,7 +188,13 @@ func (j *runcJailer) BuildJailedRootHandler(cfg *Config, socketPath *string, vmI
 						return errors.Wrapf(err, "failed to stat drive %q", drivePath)
 					}
 
-					if err := copyFile(drivePath, newDrivePath, info.Mode()); err != nil {
+					// If the file will be mounted as read-only, drop write permissions
+					mode := info.Mode()
+					if *d.IsReadOnly {
+						mode &= 0555
```
sipsma: Can this just be `0400`? Off the top of my head I can't think of why group+other would need any access and I can't think of why user would need executable access.

runtime/runc_jailer.go
sipsma: I feel like if we can get 0400 permissions by doing the chown we may as well as it's technically a reduced permission (even if marginally so). I can't see any downside to doing what's in the PR.

runtime/runc_jailer.go
sipsma: Also, in the case where this isn't read only, can we set to `0600`?

kzys: Yeah. 0400 / 0600 is much simpler.

runtime/runc_jailer.go
```diff
@@ -183,7 +188,13 @@ func (j *runcJailer) BuildJailedRootHandler(cfg *Config, socketPath *string, vmI
 						return errors.Wrapf(err, "failed to stat drive %q", drivePath)
 					}
 
-					if err := copyFile(drivePath, newDrivePath, info.Mode()); err != nil {
+					// If the file will be mounted as read-only, drop write permissions
+					mode := info.Mode()
+					if *d.IsReadOnly {
```
xibz: This value can be `nil`.

```go
if firecracker.BoolValue(d.IsReadOnly) {
}
```

runtime/runc_jailer.go
```diff
@@ -178,13 +172,12 @@ func (j *runcJailer) BuildJailedRootHandler(cfg *Config, socketPath *string, vmI
 				defer f.Close()
 
 				if !internal.IsStubDrive(f) {
-					info, err := os.Stat(drivePath)
-					if err != nil {
-						return errors.Wrapf(err, "failed to stat drive %q", drivePath)
+					mode := 0600
```
xibz: Shouldn't we use the original mode? This kind of couples this to the stub drive handler knowing that it chmods it to `0600`. I'd rather use the original mode of the file.

kzys: While the stub drive handler is using the same bits, the intention is picking the minimal permission bits we need to make this file both readable and writable from the Firecracker process which is running as the jailer user.

Even if we need to change the permission bits on the handler side (actually, we might be able to make the dummy file read-only, assuming that actual writes are happening on a different file...?), we don't have to change the bits here.

xibz: This is for non-stub drives. So, we want them to be what users have specified.

sipsma: I'm not sure what reason we have for using the permissions of the original file. We are creating a new file inside the jail (even if its contents are a copy of that on the outside) and we can know that this new file inside the jail only needs 0600 or 0400. What advantage is there to making the permissions match what's present on the file outside the jail?

kzys: As far as I understand, because of the permission of the parent directory, files which are copied to this directory are anyway not readable/writable from anyone else.

If we want to keep the original permissions, we probably re-think about the permissions on the parent directory to make a specific scenario possible.

xibz: Users may expect them to be of a certain mode as they may want to do something with those drives themselves. Firecracker's jailer does the same thing and doesn't chown or chmod any of the drives

sipsma: > As far as I understand, because of the permission of the parent directory, files which are copied to this directory are anyway not readable/writable from anyone else.

That may be true but if we are going for full-out defense in-depth (which we may as well here since there's basically no cost to setting permissions on a file in terms of complexity IMO) then I still think it makes to set every individual permission as minimal as possible.

> Users may expect them to be of a certain mode as they may want to do something with those drives themselves.

These files are not the original ones created by the users, they are created just for the jail we are creating (right?). I don't think we should operate under the assumption that users should feel free to directly interact arbitrarily with jailed files (they obviously are free to interact with the original files outside the jail) as the point of the jail is to create something isolated from the rest of the system with very fine grained permissions. I just don't see what use-cases we'd be enabling by trying to match the permissions, but if there are any I'm not thinking of let me know.

xibz: The drives may be a copy right now, but I think in the future this will be changed to be a bind mount. So having them be able to do things seems feasible, the question here is should we allow it? Maybe reading the drive is something they may want to do?

sipsma: Yes if we change to a bind-mount in the future then all of this and much more would change. But at the moment we're doing a copy and until we switch to a different approach, why wouldn't we just go with the least possible privilege we are aware there is a need for? There's no expectation for 100% backwards compatibility (changing to a bind-mount would be a much larger break anyways) and I can't imagine any scenario in which we'd ever want users to directly interact with files in the jail. If they really, really want to, they can be root and which point all of this is moot anyways.

Issue #194, opened 2019-05-30
Use devmapper snapshotter in quickstart guide


samuelkarp:
The quickstart guide uses the naive snapshotter, which works but has poor performance.  We should update the quickstart to use the devmapper snapshotter instead.

mxpv:
https://github.com/firecracker-microvm/firecracker-containerd/pull/361

PR #317, opened 2019-10-31
Modify getting-started to use devmapper snapshottter

IRCody:

*Issue #, if available:* #194 

*Description of changes:*
Modifies the getting started guide to use dev-mapper snapshotter instead of the naive snapshotter.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -136,9 +136,9 @@ state = "/run/firecracker-containerd"
 [grpc]
   address = "/run/firecracker-containerd/containerd.sock"
 [proxy_plugins]
-  [proxy_plugins.firecracker-naive]
+  [proxy_plugins.firecracker-devmapper]
     type = "snapshot"
-    address = "/var/run/firecracker-containerd/naive-snapshotter.sock"
+    address = "/var/run/firecracker-containerd/devmapper-snapshotter.sock"
 
 [debug]
   level = "debug"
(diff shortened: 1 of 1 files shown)


mxpv:
We should mention that this kind of configuration is slow, should be avoided in production and suitable only for local testing. https://docs.docker.com/storage/storagedriver/device-mapper-driver/#performance-best-practices

IRCody:
> We should mention that this kind of configuration is slow, should be avoided in production and suitable only for local testing. https://docs.docker.com/storage/storagedriver/device-mapper-driver/#performance-best-practices

Do you think a Note: This is not a production ready pool configuration, see : <link> is enough or do you think it might be worth saying more?

kzys:
containerd 1.3 is already having the devmapper snapshotter. Don't we want to use the upstream one instead of our plugin one?

mxpv:
> Do you think a Note: This is not a production ready configuration, see link is enough or do you think it might be worth saying more?

May be something like: `Note: the configuration with loopback devices is slow and not intended for use in production` ?

> containerd 1.3 is already having the devmapper snapshotter. Don't we want to use the upstream one instead of our plugin one?

Yeah. We should start this transition. But I don't think it has be done here as this PR focuses on thin-pool creation. I've created an issue to track: https://github.com/firecracker-microvm/firecracker-containerd/issues/318

PR #319, opened 2019-11-02
Move proto compiling to a container

IRCody:
*Description of changes:*
Makes it so that running `make proto` from the root of the project
compiles the protos in a docker container defined in
tools/docker/Docker.proto-builder. This image is build as part of `make
deps`.

This changes the default behavior but the previous behavior is still available by running `make proto` in the proto/ dir.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,8 @@ RUNC_DIR=$(SUBMODULES)/runc
 RUNC_BIN=$(RUNC_DIR)/runc
 RUNC_BUILDER_NAME?=runc-builder
 
+PROTO_BUILDER_NAME?=proto-builder
+
 # Set this to pass additional commandline flags to the go compiler, e.g. "make test EXTRAGOARGS=-v"
 export EXTRAGOARGS?=
 
@@ -45,7 +47,11 @@ $(SUBDIRS):
 	$(MAKE) -C $@
 
 proto:
-	PATH=$(BINPATH):$(PATH) $(MAKE) -C proto/ proto
+	DOCKER_BUILDKIT=1 docker build \
+		--file tools/docker/Dockerfile.proto-builder \
+		--tag localhost/$(PROTO_BUILDER_NAME):${DOCKER_IMAGE_TAG} \
+		$(CURDIR)/tools/docker
+	PATH=$(BINPATH):$(PATH) $(MAKE) -C proto/ proto-docker
 
 clean:
 	for d in $(SUBDIRS); do $(MAKE) -C $$d clean; done
@@ -64,6 +70,7 @@ distclean: clean
 	docker volume rm -f $(CARGO_CACHE_VOLUME_NAME)
 	$(call rmi-if-exists,localhost/firecracker-containerd-integ-test:$(DOCKER_IMAGE_TAG))
 	$(call rmi-if-exists,localhost/firecracker-containerd-test:$(DOCKER_IMAGE_TAG))
+	$(call rmi-if-exists,localhost/$(PROTO_BUILDER_NAME):$(DOCKER_IMAGE_TAG))
 
 lint:
 	$(BINPATH)/ltag -t ./.headers -excludes "tools $(SUBMODULES)" -check -v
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -45,7 +47,13 @@ $(SUBDIRS):
 	$(MAKE) -C $@
 
 proto:
-	PATH=$(BINPATH):$(PATH) $(MAKE) -C proto/ proto
+	DOCKER_BUILDKIT=1 docker build \
+		--progress=plain \
```
samuelkarp: Do you really need `--progress plain` and `--quiet` here?

proto/Makefile
```diff
@@ -23,6 +23,10 @@ proto: $(PROTO_GEN_SRC)
 	PROTOPATH=$(CURDIR) $(MAKE) -C service/fccontrol proto
 	PROTOPATH=$(CURDIR) $(MAKE) -C service/drivemount proto
 
+proto-docker:
+	docker run --rm \
+		-v $(CURDIR):/protobuf \
+		localhost/proto-builder
```
samuelkarp: This probably needs `localhost/proto-builder:$(DOCKER_IMAGE_TAG)` for the override specified in the other Makefile to work.

You may also need to add the `DOCKER_IMAGE_TAG?=latest` line.

tools/docker/Dockerfile.proto-builder
```diff
@@ -0,0 +1,14 @@
...
+	libprotobuf-dev \
+	protobuf-compiler \
+	&& go get -u github.com/containerd/ttrpc/cmd/protoc-gen-gogottrpc \
+	&& go get -u github.com/gogo/protobuf/protoc-gen-gogo \
+	&& mkdir /protobuf
+
+WORKDIR /protobuf
+ENTRYPOINT ["/usr/bin/make"]
+CMD ["proto"]
+
```
samuelkarp: nit: extra blank lines

tools/docker/Dockerfile.proto-builder
```diff
@@ -0,0 +1,14 @@
+FROM golang:1.12-stretch
```
samuelkarp: This file needs the license header.

Makefile
IRCody: Nope.

proto/Makefile
IRCody: Good call, thanks!

tools/docker/Dockerfile.proto-builder
IRCody: Went ahead and added it to tools/docker/Dockerfile.runc-builder to.

proto/Makefile
```diff
@@ -23,6 +24,11 @@ proto: $(PROTO_GEN_SRC)
 	PROTOPATH=$(CURDIR) $(MAKE) -C service/fccontrol proto
 	PROTOPATH=$(CURDIR) $(MAKE) -C service/drivemount proto
 
+proto-docker:
+	docker run --rm \
```
sipsma: Does this result in the proto files being owned by root? If so, we could set `UID:=$(shell id -u)` in the makefile and then specify `--user $(UID)` for the docker run command (which we do for some of the other builders in this repo). Then users don't have to deal with files that require `sudo` to modify/delete

tools/docker/Dockerfile.proto-builder
```diff
@@ -0,0 +1,25 @@
...
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+FROM golang:1.12-stretch
+
+RUN apt-get update && apt-get install --yes --no-install-recommends \
+	libprotobuf-dev \
+	protobuf-compiler \
+	&& go get -u github.com/containerd/ttrpc/cmd/protoc-gen-gogottrpc \
```
sipsma: Maybe we should remove these binaries from the `deps` target in the repo's top-level Makefile since we now want the proto builds to be done from this container instead.

proto/Makefile
IRCody: It does! I think we need to add the GID also.

tools/docker/Dockerfile.proto-builder
IRCody: I thought about that but I wasn't sure if we wanted to leave the ability to do builds outside the container? I don't feel strongly about it.

kzys: +1 for removing them from deps, to make our full build faster.

IRCody: Removed.

IRCody:
@sipsma: I created #323 to track adding a CI step to generate/compare protos.

PR #322, opened 2019-11-05
Update to go1.13

IRCody:
- Add go1.13 as a build target for travis.
- Set the go version in go.mod at 1.11.
- Update a line in go.mod that wasn't compliant with the more stringent
validation in 1.13.
- Update Dockerfiles to use 1.13.

See [here](https://tip.golang.org/doc/go1.13#version-validation) for information on the version validation. The error I was seeing was:

```go: github.com/firecracker-microvm/firecracker-go-sdk@v0.17.1-0.20191030150325-dbf9a1e05f090: invalid pseudo-version: revision is longer than canonical (dbf9a1e05f09)```

This seems to be happening b/c the git prefix was 1 char to long (13 instead of 12).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/docker/go-events v0.0.0-20170721190031-9461782956ad // indirect
 	github.com/docker/go-metrics v0.0.0-20181218153428-b84716841b82 // indirect
 	github.com/docker/go-units v0.3.3
-	github.com/firecracker-microvm/firecracker-go-sdk v0.17.1-0.20191030150325-dbf9a1e05f090
+	github.com/firecracker-microvm/firecracker-go-sdk v0.17.1-0.20191029213755-dbf9a1e05f09
 	github.com/go-ole/go-ole v1.2.4 // indirect
 	github.com/godbus/dbus v0.0.0-20181025153459-66d97aec3384 // indirect
 	github.com/gofrs/uuid v3.2.0+incompatible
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: go.mod [hunk 1]
```diff
@@ -53,3 +53,5 @@ require (
 
 // Workaround for github.com/containerd/containerd issue #3031
 replace github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
+
+go 1.11
```
samuelkarp: What's this line for?

IRCody: Snipped from go help go.mod
```
The expected language version, set by the go directive, determines
which language features are available when compiling the module.
Language features available in that version will be available for use.
Language features removed in earlier versions, or added in later versions,
will not be available. Note that the language version does not affect
build tags, which are determined by the Go release being used.
```

When you build with go1.12 or go 1.13 without a go directive it modifies the go.mod to adds it to be whatever you're using. Since we support/CI/test building back to go1.11 I thought it'd be worth adding it to be more explicit about the minimum version we are supporting.

samuelkarp: I'm not wholly opposed to it, but I don't know that it's important at this point in our development cycle that we stick to a particular version of Go.  If you wanted to move us fully to 1.13 (and set `go 1.13` here), I think that would be fine as well.

mxpv: Considering Go's [promise of compatibility](https://golang.org/doc/go1compat) I don't think we should force users to use 1.13 unless there are strong reasons for it.

IRCody: I don't see this as making a commitment to a particular version of go other than the commitment we've made in the getting started docs already which say go1.11 or later.

I don't feel super strongly about it either way.

PR #324, opened 2019-11-07
Cleanup around Shutdown() and StopVM() 

kzys:
*Issue #, if available:*

NA

*Description of changes:*

- On rootfs side, we `reboot(2)` after the successful completion of VM agent. That allow us to not killing Firecracker
- On firecracker-contol, `os.RemoveAll()` is added to simplify shim. Since firecracker-control is responsible for creating a shim directory, corresponding removal should be happening on there as well.
- On runtime, `Shutdown()` is now waiting the completion of the VMM process and not killing Firecracker if possible.

### DIFF
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -19,6 +19,7 @@ import (
 	"math"
 	"net"
 	"os"
+	"os/exec"
 	"runtime/debug"
 	"strconv"
 	"strings"
@@ -45,7 +46,6 @@ import (
 	"github.com/gofrs/uuid"
 	ptypes "github.com/gogo/protobuf/types"
 	"github.com/golang/protobuf/ptypes/empty"
-	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
@@ -71,6 +71,7 @@ const (
 	minVsockIOPort          = uint32(11000)
 	firecrackerStartTimeout = 5 * time.Second
 	defaultStopVMTimeout    = 5 * time.Second
+	defaultShutdownTimeout  = 5 * time.Second
 
 	// StartEventName is the topic published to when a VM starts
 	StartEventName = "/firecracker-vm/start"
@@ -548,46 +549,31 @@ func (s *service) mountDrives(requestCtx context.Context, driveMounts []*proto.F
 	return nil
 }
 
-// StopVM will shutdown the firecracker VM and start this shim's shutdown procedure. If the VM has not been
-// created yet and the timeout is hit waiting for it to exist, an error will be returned but the shim will
-// continue to shutdown.
+// StopVM will shutdown the VMM. Unlike Shutdown, this method is exposed to containerd clients.
+// If the VM has not been created yet and the timeout is hit waiting for it to exist, an error will be returned
+// but the shim will continue to shutdown.
 func (s *service) StopVM(requestCtx context.Context, request *proto.StopVMRequest) (_ *empty.Empty, err error) {
+	defer logPanicAndDie(s.logger)
+	s.logger.WithFields(logrus.Fields{"timeout_seconds": request.TimeoutSeconds}).Debug("StopVM")
+
 	timeout := defaultStopVMTimeout
 	if request.TimeoutSeconds > 0 {
 		timeout = time.Duration(request.TimeoutSeconds) * time.Second
 	}
-	timer := time.NewTimer(timeout)
-	defer logPanicAndDie(s.logger)
-	// If something goes wrong here, just shut down ungracefully. This eliminates some scenarios that would result
-	// in the user being unable to shut down the VM.
-	defer func() {
-		if err != nil {
-			s.logger.WithError(err).Error("StopVM error, shim is shutting down ungracefully")
-			s.shimCancel()
-		}
-	}()
 
+	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), timeout)
+	defer shutdownCancel()
+
+	defer s.shimCancel()
 	err = s.waitVMReady()
 	if err != nil {
 		return nil, err
 	}
 
-	shutdownCh := make(chan error)
-	go func() {
-		defer close(shutdownCh)
-		_, err := s.Shutdown(requestCtx, &taskAPI.ShutdownRequest{Now: true})
-		shutdownCh <- err
-	}()
-
-	select {
-	case <-timer.C:
-		return nil, status.Error(codes.DeadlineExceeded, "timed out waiting for VM shutdown")
-	case err = <-shutdownCh:
-		if err != nil {
-			return nil, err
-		}
-		return &empty.Empty{}, nil
+	if err = s.shutdown(requestCtx, shutdownCtx, &taskAPI.ShutdownRequest{Now: true}); err != nil {
+		return nil, err
 	}
+	return &empty.Empty{}, nil
 }
 
 // GetVMInfo returns metadata for the VM being managed by this shim. If the VM has not been created yet, this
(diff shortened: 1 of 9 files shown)


kzys:
I'm going to use 0.19 SDK instead of 90f4bc3f981c86af11251e4a3961f894609638bd.

runtime/service.go
```diff
@@ -1182,7 +1183,7 @@ func (s *service) Shutdown(requestCtx context.Context, req *taskAPI.ShutdownRequ
 
 	err = s.machine.StopVMM()
 	if err != nil {
-		shutdownErr = multierror.Append(shutdownErr, errors.Wrap(err, "failed to gracefully stop VM"))
+		shutdownErr = multierror.Append(shutdownErr, errors.Wrap(err, "failed to stop VM"))
 	}
```
sipsma: It's good to know that `StopVMM` is not implemented as "graceful" shutdown, but can we open a followup issue to update this to actually be graceful (i.e. send `CtlAltDel` and wait a bit)? Based on [this issue](https://github.com/firecracker-microvm/firecracker/issues/1370) it seems like there are risks associated with just doing a `SIGTERM` right away. We'll need to update the vm guest kernel used by our CI system's I'm guessing, but that probably worth the effort here.

runtime/service.go
```diff
@@ -1261,8 +1261,13 @@ func (s *service) monitorVMExit() {
...
-	if waitErr != nil && waitErr != context.Canceled {
-		s.logger.WithError(waitErr).Error("error returned from VM wait")
+	switch waitErr {
+	case nil:
+		s.logger.Error("the VMM has been terminated without cancelling the context")
+	case context.Canceled:
+		s.logger.Info("the VMM has been terminated by cancelling the context")
+	default:
+		s.logger.WithError(waitErr).Error("the VMM has been terminated unexpectedly")
 	}
```
sipsma: For the `context.Canceled` case, the context we're providing is the `shimCtx`, which is only going to be cancelled when the shim is shutting down, in which case the VMM is going to receive a `SIGKILL` if not already dead. So maybe a better way of approaching this would be to change `s.shimCtx` on line 1263 to just be `context.Background()` (which will never be cancelled) and then let `machine.Wait` get unblocked only by the VMM actually exiting (due to the `SIGKILL`). Then we can log whatever exit code is returned by a process exiting via `SIGKILL`.

Given that, I'm imagining something like:
```
switch waitErr {
  case nil:
     s.logger.Info("the VMM process exited with exit code 0")
  default:
   // waitErr will have the exit code or other cleanup error presumably.
   // I don't think we should say "unexpectedly" because it can happen during shim shutdown, at which time it's expected
    s.logger.WithError(waitErr).Error("the VMM process has exited")
}
```

runtime/service.go
kzys: I didn't know the issue. Thanks!

Regarding SendCtrlAltDel, it doesn't work with the kernel we are using, it has been [disabled on ARM](https://github.com/firecracker-microvm/firecracker/pull/1341)since [there is no i8042 device](https://github.com/firecracker-microvm/firecracker/issues/1339).

How about rebooting a guest from our agent? It is slightly complicated from SendCtrlAltDel, but it doesn't have the issues I've mentioned.

sipsma: Oh +100 to doing the reboot from Agent, I like that idea a lot more

runtime/service.go
kzys: The way we use shimCtx is complicated. This method itself is waiting shimCtx but cancelling the context. Shutdown() is also cancelling the shim.

But, not sending SIGTERM to Firecracker may make things simpler. Let me check after adding reboot on the agent.

sipsma: Yeah the `defer s.shimCancel()` is there to handle the case where the VM process exits completely unexpectedly, but running it in other cases where the VM exited because the context was already canceled [is just a no-op](https://golang.org/pkg/context/#CancelFunc).

That's why I think it would simplify things to just use `context.Background` instead of the `shimCtx` for the call to `machine.Wait`. It should result in `Wait` only unblocking when the process has actually exited, which is easier to reason about and results in more helpful logs containing an actual exit code (`context was canceled` doesn't have much meaning to a user reading logs).

I suspect the result of having agent attempt a graceful shutdown will be that it's actually possible for `Wait()` to return a nil error (i.e. the VMM to actually exit with code 0). I think right now because we only terminate the VMM via SIGTERM [it will always exit with a non-zero code](https://unix.stackexchange.com/a/485657), meaning `Wait` should in practice never return a nil error.

kzys:
Let me split the last commit. I think it is too big.

runtime/service.go
```diff
@@ -1178,24 +1178,21 @@ func (s *service) Shutdown(requestCtx context.Context, req *taskAPI.ShutdownRequ
...
+		if err := s.machine.StopVMM(); err != nil {
+			shutdownErr = multierror.Append(shutdownErr, errors.Wrap(err, "failed to shutdown VMM"))
+		}
 	}
 
-	err = os.RemoveAll(s.shimDir.RootPath())
-	if err != nil {
-		shutdownErr = multierror.Append(shutdownErr, errors.Wrapf(err, "failed to remove VM dir %q during shutdown", s.shimDir.RootPath()))
+	if err := s.machine.Wait(context.Background()); err != nil {
+		shutdownErr = multierror.Append(shutdownErr, err)
```
sipsma: I'm little concerned that we are hardcoding an assumption about the rootfs (that agent exiting results in the machine exiting) which is true for our image-builder's image but not for all images that could be used by any user in general (they are free to use their own too).

I think we could address this by having a timeout on the call to `machine.Wait` that was added here (so then we don't get into situations in which `Shutdown` blocks indefinitely). However, that would create some hard-to-follow interactions between here and `StopVM`, which already has a timeout and calls `Shutdown` in a goroutine.

Overall, I'm wondering if it would make sense to migrate the timeout logic present in `StopVM` (setup on lines 555-558) to be enforced within `Shutdown` instead. Because `StopVM` calls `Shutdown`, we could then still end up with the correct timeout logic but without having confusing multiple layers of timeouts being enforced. Let me know what you think.

tools/image-builder/files_debootstrap/etc/systemd/system/firecracker-agent.service
```diff
@@ -15,9 +15,9 @@ Description=Firecracker VM agent
 StartLimitIntervalSec=2
 After=local-fs.target
 Requires=local-fs.target
+SuccessAction=reboot
```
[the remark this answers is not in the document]
kzys: Yes. https://github.com/firecracker-microvm/firecracker/blob/1a325f4aede3539636b00b82d71b638f42cbf7ee/FAQ.md#how-can-i-gracefully-reboot-the-guest-how-can-i-gracefully-poweroff-the-guest

> Running the reboot command in a Linux guest will gracefully bring down the guest system and also bring a graceful end to the Firecracker process.

runtime/service_integ_test.go
```diff
@@ -1104,3 +1104,42 @@ func TestUpdateVMMetadata_Isolated(t *testing.T) {
 	t.Logf("stdout output from task %q: %s", containerName, stdout)
 	assert.Equalf(t, "45", stdout, "container %q did not emit expected stdout", containerName)
 }
+
+func TestStopVM_Isolated(t *testing.T) {
```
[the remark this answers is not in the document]
kzys: Thanks. Will do. May extract the lines as a new helper function.

runtime/service.go
kzys: My thoughts were "Shutdown for graceful shutdown. StopVM for forcefully shutdown", but I agree that the relationship between StopVM and Shutdown is hard to follow. Will move the timeout functionality to Shutdown().

BTW, do you know the reason we expose Shutdown() as a public method (instead of being "shutdown()")?

sipsma: > BTW, do you know the reason we expose Shutdown() as a public method (instead of being "shutdown()")?

Yeah, it's because it's actually part of the API we need to implement as a containerd shim: https://github.com/containerd/containerd/blob/master/runtime/v2/task/shim.proto#L35

kzys: Thanks! And my comment was wrong. Shutdown is also calling StopVMM...

kzys: Addressed in ddbde81ea22749cba5a1ba058049ee0afd5673e1

runtime/service_integ_test.go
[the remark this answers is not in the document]
kzys: Addressed at ebb8db767266f7e4f0f643275b4bcd5a362fbf98

runtime/service.go
```diff
@@ -1149,54 +1135,70 @@ func (s *service) Connect(requestCtx context.Context, req *taskAPI.ConnectReques
...
 	shouldShutdown := req.Now || s.exitAfterAllTasksDeleted && s.taskManager.ShutdownIfEmpty()
 	if !shouldShutdown {
 		return &ptypes.Empty{}, nil
 	}
 
-	// cancel the shim context no matter what, which will result in the VM getting a SIGKILL (if not already
-	// dead from graceful shutdown) and the shim process itself to begin exiting
-	defer s.shimCancel()
-
 	s.logger.Info("stopping the VM")
```
sipsma: nit: this log could probably be put in `shutdown` now

runtime/service.go
```diff
@@ -1149,54 +1135,70 @@ func (s *service) Connect(requestCtx context.Context, req *taskAPI.ConnectReques
...
+		waitErr = status.Error(codes.DeadlineExceeded, "timed out waiting for VM shutdown")
 	}
 
-	if shutdownErr != nil {
-		s.logger.WithError(shutdownErr).Error()
-		return nil, shutdownErr
+	if waitErr != nil {
+		if err := s.machine.StopVMM(); err != nil {
+			return multierror.Append(waitErr, err)
+		}
```
sipsma: I'm wondering if we need to call `StopVMM` at all. AFAIK (correct me if wrong), Firecracker wouldn't have any different behavior between a SIGTERM and a SIGKILL because they don't have a special SIGTERM handler, so I think this should just have the same end effect as cancelling the context (which results in SIGKILL).

If we do want to leave it in, then I think it should not be done when `waitErr` indicates that the machine exited with non-zero code (as the machine is already dead, so no need to SIGTERM it).

runtime/service.go
kzys: Will move.

runtime/service.go
kzys: I wanted to capture an error from StopVMM() by explicitly calling the function. It doesn't have much right now though. https://github.com/firecracker-microvm/firecracker-go-sdk/blob/43d336c3dcbf623d7dec9241f88ca9c0afb44e11/machine.go#L521

runtime/service.go
```diff
@@ -1149,54 +1135,70 @@ func (s *service) Connect(requestCtx context.Context, req *taskAPI.ConnectReques
...
+func (s *service) shutdown(requestCtx, shutdownCtx context.Context, req *taskAPI.ShutdownRequest) error {
+	if _, err := s.agentClient.Shutdown(requestCtx, req); err != nil {
+		return err
 	}
 
-	err = os.RemoveAll(s.shimDir.RootPath())
-	if err != nil {
-		shutdownErr = multierror.Append(shutdownErr, errors.Wrapf(err, "failed to remove VM dir %q during shutdown", s.shimDir.RootPath()))
+	// cancel the shim at the end, in case even the last StopVMM() doesn't work
+	defer s.shimCancel()
```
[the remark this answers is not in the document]
kzys: I'd like to keep shimCancel() here to reduce the number of functions that are aware about the context. Will move that above agentClient.Shutdown() as you suggested.

runtime/service.go
kzys: I think cancelling the context to kill Firecracker depends too much of the current implementation: 1) Firecracker doesn't distinguish SIGTERM/SIGKILL and 2) StopVMM is just SIGTERM. So the last revision still has an explicit StopVMM call.

runtime/service.go
kzys: Moved

runtime/service.go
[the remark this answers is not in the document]
kzys: Moved agentClient.Shutdown(). Still having shimCancel inside shutdown.

Issue #325, opened 2019-11-07
Low entropy causes VM agent to hang indefinitely during start


sipsma:
It appears for unknown reasons a Firecracker VM may not have sufficient entropy during boot, which can cause our VM agent to experience long pauses very early during process start (possibly as the go runtime itself is still starting, before any of our code is actually executing). This results timeouts when the runtime shim attempts to connect to the agent over vsock, resulting in errors like
```
failed to create VM: failed to dial the VM over vsock: context deadline exceeded
```

Because the pause happens so early when agent is starting, it never writes any output to stdout/stderr, so the debug logs show nothing from agent.

[Running low on entropy during boot is an open issue w/ Firecracker](https://github.com/firecracker-microvm/firecracker/issues/663). There are some suggested fixes there (which I have not yet tried), ~but it doesn't appear to have a strong conclusion at this time~. I just saw the `rngd` suggestion in the thread, I will give that a try and update this issue with the results.

I don't know why this suddenly started happening to me, it's on the dev machine I normally use (an `i3.metal` with very similar setup to the CI machines). It may just be entirely random when it starts occurring.

We need to follow that Firecracker thread on solutions to this issue and decide how we can possibly address this for Firecracker-containerd users.

sipsma:
For more background on how I encountered/debugged this:
When testing out a (unrelated) change to agent, I suddenly started getting those `failed to dial the VM over vsock`. I finally resorted to just change the systemd configuration to run strace on agent, which showed the normal output you see but then a sudden pause after these lines:
```
time="2019-11-07T23:04:02.359831030Z" level=debug msg="[    1.693074] strace[733]: [pid   737] getrandom( <unfinished ...>" jailer=runc ociBundlePath=/run/firecracker-containerd/default/test-drive-mount runcBinaryPath=/usr/local/bin/runc runtime=aws.firecracker vmID=test-drive-mount vmm_stream=stdout
time="2019-11-07T23:04:02.362735232Z" level=debug msg="[    1.695840] strace[733]: [pid   738] <... nanosleep resumed> NULL) = 0" jailer=runc ociBundlePath=/run/firecracker-containerd/default/test-drive-mount runcBinaryPath=/usr/local/bin/runc runtime=aws.firecracker vmID=test-drive-mount vmm_stream=stdout
[...]

PR #326, opened 2019-11-11
Add haveged to image-builder's rootfs.

sipsma:
haveged ensures that there is sufficient entropy available to processes running
within the microvm. Without sufficient entropy, it is fairly easy for processes
to get blocked on reading /dev/random or making getrandom() syscalls, including
during boot (which can result in CreateVM calls to fail if the agent process
gets blocked).

haveged was chosen as it enforces no minimum kernel requirements, does not
add CPU requirements (i.e. existence of RDRAND or similar instructions) and is
currently used by Debian for related use cases such as seeding entropy in their
installer.

One other option was "rngd", which has versions that support use of RDRAND.
It was not chosen as RDRAND is not universally trusted or portable. Similarly,
use of the "random.trust_cpu=on" kernel boot parameter was ruled out for now
as it relies on RDRAND and additionally enforces a minimum kernel version of
4.19.

Addresses #325 

Verified the test by running it w/out starting the haveged daemon, which resulted in the container just getting blocked on /dev/random and the test timing out.

### DIFF
--- a/runtime/service_integ_test.go
+++ b/runtime/service_integ_test.go
@@ -1104,3 +1104,116 @@ func TestUpdateVMMetadata_Isolated(t *testing.T) {
 	t.Logf("stdout output from task %q: %s", containerName, stdout)
 	assert.Equalf(t, "45", stdout, "container %q did not emit expected stdout", containerName)
 }
+
+// TestRandomness validates that there is a reasonable amount of entropy available to the VM and thus
+// randomness available to containers (test reads about 2.5MB from /dev/random w/ an overall test
+// timeout of 60 seconds). It also validates that the quality of the randomness passes the rngtest
+// utility's suite.
+func TestRandomness_Isolated(t *testing.T) {
+	prepareIntegTest(t)
+
+	ctx, cancel := context.WithTimeout(namespaces.WithNamespace(context.Background(), defaultNamespace), 60*time.Second)
+	defer cancel()
+
+	client, err := containerd.New(containerdSockPath, containerd.WithDefaultRuntime(firecrackerRuntime))
+	require.NoError(t, err, "unable to create client to containerd service at %s, is containerd running?", containerdSockPath)
+	defer client.Close()
+
+	image, err := alpineImage(ctx, client, defaultSnapshotterName())
+	require.NoError(t, err, "failed to get alpine image")
+	containerName := "test-entropy"
+
+	const blockCount = 1024
+	ddContainer, err := client.NewContainer(ctx,
+		containerName,
+		containerd.WithSnapshotter(defaultSnapshotterName()),
+		containerd.WithNewSnapshot("test-entropy-snapshot", image),
+		containerd.WithNewSpec(
+			oci.WithDefaultUnixDevices,
+			// Use blocksize of 2500 as rngtest consumes data in blocks of 2500 bytes.
+			oci.WithProcessArgs("/bin/dd", "iflag=fullblock", "if=/dev/random", "of=/dev/stdout", "bs=2500",
+				fmt.Sprintf("count=%d", blockCount)),
+		),
+	)
+	require.NoError(t, err, "failed to create container %s", containerName)
+
+	// rngtest is a utility to "check the randomness of data using FIPS 140-2 tests", installed as part of
+	// the container image this test is running in. We pipe the output from "dd if=/dev/random" to rngtest
+	// to validate the quality of the randomness inside the VM.
+	// TODO It would be conceptually simpler to just run rngtest inside the container in the VM, but
+	// doing so would require some updates to our test infrastructure to support custom-built container
+	// images running in VMs (right now it's only feasible to use publicly available container images).
+	// Right now, it's instead run as a subprocess of this test outside the VM.
+	var rngtestStdout bytes.Buffer
+	var rngtestStderr bytes.Buffer
+	rngtestCmd := exec.CommandContext(ctx, "rngtest",
+		// we set this to 1 less than the number of blocks read by dd above to account for the fact that
+		// the first 32 bits read by rngtest are not used for the tests themselves
+		fmt.Sprintf("--blockcount=%d", blockCount-1),
+	)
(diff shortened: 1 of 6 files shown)


runtime/service_integ_test.go
```diff
@@ -1104,3 +1104,91 @@ func TestUpdateVMMetadata_Isolated(t *testing.T) {
...
+	err = task.Start(ctx)
+	require.NoError(t, err, "failed to start task for dd container")
+
+	err = rngtestCmd.Start()
+	require.NoError(t, err, "failed to start rngtest")
+
+	select {
+	case exitStatus := <-exitCh:
+		assert.NoError(t, exitStatus.Error(), "failed to retrieve exitStatus")
+		assert.Equal(t, uint32(0), exitStatus.ExitCode())
```
mxpv: nit: we can get rid of cast with `EqualValues`

runtime/service_integ_test.go
```diff
@@ -1104,3 +1104,91 @@ func TestUpdateVMMetadata_Isolated(t *testing.T) {
...
+		}
+
+		t.Logf("stderr output from dd:\n %s", ddStderr.String())
+	case <-ctx.Done():
+		require.Fail(t, "context cancelled",
+			"context cancelled while waiting for dd container to exit (is it blocked on reading /dev/random?), err: %v", ctx.Err())
+	}
+	rngtestStdin.Close()
+
+	// rngtest will exit non-zero if any blocks fail its randomness tests
```
mxpv: If I get this properly, rngtest will return `1 if no errors happen, but at least one block fails the FIPS tests`. Can this potentially be a flaky test that we'll have to retry? May be it worth to define some success threshold, like no more than 1% or something.

sipsma: Good question, I ran a test where I locally increased the number of blocks by a factor of 1024 and got 27 block failures out of 32863 total tested. So, presuming those are all false-negatives and that I'm remembering how to math, if we test 32 blocks in a single test case, the probability of there being 1 false-negative is [~2.6%](https://www.wolframalpha.com/input/?i=%2832+choose+1%29+*+%2827%2F32863%29%5E1*%281-27%2F32863%29%5E%2832-1%29) and the probability of there being 2 false-negatives is [~0.03%](https://www.wolframalpha.com/input/?i=%2832+choose+2%29+*+%2827%2F32863%29%5E2*%281-27%2F32863%29%5E%2832-2%29).

2.6% is a little high, I agree, so I'll update the test case to look at the output and check how many failures there were in case of a non-zero exit code, only failing if there was more than 2.

sipsma: Just as a note, I ended up actually testing 1023 blocks (not just 32) because I fixed a separate problem that allowed me to actually test the expected number of blocks (had to specify `iflag=fullblock` in the dd command). So re-running the calculations I went with a failure tolerance of 4 blocks as [the chances of 5 block failures drops down to ~0.1%](https://www.wolframalpha.com/input/?i=%281023+choose+5%29+*+%2827%2F32863%29%5E5*%281-27%2F32863%29%5E%281023-5%29)

tools/image-builder/files_debootstrap/etc/systemd/system/haveged.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
```
nmeyerhans: Newer version of the haveged package provide their own unit file. You might consider borrowing the file from buster, or at least some of its settings.  https://salsa.debian.org/debian/haveged/blob/buster/debian/haveged.service

We should upgrade the VM image to buster one of these days...

runtime/service_integ_test.go
```diff
@@ -1104,3 +1104,116 @@ func TestUpdateVMMetadata_Isolated(t *testing.T) {
 	t.Logf("stdout output from task %q: %s", containerName, stdout)
 	assert.Equalf(t, "45", stdout, "container %q did not emit expected stdout", containerName)
 }
+
+// TestRandomness validates that there is a reasonable amount of entropy available to the VM and thus
+// randomness available to containers (test reads about 2.5MB from /dev/random w/ an overall test
```
nmeyerhans: That's a lot of data to pull from /dev/random. How did you arrive at that number?

tools/image-builder/files_debootstrap/etc/systemd/system/haveged.service
kzys: buster upgrade is coming... https://github.com/firecracker-microvm/firecracker-containerd/pull/324/commits/d56b94b91645a778792bd716d301731223db2643

runtime/service_integ_test.go
sipsma: I wanted to provide a fair number of blocks to `rngtest`, so I started at this arbitrary number (1024 blocks). I also wanted to read enough so that we can easily catch the container getting blocked on reading `/dev/random` when the entropy isn't being seeded by `haveged`.

When I ran this test w/out `haveged`, I did indeed see the container get blocked and thus the test timeout+fail, so that worked as expected. When `haveged` is present, the test runs relatively quickly so I didn't see any reason to change the arbitrary number I started at. Is there any reason to that you see?

tools/image-builder/files_debootstrap/etc/systemd/system/haveged.service
sipsma: Will update

sipsma: Updated, only difference between the service files now is I inlined the args to the daemon instead of maintaining a separate env file (just seems simpler for us) and left out `apparmor.service systemd-random-seed.service systemd-tmpfiles-setup.service` from the `After=` parameters as I don't see any of those services under image-builder's rootfs

runtime/service_integ_test.go
nmeyerhans: Nope. It still feels like a lot to me. Applications would typically not read more than a few kB at most. But if the test passes with 2.5 MB, then let's go with it.

PR #330, opened 2019-11-14
Run "make test" on Travis CI

kzys:
"make" doesn't compile *_test.go, but "make test" does.

121afa34449077eef99e533712d2a3cce501612a introduced
ProcessState.ExitCode which is not available in Go 1.11.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,3 +1,6 @@
+run:
+  timeout: 5m
+
 linters:
   disable-all: true
   enable:
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,5 +13,7 @@ install:
   - make deps
 
 script:
+  # "make" must be happening before "make lint" to download dependencies
   - make
   - make lint
+  - make test
--- a/runtime/service_integ_test.go
+++ b/runtime/service_integ_test.go
@@ -1105,6 +1105,13 @@ func TestUpdateVMMetadata_Isolated(t *testing.T) {
 	assert.Equalf(t, "45", stdout, "container %q did not emit expected stdout", containerName)
 }
 
+func exitCode(err *exec.ExitError) int {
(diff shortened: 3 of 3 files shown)


nmeyerhans:
While you're in here, it's probably worth fixing the `make lint` failures in travis...

sipsma:
Good catch on that test using that new API, I really, really wish the go docs made any effort to say that a given api was introduced in a particular version (let me know if there's some way I'm not aware of).... But that is what it is.

I can submit a separate PR to check the exit code [the old way](https://github.com/golang/go/issues/26539) so we don't have a 1.12 requirement.

For the travis build, we already run the unit tests in buildkite, so I'm wondering if all we should do in travis is `go test -c` which will compile the tests into a binary but not run them: https://golang.org/cmd/go/#hdr-Test_packages

If that doesn't work it's not the end of the world to just run the unit tests in travis too though.

kzys:
Well, they have "1.12" on the right :)

https://golang.org/pkg/os/#ProcessState.ExitCode

Actually I would rather just drop Go 1.11 support. Firecracker Go SDK may need to support older versions, but firecracker-containerd doesn't have to, I think.

sipsma:
> Well, they have "1.12" on the right :)

Haha! My widescreen monitor is so widescreen I have to turn my neck to look at the right side. Good to know though, thanks!

sipsma:
> Actually I would rather just drop Go 1.11 support. Firecracker Go SDK may need to support older versions, but firecracker-containerd doesn't have to, I think.

Given the fix to this issue is very straightforward, I feel like we may as well not drop go 1.11 if we don't have to. Debian buster, for example, is still on go 1.11 by default afaik, so it might make it easier for some users to build if we don't require 1.12+

kzys:
You are right. Debian... https://packages.debian.org/buster/golang

Will update the test in this PR then.

PR #332, opened 2019-11-15
Support rate-limiters and read-only for drive mounts.

### LINKED ISSUE #296: Support RateLimiters and ReadOnly for DriveMounts

sipsma:

Resolves #296

There is an unrelated failing integ test that seems like it will be fixed by #331


proto/types.proto
```diff
@@ -144,7 +144,14 @@ message FirecrackerDriveMount {
...
 
-  // TODO support for ratelimiters and read-only drives (issue #296)
+  // (Optional) RateLimiter configuration that will be applied to the
+  // backing-drive for the VM's rootfs
+  FirecrackerRateLimiter RateLimiter = 5;
+
+  // (Optional) If set to true, IsReadOnly results in the backing file for the
+  // drive being opened as read-only by the Firecracker VMM on the host, preventing
+  // any writes to the image from within the guest.
+  bool IsReadOnly = 6;
```
kzys: Can we negate the flag to IsWritable? Mostly because FirecrackerRootDrive has that.

runtime/drive_handler.go
```diff
@@ -15,125 +15,194 @@ package main
...
+type MountableStubDrive interface {
+	PatchAndMount(
+		requestCtx context.Context,
+		machine firecracker.MachineIface,
+		driveMounter drivemount.DriveMounterService,
+	) error
+}
 
-	return paths, nil
+func stubPathToDriveID(stubPath string) string {
```
kzys: Can you explain which characters are problematic in a comment.

proto/types.proto
sipsma: Good call, updated. As a result I needed to add a little more functionality though because Linux will default a mount to "rw" if not otherwise specified. So this meant that if a DriveMount was read-only by default then the user would get an error unless they explicitly set "ro" in their mount options, which kind of sucks. So I updated the implementation to check that the `IsWritable` field matches what was specified in the mount options or, if there was no ro/rw option specified, explicitly sets it on behalf of the user.

runtime/drive_handler.go
sipsma: Done

kzys: Base64 will remove "/" but may contain "-" in the output. We could use Base32 instead.

sipsma: Good catch! I didn't know that. Switched to Base32 and also switched to using just the base name of the stub file (which is unique per shim/vm anyways) so the length of the drive ids doesn't get too outrageous.

runtime/drive_handler.go
```diff
@@ -15,125 +15,227 @@ package main
...
-// stubDrivesOpt is used to make and modify changes to the stub drives.
-type stubDrivesOpt func(stubDrives []models.Drive) error
+// StubDriveHandler manages a set of stub drives. It currently only supports reserving
+// one of the drives from its set.
+// In the future, it may be expanded to also support recycling a drive to be used again
+// for a different mount.
+type StubDriveHandler struct {
+	freeDrives []*stubDrive
+	// map of id -> stub drive being used by that task
+	usedDrives map[string]*stubDrive
```
xibz: What is the purpose of this field? I don't really see it used anywhere other than being set

runtime/drive_handler.go
xibz: Can we add a test that has an invalid `stubPath` just to ensure that this doesn't regress?

runtime/drive_handler.go
```diff
@@ -15,125 +15,227 @@ package main
...
+	h.usedDrives[id] = freeDrive
+	return freeDrive.withMountConfig(
+		hostPath,
+		vmPath,
+		filesystemType,
+		options,
+	), nil
+}
+
+// CreateDriveMountStubs creates a set of MountableStubDrives from the provided DriveMount configs.
```
xibz: We may want to document that the rate limiters and permissions need to be known upfront

runtime/drive_handler.go
sipsma: > Can we add a test that has an invalid stubPath just to ensure that this doesn't regress and is being used properly?

I'm not sure I follow how we could create an invalid stubPath as the implementation of our code now shouldn't have any invalid values for a stubPath. Would the test create a drive id with an invalid character and then provide that directly to the firecracker API? That wouldn't really be a test of our code.

The only thing I can think of here would be to do some kind of general fuzz testing on our api to try to find inputs that create errors, but that seems like a much, much larger effort outside of scope here.

runtime/drive_handler.go
sipsma: I'm just laying some groundwork for the implementation of #243 to build off of, which I am personally okay with since it's not complicating anything at the moment. If there's disagreement with including it even though it technically doesn't need to exist yet, then I can remove for now, let me know.

runtime/drive_handler.go
sipsma: Good call, fixed

runtime/drive_handler.go
xibz: > Would the test create a drive id with an invalid character and then provide that directly to the firecracker API? That wouldn't really be a test of our code.

Yea, I was thinking this, just to ensure that it is encoding it properly. You should then be able to read the stub drive and compare the base32 value?

runtime/drive_handler.go
xibz: Ah, okay. I was just making sure it was there intentionally. I am okay with leaving it.

runtime/drive_handler.go
sipsma: Ohh you just mean this function in isolation, I see, I was thinking about some way to get an invalid stub path passed to this function. Yeah I will add a test that does a regex check on the output of this function for some paths with invalid chars.

sipsma: Updated with that test

Issue #313, opened 2019-10-28
Simplify test-image Dockerfile


sipsma:
[In reference to the images we build to run unit+integ-tests](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/tools/docker/Dockerfile)
* The multi-stage build no longer has any benefit as the previous parallel stages have been migrated out to separate Dockerfiles (image-builder and firecracker-builder).
* In general, a better approach would be to use that of image-builder and firecracker-builder, which is to just have a simple Dockerfile with the host-dependencies needed to perform the build (i.e. golang at some version and few other system libraries) and then define all the logic of the build as steps in the main Makefile that run the build via docker run with the directory of the host bind mounted in. That allows the host filesystem to serve as the "cache" and normal Makefile logic to determine when to re-run a step.
   * One benefit of this would be that modifying an integ test case would no longer require require rebuilding the entire test-image, would might save multiple minutes each dev-cycle.
   * For caching demo-network binaries specifically I think we'd also need to first download to BINPATH and then install to CNI_BIN_PATH so we can use the BINPATH as the cache and only copy binaries to the image's /opt/cni/bin.

PR #333, opened 2019-11-15
Create a small Docker image for running integration tests

kzys:
*Issue #, if available:*

#313

*Description of changes:*

As described in #313, firecracker-containerd-integ-test is copying the entire source directory to the container in the very early stage, which makes incremental build slower. This PR changes the strategy to

- Build binaries without Docker
- Copy the binaries to a Docker image later
- Only use go.mod and go.sum to download dependencies inside a container

That will make incremental build much faster.

### DIFF
--- a/runtime/drive_handler_test.go
+++ b/runtime/drive_handler_test.go
@@ -33,7 +33,7 @@ import (
 )
 
 func TestStubDriveHandler(t *testing.T) {
-	tempPath, err := ioutil.TempDir("./", "TestStubDriveHandler")
+	tempPath, err := ioutil.TempDir("", "TestStubDriveHandler")
 	require.NoError(t, err)
 	defer func() {
 		os.RemoveAll(tempPath)
--- a/tools/docker/Dockerfile.integ-test
+++ b/tools/docker/Dockerfile.integ-test
@@ -1,0 +1,74 @@
+# Test image that starts up containerd and the naive snapshotter. The default CMD will drop to a bash shell. Overrides
+# to CMD will be provided appended to /bin/bash -c
+FROM golang:1.13-stretch
+ENV PATH="/bin:/usr/bin:/usr/local/bin:/sbin:/usr/sbin:/usr/local/sbin:/usr/local/go/bin" \
+	FICD_LOG_DIR="/var/log/firecracker-containerd-test"\
+	DEBIAN_FRONTEND="noninteractive"
+ENV FICD_SNAPSHOTTER="naive" \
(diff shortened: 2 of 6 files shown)


kzys:
I'm going to move "runtime" tests as well, hopefully in this PR.

runtime/drive_handler_test.go
```diff
@@ -33,7 +33,7 @@ import (
 )
 
 func TestStubDriveHandler(t *testing.T) {
-	tempPath, err := ioutil.TempDir("./", "TestStubDriveHandler")
+	tempPath, err := ioutil.TempDir("", "TestStubDriveHandler")
```
IRCody: Would we want to keep this behavior if people run the tests manually?

tools/docker/Dockerfile.integ-test
```diff
@@ -0,0 +1,74 @@
...
+    /opt/cni/bin \
+    /firecracker-containerd \
+    /var/lib/firecracker-containerd/naive \
+    ${FICD_LOG_DIR}
+
+RUN curl --silent --show-error --retry 3 --max-time 30 --output default-vmlinux.bin \
+	"https://s3.amazonaws.com/spec.ccfc.min/img/hello/kernel/hello-vmlinux.bin" \
+	&& echo "882fa465c43ab7d92e31bd4167da3ad6a82cb9230f9b0016176df597c6014cef default-vmlinux.bin" | sha256sum -c - \
+	&& chmod 0444 default-vmlinux.bin \
+    && mv default-vmlinux.bin /var/lib/firecracker-containerd/runtime/default-vmlinux.bin
```
IRCody: nit: line up the &&?

kzys: Will do.

runtime/drive_handler_test.go
kzys: I was wondering why we are using "./". That would be the reason.

How about asking people to manually set TMPDIR? Seems Go honors that.

* https://golang.org/pkg/io/ioutil/#TempDir
* https://golang.org/pkg/os/#TempDir

kzys:
27c842ad7d9b1a7f40a567cdb7370e31997bf016 should be a separate commit. I'm fine about squashing other three into one.

runtime/drive_handler_test.go
IRCody: Good point. I think it should be fine then.

IRCody:
That SGTM.

PR #334, opened 2019-11-18
Add numa cpuset support

xibz:
This change adds support for specifying a set of cpu and memory nodes
during the CreateVM call. This will run the jailed process on those set
of cpus and restrict access to the specified nodes. This change also
adds the cpuset.Builder which allows for easy creation of cpus and mems
strings

*Issue #, if available:*

*Description of changes:*


proto/firecracker.proto
```diff
@@ -81,5 +81,12 @@ message GetVMMetadataResponse {
 
 message JailerConfig {
     string NetNS = 1;
+
+	// The list of CPUs to run the jailed application on. This field can only
+	// be used on numa architectures
+    string CPUs = 2;
```
sipsma: Why aren’t we using “repeated string CPUs” if it is a list? Also, is there a more descriptive name than CPUs? I’m not familiar with the underlying numa api but does it call them anything more specific, even just like CPUID or something?

Same basic comment applies to “Mems” below

xibz: This is a format that most people will be familiar with when using numa nodes which is a string to represent the CPU set and memory node sets. That's why I added the builder to help people that aren't familiar with the format.

I will update the docs to point to the format and add a little more context.

sipsma: Thanks! Makes much more sense now

runtime/cpuset/cpuset_builder.go
```diff
@@ -0,0 +1,119 @@
...
+		str += fmt.Sprintf("%s%d", sep, elem)
+	}
+
+	for _, r := range ranges {
+		sep := ","
+		if len(str) == 0 {
+			sep = ""
+		}
+
+		str += fmt.Sprintf("%s%d-%d", sep, r.min, r.max)
```
sipsma: nit: `_range` seems like a good candidate for a `String()` method

runtime/cpuset/cpuset_builder.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// AddMem adds a memory node which limit where the cpus can allocate memory
+func (b Builder) AddMem(mem int) Builder {
+	b.mems = append(b.mems, mem)
+	return b
+}
+
+// AddMemRange adds a range of memory nodes by utilizing the minimum node to
+// use to the maximum node to use.
```
nmeyerhans: I'm not quite sure what you're trying to say here.

runtime/cpuset/cpuset_builder.go
```diff
@@ -0,0 +1,119 @@
...
+func (b Builder) AddCPURange(min, max int) Builder {
+	b.cpuRanges = append(b.cpuRanges, _range{
+		min: min,
+		max: max,
+	})
+
+	return b
+}
+
+// AddMem adds a memory node which limit where the cpus can allocate memory
```
nmeyerhans: typo: limit**s**

runtime/cpuset/cpuset_builder.go
```diff
@@ -0,0 +1,119 @@
...
+// permissions and limitations under the License.
+
+package cpuset
+
+import (
+	"fmt"
+)
+
+// Builder will allow building of cpuset fields
+// http://man7.org/linux/man-pages/man7/cpuset.7.html
```
nmeyerhans: I think it makes more sense to link to the cpuset docs near the definition of the CPUSet structure, rather than this builder. Or maybe in some introductory text in the file header in which we explain what cpusets are and how we use them.

runtime/jailer_integ_test.go
```diff
@@ -79,3 +80,17 @@ func testJailer(t *testing.T, jailerConfig *proto.JailerConfig) {
 		require.NoError(err, "failed to delete a container")
 	}()
 }
+
+func TestJailerWithNumaNode_Isolated(t *testing.T) {
+	prepareIntegTest(t, withJailer())
+
+	t.Run("Jailer NumaNode", func(t *testing.T) {
+		b := cpuset.Builder{}
+		cset := b.AddCPU(0).AddMem(0).Build()
```
nmeyerhans: This will work even on non-NUMA hardware, right?

runtime/runc_jailer.go
```diff
@@ -41,38 +41,33 @@ const (
...
-func newRuncJailer(
-	ctx context.Context,
-	logger *logrus.Entry,
-	vmID string,
-	ociBundlePath string,
-	runcBinaryPath string,
-	uid uint32,
-	gid uint32) (*runcJailer, error) {
-	l := logger.WithField("ociBundlePath", ociBundlePath).WithField("runcBinaryPath", runcBinaryPath)
+type runcJailerConfig struct {
```
nmeyerhans: One general suggestion for the future is to consider making these changes (the introduction and use of the runcJailerConfig struct) in a separate commit. Still part of the same PR. Introduce the structure in an earlier commit, and refactor existing code to use it. Then add support for the cpuset fields in a second commit. It makes it very clear exactly why a given line of code is changing, and it makes it easier to debug and bisect the code if we ever need to.

Just something to think about for the future. Don't worry at all about changing what you've already done here.

runtime/cpuset/cpuset_builder.go
xibz: Good call out. Will go ahead and do that.

runtime/jailer_integ_test.go
xibz: I cannot find an instance to test this on that has numa support disabled. I will try to test this on my local machine by disabling numa support all together in the BIOs.

runtime/runc_jailer.go
xibz: Damn, that's smart haha. Will definitely do that in the future.

runtime/cpuset/cpuset_builder.go
```diff
@@ -0,0 +1,119 @@
...
+
+	return CPUSet{
+		cpus: cpus,
+		mems: mems,
+	}
+}
+
+func stringify(elems []int, ranges []_range) string {
+	str := ""
+	for _, elem := range elems {
```
samuelkarp: Any reason for doing this instead of using `strings.Join`?

runtime/cpuset/cpuset_builder.go
```diff
@@ -0,0 +1,119 @@
...
+		sep := ","
+		if len(str) == 0 {
+			sep = ""
+		}
+
+		str += fmt.Sprintf("%s%d", sep, elem)
+	}
+
+	for _, r := range ranges {
```
samuelkarp: Same question here.

runtime/jailer_integ_test.go
nmeyerhans: It should. I'm mainly asking because I wonder if we should give the test a more accurate name. Firecracker uses "NUMA" all over the place, even on non-NUMA systems, so it could get confusing anyway. Up to you...

runtime/cpuset/cpuset_builder_test.go
```diff
@@ -0,0 +1,166 @@
...
+	cases := []struct {
+		name           string
+		cpus           []int
+		mems           []int
+		cpuRanges      []_range
+		memRanges      []_range
+		expectedCPUSet CPUSet
+	}{
+		{
+			name: "single cpu only",
```
IRCody: Worth adding cases for empty cpus/empty cpuRanges?

runtime/cpuset/cpuset_builder.go
xibz: Gah, good call out. Just slipped the mind.

runtime/cpuset/cpuset_builder_test.go
xibz: Done

runtime/jailer_integ_test.go
xibz: Renamed to `TestJailerCPUSet_Isolated`

runtime/cpuset/cpuset_builder.go
xibz: Fixed

runtime/cpuset/cpuset_builder.go
xibz: Fixed

runtime/cpuset/cpuset_builder.go
xibz: Done

runtime/jailer_integ_test.go
```diff
@@ -79,3 +80,17 @@ func testJailer(t *testing.T, jailerConfig *proto.JailerConfig) {
 		require.NoError(err, "failed to delete a container")
 	}()
 }
+
+func TestJailerCPUSet_Isolated(t *testing.T) {
+	prepareIntegTest(t, withJailer())
+
+	t.Run("Jailer NumaNode", func(t *testing.T) {
```
nmeyerhans: Can you update the name to reflect the new function name?

xibz: Done!

runtime/cpuset/cpuset_builder.go
samuelkarp: @xibz Can you update this to use `strings.Join`?

xibz: This returns a string using Join. Was there something else you had in mind? Cause I think this is probably the simplest way to go?

samuelkarp: @xibz My apologies, I misread this.  I might use `strconv.Itoa` instead of `fmt.Sprintf`, but that's really minor.

xibz: Ah cool, and no worries! I'll go ahead and fix the `fmt.Sprintf` issue. Thanks!

PR #339, opened 2019-11-21
Support for container stub drive reuse

Zyqsempai:

Issue #243 

Added possibility to reuse stub drives of just deleted containers:
- The `Unmount()` method was added to `agent/drive_handler` to support unmounting used stub drives.
- The `Release()` method was added to aggregate all steps we need to release used stub.
- Modified task `Delete()` method to use new `Release()` method.

### DIFF
--- a/agent/drive_handler.go
+++ b/agent/drive_handler.go
@@ -177,22 +177,11 @@ func (dh driveHandler) MountDrive(ctx context.Context, req *drivemount.MountDriv
 	logger = logger.WithField("drive_path", drive.Path())
 
 	// Do a basic check that we won't be mounting over any important system directories
-	resolvedDest, err := evalAnySymlinks(req.DestinationPath)
-	if err != nil {
-		return nil, errors.Wrapf(err,
-			"failed to evaluate any symlinks in drive mount destination %q", req.DestinationPath)
+	if err := isSystemDir(req.DestinationPath); err != nil {
+		return nil, err
 	}
 
-	for _, systemDir := range bannedSystemDirs {
-		if isOrUnderDir(resolvedDest, systemDir) {
-			return nil, errors.Errorf(
-				"drive mount destination %q resolves to path %q under banned system directory %q",
-				req.DestinationPath, resolvedDest, systemDir,
-			)
-		}
-	}
-
-	err = os.MkdirAll(req.DestinationPath, 0700)
+	err := os.MkdirAll(req.DestinationPath, 0700)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to create drive mount destination %q", req.DestinationPath)
 	}
@@ -227,6 +216,40 @@ func (dh driveHandler) MountDrive(ctx context.Context, req *drivemount.MountDriv
 
 	return nil, errors.Errorf("exhausted retries mounting drive from %q to %q",
 		drive.Path(), req.DestinationPath)
+}
+
+func (dh driveHandler) UnmountDrive(ctx context.Context, req *drivemount.UnmountDriveRequest) (*empty.Empty, error) {
+	drive, ok := dh.GetDrive(req.DriveID)
+	if !ok {
+		return nil, fmt.Errorf("drive %q could not be found", req.DriveID)
+	}
+
+	err := mount.Unmount(drive.Path(), 0)
+	if err == nil {
+		return &empty.Empty{}, nil
+	}
+
+	return nil, errors.Errorf("failed to unmount the drive %q",
+		drive.Path())
+}
+
+func isSystemDir(path string) error {
+	resolvedDest, err := evalAnySymlinks(path)
+	if err != nil {
+		return errors.Wrapf(err,
+			"failed to evaluate any symlinks in drive ummount destination %q", path)
+	}
+
+	for _, systemDir := range bannedSystemDirs {
+		if isOrUnderDir(resolvedDest, systemDir) {
+			return errors.Errorf(
+				"drive mount destination %q resolves to path %q under banned system directory %q",
+				path, resolvedDest, systemDir,
+			)
+		}
+	}
+
+	return nil
 }
 
 // evalAnySymlinks is similar to filepath.EvalSymlinks, except it will not return an error if part of the
--- a/runtime/drive_handler.go
+++ b/runtime/drive_handler.go
@@ -96,32 +96,79 @@ type StubDriveHandler struct {
 // Reserve pops a unused stub drive and returns a MountableStubDrive that can be
 // mounted with the provided options as the patched drive information.
 func (h *StubDriveHandler) Reserve(
+	requestCtx context.Context,
 	id string,
 	hostPath string,
 	vmPath string,
 	filesystemType string,
 	options []string,
-) (MountableStubDrive, error) {
+	driveMounter drivemount.DriveMounterService,
+	machine firecracker.MachineIface,
+) error {
 	h.mu.Lock()
 	defer h.mu.Unlock()
 	if len(h.freeDrives) == 0 {
-		return nil, ErrDrivesExhausted
+		return ErrDrivesExhausted
+	}
+	if _, ok := h.usedDrives[id]; ok {
+		// This case means that driver wasn't released or removed properly
+		return fmt.Errorf("drive with ID %s already in use, a previous attempt to remove it may have failed", id)
 	}
 
 	freeDrive := h.freeDrives[0]
 	options, err := setReadWriteOptions(options, freeDrive.driveMount.IsWritable)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	h.freeDrives = h.freeDrives[1:]
-	h.usedDrives[id] = freeDrive
-	return freeDrive.withMountConfig(
+	stubDrive := freeDrive.withMountConfig(
 		hostPath,
 		vmPath,
 		filesystemType,
 		options,
-	), nil
+	)
+
+	err = stubDrive.PatchAndMount(requestCtx, machine, driveMounter)
+	if err != nil {
+		err = errors.Wrapf(err, "failed to mount drive inside vm")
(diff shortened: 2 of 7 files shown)


Zyqsempai:
@sipsma PTAL and maybe you will have an idea how to fix test, it looks like `hostPath` is missing, but I have no clue how to set it in that test.

agent/drive_handler.go
```diff
@@ -177,22 +177,11 @@ func (dh driveHandler) MountDrive(ctx context.Context, req *drivemount.MountDriv
...
-	}
-
-	for _, systemDir := range bannedSystemDirs {
-		if isOrUnderDir(resolvedDest, systemDir) {
-			return nil, errors.Errorf(
-				"drive mount destination %q resolves to path %q under banned system directory %q",
-				req.DestinationPath, resolvedDest, systemDir,
-			)
-		}
+	if err := checkImportantDirs(req.DestinationPath); err != nil {
```
kzys: How about renaming the function to isImportantDir() or isSystemDir()? `Check` is slightly ambiguous about what true/false means.

agent/drive_handler.go
```diff
@@ -229,6 +218,45 @@ func (dh driveHandler) MountDrive(ctx context.Context, req *drivemount.MountDriv
...
+func (dh driveHandler) UnmountDrive(ctx context.Context, req *drivemount.UnmountDriveRequest) (*empty.Empty, error) {
+	drive, ok := dh.GetDrive(req.DriveID)
+	if !ok {
+		return nil, fmt.Errorf("drive %q could not be found", req.DriveID)
+	}
+
+	// Do a basic check that we won't be mounting over any important system directories
+	if err := checkImportantDirs(req.DestinationPath); err != nil {
+		return nil, err
+	}
```
kzys: Does it need to check the path here?

`UnmoutDriveRequest { DriveID = (some valid ID), DestinationPath = (some invalid path) }` will be rejected by this check, but it wouldn't harm anything since the valid ID is anyway pointing the valid (non-important) directory.

runtime/drive_handler.go
```diff
@@ -124,6 +124,39 @@ func (h *StubDriveHandler) Reserve(
 	), nil
 }
 
+// Release unmounts stub drive of just deleted container
+// and pushes just released drive to freeDrives
+func (h *StubDriveHandler) Release(
+	requestCtx context.Context,
+	taskID string,
```
kzys: taskID -> id? The handler is not aware about tasks so far.

runtime/drive_handler.go
```diff
@@ -124,6 +124,39 @@ func (h *StubDriveHandler) Reserve(
...
+	machine firecracker.MachineIface,
+) error {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	drive, ok := h.usedDrives[taskID]
+	if !ok {
+		return errors.Errorf("container %s drive wasn't found", taskID)
+	}
+
+	_, err := driveMounter.UnmountDrive(requestCtx, &drivemount.UnmountDriveRequest{
```
kzys: So far, the handler was only doing bookkeeping but now it actually modify a microVM. I'm unsure this is the direction we want to go. Want to know @sipsma and others' views.

runtime/drive_handler.go
```diff
@@ -124,6 +124,39 @@ func (h *StubDriveHandler) Reserve(
 	), nil
 }
 
+// Release unmounts stub drive of just deleted container
```
sipsma: I'm thinking we may also need to update `Reserve` to check if `id` is already in `h.usedDrives` and return an error if so. If that situation happened, it would potentially mean we entered some inconsistent state where a task with the given `id` was deleted according to containerd but the backing drive failed to get `Release`'d

agent/drive_handler.go
sipsma: +1 I'm not sure there's a need to check this here too as the mount has already occurred.

runtime/drive_handler.go
sipsma: In practice, what happens now is [you `Reserve` the drive and then immediately call `PatchAndMount`, which takes the same `driveMounter` and `machine` interfaces](https://github.com/firecracker-microvm/firecracker-containerd/blob/802c35dac582ce5543ba6df7e15c215a830c5984/runtime/service.go#L813-L826). So whether you provide `driveMounter` and `machine` to the stub drive handler or a separate method is kind of arbitrary to me.

However, I do think it may be a good idea to leave `Release` as is and update `Reserve` to also take in `driveMounter` and `machine` and directly make the `PatchAndMount` call within its implementation. That not only makes the API a bit more "symmetrical", it also means that we can put the drive in `usedDrives` only when `PatchAndMount` actually succeeds, which reduces the cases where we can enter a state where a stub drive can't be released.

runtime/service.go
```diff
@@ -928,23 +929,29 @@ func (s *service) Delete(requestCtx context.Context, req *taskAPI.DeleteRequest)
 	if req.ExecID != "" {
 		return resp, nil
 	}
+	var result *multierror.Error
+	// Trying to release stub drive for further reuse
+	err = s.containerStubHandler.Release(requestCtx, req.ID, s.driveMountClient, s.machine)
+	if err != nil {
+		return nil, multierror.Append(result, errors.Wrapf(err, "failed to release stub drive for container: %s", req.ID))
```
sipsma: The way I think we want to use `multierror` here is to just store the error returned by `Release` rather than return it right away. So something more like this:
```
err = s.containerStubHandler.Release(...)
if err != nil {
    result = multierror.Append(result, errors.Wrapf(err, ...))
}
```

The rest of the code can stay as returns though as the previous logic was fine.

runtime/drive_handler.go
```diff
@@ -107,6 +107,10 @@ func (h *StubDriveHandler) Reserve(
 	if len(h.freeDrives) == 0 {
 		return nil, ErrDrivesExhausted
 	}
+	if _, ok := h.usedDrives[id]; ok {
+		// This case means that driver wasn't released or removed properly
+		return nil, fmt.Errorf("drive with ID %s already in use, try to release it")
```
sipsma: I think this is missing the `id` to fill in `%s` with. Also, while `try to release it` is good in that it's trying to guide the user to fixing the problem, I'm not sure it's going to mean much to someone unfamiliar with the internals of our code. Maybe just including something similar to the comment above this line like `drive with ID %s already in use, a previous attempt to remove it may have failed` would be better in that it could guide users to their logs to see a previous failure or something like that.

runtime/drive_handler.go
```diff
@@ -96,32 +96,79 @@ type StubDriveHandler struct {
...
+	}
+
+	_, err := driveMounter.UnmountDrive(requestCtx, &drivemount.UnmountDriveRequest{
+		DriveID: drive.driveID,
+	})
+	if err != nil {
+		return errors.Wrap(err, "failed to unmount drive")
+	}
+
+	err = machine.UpdateGuestDrive(requestCtx, drive.driveID, drive.driveMount.HostPath)
```
sipsma: I think we want to patch this back to the original stubfile, so instead of `drive.driveMount.HostPath` (which is the path to real mount after `withMountConfig` is called) we could do `drive.stubPath`.

Not 100% sure, but I think this might fix the integ test failure occurring in buildkite, which looking at the logs seems to be caused by a failure to `Release` a drive due to the patch failing during container delete.

runtime/drive_handler.go
```diff
@@ -96,32 +96,79 @@ type StubDriveHandler struct {
...
 	options, err := setReadWriteOptions(options, freeDrive.driveMount.IsWritable)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	h.freeDrives = h.freeDrives[1:]
-	h.usedDrives[id] = freeDrive
-	return freeDrive.withMountConfig(
+	stubDrive := freeDrive.withMountConfig(
```
sipsma: I think we should probably not use a separate `stubDrive` variable here and instead do `freeDrive = freeDrive.withMountConfig(...)`. This will ensure we get all the relevent state stored in the map and should also fix some of the unit test failures you're getting right now (I suspect anyways).

runtime/drive_handler.go
```diff
@@ -96,32 +96,80 @@ type StubDriveHandler struct {
...
+	})
+	if err != nil {
+		return errors.Wrap(err, "failed to unmount drive")
+	}
+
+	err = machine.UpdateGuestDrive(requestCtx, drive.driveID, drive.stubPath)
+	if err != nil {
+		return errors.Wrap(err, "failed to patch drive")
+	}
+
```
sipsma: I just took a look at the errors in buildkite in depth, they are more complicated than I was originally thinking, but there is an easy fix that will work in the short-term: change `drive.stubPath` to `filepath.Base(drive.stubPath)`.

The problem here is that there's some pre-existing implicit dependencies between the jailer and this drive handler. The error that was occurring in buildkite was due to the fact that here we specify a full path on the host (i.e. `/var/lib/fccd/<vmID>/rootfs/stub0`), but a jailed firecracker binary lives inside a rootfs where paths for the same file are different (i.e. `/stub0`).

The reason just using `filepath.Base(drive.stubPath)` fixes the problem is because that will set it to the correct path for both the no-op jailer and the [runc jailer](https://github.com/firecracker-microvm/firecracker-containerd/blob/e80d4cc83f1207c0b9cd5eb1e5bacc67e55e6415/runtime/runc_jailer.go#L195-L222). This fix is extremely delicate, but I am okay with it for now as there are lots of other changes being made to the jailer at the moment and I don't want to create unnecessary conflicts. If you update with `filepath.Base(drive.stubPath)` and it works, I'll just go ahead and create a followup issue for finding a better solution.

Zyqsempai: Great, thanks for explanation.

Zyqsempai:
@sipsma PTAL

PR #342, opened 2019-11-25
Unify test images; cleanup integ-test image build.

### LINKED ISSUE #313: Simplify test-image Dockerfile

sipsma:
There are a few related changes here. The overall goal is to have a single image
used for running tests, to simplify the build of that image and to make it
faster to rebuild test-images locally in order to shorten dev-build-test loops.

This is primarily accomplished by:
1. Adding an "all-in-docker" Makefile target that builds all the binaries
   produced by firecracker-containerd inside a docker container and places
   those binaries on the host filesystem (allowing for the host to serve as the
   cache and normal Makefile logic to determine when a rebuild is necessary).
2. Using the above to replace the builds that previously happened in some of
   the integ-test Dockerfiles, allowing there to be just one single
   Dockerfile.integ-test.
3. Replacing a lot of the inlined logic in Dockerfile.integ-test that installs
   binaries in various places via new Makefile targets. This is beneficial in
   that the targets are useful both inside and outside the test-image build
   and allow for more caching/skipping of unnecessary steps, such as for
   downloading the default vmlinux binary.
4. Use of a docker volume for the GOPATH inside running docker images, allowing
   a cache to be shared between all of them and simplifying the process of
   ensuring all required packages are pulled before running tests w/ no network.

Closes #313 


runtime/Makefile
```diff
@@ -18,10 +18,18 @@ SOURCES:=$(shell find . -name '*.go')
 GOMOD := $(shell go env GOMOD)
 GOSUM := $(GOMOD:.mod=.sum)
 DOCKER_IMAGE_TAG?=latest
+GO_CACHE_VOLUME_NAME?=gocache
+FIRECRACKER_CONTAINERD_TEST_IMAGE?=localhost/firecracker-containerd-test
 
 INTEG_TEST_SUFFIX := _Isolated
-INTEG_TESTNAMES=$(shell go test -list . | sed '$$d' | grep $(INTEG_TEST_SUFFIX))
-
+INTEG_TESTNAMES=$(shell docker run --rm \
```
kzys: "make test" needs Go and would download the dependencies of firecracker-containerd. Do we need "docker run" here? Do we want to reduce what developers need to have on the host machine where Docker containers are running?

sipsma: This allows us to reuse the `gocache` volume and, like you were getting at, reduces the places where we require go actually be installed on the host system (I would need to check if it fully eliminates all of them, it might). Unless I'm missing something, `make test` shouldn't be a strict requirement for `go` because there is also the `test-in-docker` step right below it that calls `make test` in a container, which can be used instead.

At worst, this is harmless because the integ-tests need to run in docker containers anyways, so given the small benefits doing it this way is my preference. Let me know what you think.

kzys: I see. Yeah. I completely forgot about the existence of "make test-in-docker".

Our BuildKite host is currently having Go, but we might remove that to verify all (or most of core) targets work without having Go on the host.

Issue #264, opened 2019-09-21
Shorten the firecracker unix socket path


haikuoliu:
Firecracker containerd constructs the firecracker socket path using pattern `/var/run/firecracker-containerd/$NAMESPACE/$VMID/firecracker.sock`, however, Linux has a limit on the length of socket path, that means if I am using uuid for both namespace and vmid, the firecracker socket path will be 122 characters long which is beyond the limit. 

Containerd used to have this problem to create shim socket, they fixed the issue in this [PR](https://github.com/containerd/containerd/pull/3046) where they do sha256 on the path, is it possible for firecracker containerd to do the same?

Thanks,
Haikuo

sipsma:
We definitely need to address this, but unfortunately I don't believe we'll be able to do what containerd did with the shim socket path. The difference is containerd shims use abstract unix sockets, which don't actually exist on the filesystem (which is confusing because it's still specified with a "path", but the path is really just an id unique within the network namespace the abstract socket lives in).

Firecracker uses normal unix sockets, not abstract, so we can't solve the problem by just taking a hash. What I believe we can do is specify the socket path as a relative one. The shim process and Firecracker VMM process will be in the `$VMID` directory, so I think they can get away with just specifying `./firecracker.sock`. They should never change working dir once started up, so this should be safe and well under the character limit.

PR #346, opened 2019-11-27
Adds configurable jail path

### LINKED ISSUE #345: jailed VM creation fails on Amazon Linux 2
firecracker-containerd currently uses jail locations under `/run`. On AL2 (and presumably CentOS and other RHEL variants) this filesystem is mounted with the `nodev` flag, so the device nodes created by runc are not usable.

firecracker-containerd debug logs show:

```
INFO[2019-11-26T22:15:06.264606168Z] Successfully ran jailer handler               jailer=runc ociBundlePath=/run/firecracker-containerd/firecracker-containerd-example/fc-example runcBinaryPath=/usr/local/bin/runc runtime=aws.firecracker vmID=fc-example
INFO[2019-11-26T22:15:06.264668138Z] Called startVMM(), setting up a VMM on /run/firecracker-containerd/firecracker-containerd-example/fc-example/rootfs/api.socket  runtime=aws.firecracker vmID=fc-example
WARN[2019-11-26T22:15:06.381824962Z] firecracker exited: exit status 1             runtime=aws.firecracker vmID=fc-example
WARN[2019-11-26T22:15:06.381911059Z] Failed handler "fcinit.StartVMM": Firecracker did not create API socket /run/firecracker-containerd/firecracker-containerd-example/fc-example/rootfs/api.socket: 1 error occurred:
        * exit status 1
  runtime=aws.firecracker vmID=fc-example
ERRO[2019-11-26T22:15:06.381939624Z]                                               error="failed to create VM: failed to start the VM: Firecracker did not create API socket /run/firecracker-containerd/firecracker-containerd-example/fc-example/rootfs/api.socket: 1 error occurred:\n\t* exit status 1\n\n" runtime=aws.firecracker vmID=fc-example
ERRO[2019-11-26T22:15:06.382144578Z]                                               error="shim CreateVM returned error: rpc error: code = Unknown desc = failed to create VM: failed to start the VM: Firecracker did not create API socket /run/firecracker-containerd/firecracker-containerd-example/fc-example/rootfs/api.socket: 1 error occurred:\n\t* exit status 1\n\n"
DEBU[2019-11-26T22:15:06.383322943Z] shim has been terminated                      error="signal: killed" vmID=fc-example
```

...which is not the most helpful error.

The location used for VM jailing should be configurable.  Appropriate locations will vary from site to site.

xibz:
Due to the issue of Amazon Linux 2 having restriction on /run, this was causing
no devices to be able to be ran. To fix this we added a configurable jail path
which should allow for differences in environments that have restrictions on
the default jail path. Also changes the default from being the location of the
shim dir to /srv/jailer.

This also bind mounts instead of hard linking the fifo files as there were issues cross linking since
/run, which is where the shim dir lives, being a tmpfs

Fixes #345 

*Issue #, if available:*

*Description of changes:*


internal/vm/dir.go
```diff
@@ -183,7 +185,7 @@ func relPathTo(absPath string) (string, error) {
 
 	relPath, err := filepath.Rel(cwd, absPath)
 	if err != nil {
-		return "", errors.Wrapf(err, "failed to get relative path from %q to %q", cwd, absPath)
+		return absPath, nil
```
xibz: If anyone feels ill about this change, another option is to return either a typed error or a known error. That way we can check against the error type and just use the absolute path instead. However, we'd do that literally for all relative methods, which is why I decided to just return the absolute path here

kzys: We wanted to return relative paths to workaround #264. Wouldn't we hit the issue again by returning absolute paths?

runtime/jailer.go
```diff
@@ -81,6 +83,10 @@ func newJailer(
 		return newNoopJailer(ctx, l, service.shimDir), nil
 	}
 
+	if err := os.MkdirAll(ociBundlePath, 0644); err != nil {
+		return nil, errors.Wrap(err, "failed to create oci bundle path")
```
kzys: How about including the path itself in the error, as like errors in runc_jailer.go?

runtime/runc_jailer.go
```diff
@@ -246,15 +246,30 @@ func (j *runcJailer) BuildLinkFifoHandler() firecracker.Handler {
...
 			newFifoPath := filepath.Join(contentsPath, fifoFileName)
-			if err := os.Link(m.Cfg.LogFifo, newFifoPath); err != nil {
-				return err
+
+			fifoFd, err := os.Create(newFifoPath)
+			if err != nil {
+				return errors.Wrapf(err, "failed to create %s", newFifoPath)
+			}
+			if err := fifoFd.Close(); err != nil {
+				j.logger.WithError(err).Debug("failed to close fifo file")
```
kzys: How about returning the error instead of just logging? I'm unsure what would cause the error.

runtime/runc_jailer.go
```diff
@@ -246,15 +246,30 @@ func (j *runcJailer) BuildLinkFifoHandler() firecracker.Handler {
...
 			}
 			m.Cfg.LogFifo = newFifoPath
 
 			metricFifoFileName := filepath.Base(m.Cfg.MetricsFifo)
 			newMetricFifoPath := filepath.Join(contentsPath, metricFifoFileName)
-			if err := os.Link(m.Cfg.MetricsFifo, newMetricFifoPath); err != nil {
-				return err
+			metricFifoFd, err := os.Create(newMetricFifoPath)
+			if err != nil {
+				return errors.Wrapf(err, "failed to create %s", newMetricFifoPath)
```
kzys: Same as above

sipsma:
What if we just made the entire shimdir's location configurable? I.e. make the string here configurable via the runtime config file instead of a constant: https://github.com/firecracker-microvm/firecracker-containerd/blob/9c17257c296667660c0d4e19ffe379a6b74e490c/internal/vm/dir.go#L34

I suspect that would simplify a lot (we don't need to manage separate shim and jail dirs, don't need to figure out what to do the relpath issue, etc.) and is probably a useful feature in general anyways.

runtime/config.go
```diff
@@ -58,6 +58,9 @@ type Config struct {
 // TODO: Add netns field
 type JailerConfig struct {
 	RuncBinaryPath string `json:"runc_binary_path"`
+	// JailPath is used to determine where the jailed application should be
+	// contained in.
```
nmeyerhans: This comment should be a bit more precise. "Jailing" is not necessarily a well defined term, so we should be really clear about what we mean when we talk about it. Something like "JailPath specifies the location of a directory contain per-VM minimal root directories for use in private mount namespaces".

internal/vm/dir.go
```diff
@@ -22,16 +22,29 @@ import (
...
-	"github.com/firecracker-microvm/firecracker-containerd/internal/bundle"
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
+
+	"github.com/firecracker-microvm/firecracker-containerd/config"
+	"github.com/firecracker-microvm/firecracker-containerd/internal"
+	"github.com/firecracker-microvm/firecracker-containerd/internal/bundle"
 )
 
+var jailPath string
```
xibz: Going to change this to just have an extra parameter when calling `ShimDir` rather than having a singleton.

runtime/helpers.go
```diff
@@ -22,10 +22,15 @@ import (
 	"github.com/firecracker-microvm/firecracker-go-sdk"
 	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
 
+	"github.com/firecracker-microvm/firecracker-containerd/config"
 	"github.com/firecracker-microvm/firecracker-containerd/proto"
 )
 
-func machineConfigurationFromProto(cfg *Config, req *proto.FirecrackerMachineConfiguration) models.MachineConfiguration {
+const (
+	defaultMemSizeMb = 128
```
IRCody: This is minor but why move this?

xibz: Basically due to the `Config` needing to be relocated to the `config` package, this `defaultMemSizeMb` was also in the `config.go` file. And while keeping it there makes sense at first, I realized it wasn't used in the `config.go` file at all and only here. So basically, either we expose `defaultMemSizeMb` or we relocate it. And I thought relocating was better than exposing.

runtime/runc_jailer.go
```diff
@@ -246,15 +246,30 @@ func (j *runcJailer) BuildLinkFifoHandler() firecracker.Handler {
...
-				return err
+
+			fifoFd, err := os.Create(newFifoPath)
+			if err != nil {
+				return errors.Wrapf(err, "failed to create %s", newFifoPath)
+			}
+			if err := fifoFd.Close(); err != nil {
+				j.logger.WithError(err).Debug("failed to create fifo file")
+			}
+			if err := syscall.Mount(m.Cfg.LogFifo, newFifoPath, "none", syscall.MS_BIND, ""); err != nil {
```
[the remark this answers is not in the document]
xibz: @nmeyerhans - That's an interesting idea, but yea, I can definitely explore that after this PR. I'll create a github issue outlining the problem. What is the issue with having cross-namespace mounts? Is it just generally bad practice?

runtime/helpers.go
IRCody: It's unrelated to this PR, but it seems weird that defaultMemSizeMb isn't used in tehs ame place as the other defaults?

Is the use of `defualtMemSizeMb` in `machineCOnfigurationFromProto` only because it's not used in `LoadConfig` with the other defaults? Would it make sense to move it's use there rather than moving the definition?

xibz: The `Config` does not have a `MemSizeMib` field. So we'd be introducing another field in the configuration, which I do not like either if it isn't needed. Currently you can only overwrite the default value by `req.MemSizeMib`

firecracker-control/local.go
```diff
@@ -117,8 +125,12 @@ func (s *local) CreateVM(requestCtx context.Context, req *proto.CreateVMRequest)
 
 	// If we're here, there is no pre-existing shim for this VMID, so we spawn a new one
 	defer shimSocket.Close()
+	if err := os.MkdirAll(s.config.ShimBaseDir, 0644); err != nil {
```
sipsma: I think this should 0700, dirs need the `x` bit set in order to grant search-permissions and I don't see any reason to grant read access to group or other.

runtime/jailer.go
```diff
@@ -81,6 +83,10 @@ func newJailer(
 		return newNoopJailer(ctx, l, service.shimDir), nil
 	}
 
+	if err := os.MkdirAll(ociBundlePath, 0644); err != nil {
```
sipsma: I also think this should be 0700 for same reasons as the other comment

runtime/runc_jailer.go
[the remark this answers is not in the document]
sipsma: Is this change to use bind mounts still necessary after updating to just make the whole shim dir configurable? Are we still crossing fs-boundaries?

tools/docker/Dockerfile.integ-test
```diff
@@ -55,6 +55,7 @@ COPY firecracker-control/cmd/containerd/firecracker-ctr /usr/local/bin/ctr
 
 COPY runtime/firecracker-runc-config.json.example /etc/containerd/firecracker-runc-config.json
 COPY tools/image-builder/rootfs.img /var/lib/firecracker-containerd/runtime/default-rootfs.img
+COPY tools/docker/do_not_edit.config.json /etc/containerd/firecracker-runtime.json
```
sipsma: We shouldn't need to put a firecracker-runtime config file here, we generate them during each test case now: https://github.com/firecracker-microvm/firecracker-containerd/blob/e80d4cc83f1207c0b9cd5eb1e5bacc67e55e6415/runtime/integ_test.go#L56

xibz: This is needed due to the `firecracker-local` package which is loaded up before the tests can write the config. I've renamed this for a more appropriate name and created a github issue, #353, so we can eventually move away from this config file.

runtime/jailer.go
xibz: Fixed!

firecracker-control/local.go
xibz: Fixed!

runtime/runc_jailer.go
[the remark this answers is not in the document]
xibz: Nope, good catch. Fixed!

config/config.go
```diff
@@ -51,7 +51,11 @@ type Config struct {
 	// the VM will default to using the network interfaces as specified here. This is especially
 	// useful when a CNI-based network interface is provided in DefaultNetworkInterfaces.
 	DefaultNetworkInterfaces []proto.FirecrackerNetworkInterface `json:"default_network_interfaces"`
-	JailerConfig             JailerConfig                        `json:"jailer"`
+	// ShimBaseDir is the location to which the shim will be ran from. In
```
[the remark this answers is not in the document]
xibz: Fixed!

runtime/helpers.go
IRCody: Do you happen to know off-hand why mem size was treated differently than other config values? It seems like it *should* be part of the config and would be simpler if it was. Maybe there is a reason I'm not seeing/aware of?

xibz: I think it's due to complications of priority. Basically, getting the memory size would need to check if `req.MemSizeMib` is zero, if it is use the `cfg.MemSizeMib`. We probably want to take all of the config values out that can be used in `CreateVM`. Meaning all the machine config stuff would now be apart of `CreateVM`. Then the config would only have values that wouldn't change, like `ShimBaseDir` or the runc bin path. Stuff like that

IRCody: Agree on out of scope of this PR, was just curious about it! Resolving.

nmeyerhans:
@xibz Could you squash all your fixup commits and let me take one last look before we merge?

config/config.go
```diff
@@ -33,7 +33,7 @@ const (
 	defaultRootfsPath  = defaultFilesPath + "default-rootfs.img"
 	defaultCPUCount    = 1
 	defaultCPUTemplate = models.CPUTemplateT2
-	defaultMemSizeMb   = 128
+	defaultShimBaseDir = "/var/run/firecracker-containerd/"
```
xibz: Debating if we should choose a different path that works on both RHEL and debian systems. Thoughts @nmeyerhans?

xibz:
@nmeyerhans - squashed the commits!

config/config.go
nmeyerhans: Given that we're copying binaries and filesystem images to subdirectories under this path, I think we should probably change the default to something that is more likely to be backed by actual disk rather than RAM. So `/var/lib/firecracker-container/shim-base` or something like that, by default, might be OK. It reduces the likelihood that we're going to exhaust system RAM, and mount options aren't likely to vary across distros. (Mounting a separate filesystem on `/var` isn't super common these days, though it once was, whereas `/run` is a separate mount point by default on most modern distros.)

If you do make this change, double check the docs/ subdirectory to make sure we don't have references that need to be updated.

xibz:
@nmeyerhans - Ah, must've happened when I was fixing the merge conflict during autosquash. Will go ahead and clean those up

docs/quickstart.md
```diff
@@ -96,7 +96,7 @@ state = "/run/firecracker-containerd"
 [proxy_plugins]
   [proxy_plugins.firecracker-naive]
     type = "snapshot"
-    address = "/var/run/firecracker-containerd/naive-snapshotter.sock"
+    address = "/var/lib/firecracker-containerd/shim-base/naive-snapshotter.sock"
```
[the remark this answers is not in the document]
xibz: It does not, but I figured since we already had it in the shim base dir, I'd continue with that example. Are you opposed to having it in there?

firecracker-control/local.go
```diff
@@ -117,8 +125,12 @@ func (s *local) CreateVM(requestCtx context.Context, req *proto.CreateVMRequest)
 
 	// If we're here, there is no pre-existing shim for this VMID, so we spawn a new one
 	defer shimSocket.Close()
+	if err := os.MkdirAll(s.config.ShimBaseDir, 0700); err != nil {
```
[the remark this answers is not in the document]
xibz: Fixed. However, I don't know how I feel about having a default location that doesn't exist...

docs/quickstart.md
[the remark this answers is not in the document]
nmeyerhans: Yeah, let's leave it in /var/lib/firecracker-containerd.

firecracker-control/local.go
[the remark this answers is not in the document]
nmeyerhans: It's pretty rare for applications to create the full directory hierarchy their own state directories. In part, it silently masks typos, and potentially has the user running in a very weird configuration without realizing it.

`os.MkDir()` would be more appropriate here.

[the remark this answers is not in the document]
xibz: Fixed!

docs/getting-started.md
```diff
@@ -259,6 +261,13 @@ configuration file has the following fields:
 ```
 </details>
 
+The default `shim_base_dir` is /var/lib/firecracker-containerd/shim-base and is
+required to be created before running firecracker-containerd.
+
+```bash
+mkdir -p /var/lib/firecracker-containerd/shim-base
```
nmeyerhans: We only require the user to create `/var/lib/firecracker-containerd`, right?

docs/quickstart.md
```diff
@@ -86,7 +86,9 @@ curl -fsSL -o hello-vmlinux.bin https://s3.amazonaws.com/spec.ccfc.min/img/hello
 # separate storage from the default containerd binary
 sudo mkdir -p /etc/firecracker-containerd
 sudo mkdir -p /var/lib/firecracker-containerd/containerd
-sudo mkdir -p /run/firecracker-containerd
+# Create the shim base directory for which firecracker-containerd will run the
+# shim from
+sudo mkdir -p /var/lib/firecracker-containerd/shim-base
```
nmeyerhans: Same as above.

docs/getting-started.md
xibz: Fixed!

nmeyerhans:
Can you resolve the outstanding conflicts in `service_integ_test.go`? Looks good, otherwise.

Issue #294, opened 2019-10-17
Propagate firecracker-containerd specific errors to clients


sipsma:
Currently, clients using the fc-control client are able to receive [grpc error codes](https://godoc.org/google.golang.org/grpc/codes) in error objects returned by API calls, but those are very generic error types that don't communicate more specific information back to the caller. More specific information is at minimum extremely useful for integration testing (allows you to assert the correct failure occurred), but is also potentially valuable for real users who want to know what specifically went wrong for specific error-handling.

I spent a day trying various ways of implementing this as part of the drive mount support (where we want to assert on specific errors occuring in integ tests), but all the ways I tried either didn't work or became too large an effort:
1. I tried updating APIs to return an actual response object (instead of `Empty`) that contained a field with an error type enum. The enum field would be set to a specific error code whenever a service returned a non-nil error. This was a nice interface for both client and server, but didn't work in practice because ttrpc (and possible grpc too, have not checked) will ignore any response object returned by a service if the service also returns a non-nil error (setting the response to nil). This makes the interface much uglier and difficult to manage correctly for both services and clients.
2. [grpc status objects contain a `Details` field](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails), which seems to be potentially useful for returning typed details of what error occurred. However, the interface is hard to use as it requires the details be protobuf messages (not just strings) and adding details to a status object itself can return an error (so we can get an error trying to generate an error object)... This may be viable but might be a large effort.
3. We can include specific error codes in the [grpc status object Message](https://godoc.org/google.golang.org/grpc/status#Status.Message) field, but this leaves clients to check error types via string matching (which is very fragile). We could also consider structuring of messages (with associate serialization and deserialization), though that's a large effort and feels a bit odd because we'd be doing ser/der within our already ser/der'd protobuf messages.

Right now, I think figuring out how to use the Details field of the grpc status object may be the most promising, but this needs more investigation.

IRCody:
[Here](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto) are some generic error protos that are meant to go into that details field.

sipsma:
> [Here](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto) are some generic error protos that are meant to go into that details field.

[...]

PR #348, opened 2019-11-27
Returns codes.DeadlineExceeded when we forcebily kill a VM

kzys:
*Issue #, if available:*

 #247 

*Description of changes:*

There are three commits in this PR
* Add a rootfs to simulate the situation: Not so sure this is the best way though.
* Remove errors.Wrap from firecracker-control: Shim is more aware about errors and fc-control should be act as a transparent proxy. We should remove all of errrors.Wrap from this layer and let shim decide errors.
* Actually fix the issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,6 +183,10 @@ DEFAULT_RUNC_JAILER_CONFIG_INSTALLPATH?=/etc/containerd/firecracker-runc-config.
 $(DEFAULT_RUNC_JAILER_CONFIG_INSTALLPATH): $(ETC_CONTAINERD) runtime/firecracker-runc-config.json.example
 	install -D -o root -g root -m400 runtime/firecracker-runc-config.json.example $@
 
+ROOTFS_SLOW_REBOOT_INSTALLPATH=$(FIRECRACKER_CONTAINERD_RUNTIME_DIR)/rootfs-slow-reboot.img
+$(ROOTFS_SLOW_REBOOT_INSTALLPATH): tools/image-builder/rootfs-slow-reboot.img $(FIRECRACKER_CONTAINERD_RUNTIME_DIR)
+	install -D -o root -g root -m400 $< $@
+
 .PHONY: default-vmlinux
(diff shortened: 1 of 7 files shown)


kzys:
Not so sure codes.DeadlineExceeded is the best error code. We might introduce more FCCD-specific error codes (tracked by #294).

Makefile
```diff
@@ -189,8 +193,8 @@ default-vmlinux: $(DEFAULT_VMLINUX_NAME)
 .PHONY: install-default-vmlinux
 install-default-vmlinux: $(DEFAULT_VMLINUX_INSTALLPATH)
 
-.PHONY: install-default-rootfs
-install-default-rootfs: $(DEFAULT_ROOTFS_INSTALLPATH)
+.PHONY: install-rootfs
```
samuelkarp: I'd prefer to leave `install-default-rootfs` as a separate target.  A user installing firecracker-containerd to run containers (and not develop firecracker-containerd) won't need the slow-reboot version.

kzys: That makes sense. We use this one for "make install"...

tools/docker/Dockerfile.integ-test
```diff
@@ -53,6 +53,7 @@ RUN --mount=type=bind,target=/src make -C /src \
   install-test-cni-bins \
   demo-network
 
+COPY tools/image-builder/rootfs-slow-reboot.img /var/lib/firecracker-containerd/runtime/rootfs-slow-reboot.img
```
sipsma: Maybe we should have something like a `install test-depends` target and include installing this along with the pre-existing `install-test-cni-bins` target. If we are going to want to support testing outside containers directly on bare-metal hosts (like we were talking about before), we'd get a lot of benefit from being able to just run the same `install test-depends` target on both the host and inside a container rather than duplicate stuff like this copy between them.

Otherwise LGTM

kzys: Sounds good. Will do.

Issue #276, opened 2019-10-02
Support optionally forwarding Agent logs from within VM to the host for debugging


sipsma:
It's often very helpful for debugging to have logs from the agent process running inside the VM. Getting those logs is not straightforward as by default they will stay inside the VM.

One quick hack to get them is [this change](https://github.com/sipsma/firecracker-containerd/commit/2be7fb3c969438e16fca23b86ff151d1ea3af08d), which forwards the agent logs over `/dev/console` to the runtime process on the host, which in turn forwards them to containerd logs.

That approach can't be merged into master though as it results in a lot of noise in the containerd logs and could possibly lead to heavy use of `/dev/console`, which has been known to result in performance issues in the past. A mergable solution to this would need to deal with those issues; some ideas to explore are only forwarding them in a debug mode and making the destination of the logs configurable (could be containerd logs, could be a file, etc.).

kzys:
Yes. It is unsettling to see some of the logs go nowhere. Regarding the console, https://github.com/firecracker-microvm/firecracker/issues/792 also mentions that the console is slow.

I think it may make sense to collect all of systemd logs. Issues like https://github.com/firecracker-microvm/firecracker-containerd/issues/253 cannot be seen by just looking the agent.

sipsma:
Agree about users having the ability to collect all systemd logs.

I think it's worth seeing if we can find a way for systemd journal logs to get streamed out of the VM over vsock, which thanks to the new Firecracker vsock implementation, could mean that logs are available on the host at a unix socket, which could be convenient for users.

I don't know enough about systemd, but if there is a way to write a journald plugin or similar that forwards logs over an AF_VSOCK socket, that could be a decent generic solution to this problem.

mmonaco:
Hey, we're streaming structured (json) journald over vsock. It works well enough but can be flaky. I came here looking for best practices / prior art on what others are doing.

PR #352, opened 2019-11-30
Add send2vsock to send systemd journal from a guest to the host

kzys:

*Issue #, if available:*

#276

*Description of changes:*

This PR is still work-in-progress. Early feedback is welcome!

Basically I want to run `journalctl -f | send2vsock` as a systemd unit. systemd can forward logs to syslogd, but supporting a remote syslogd location is not supported and using a mini daemon is recommended instead (see https://github.com/systemd/systemd/issues/7170).

### DIFF
--- a/send2vsock/Makefile
+++ b/send2vsock/Makefile
@@ -1,0 +1,35 @@
+# Copyright 2018-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may
+# not use this file except in compliance with the License. A copy of the
+# License is located at
+#
+# 	http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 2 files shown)


sipsma:
Think the basic idea here is good, just some more things we'll need to figure out:
1. Users are going to need a way to consume these logs
1. We'll need to make sure that the vsock port used by this doesn't conflict with the vsock ports used internally for apis/stdio.

One initial thought I had was to run this code as a semi-privileged "sidecar" container (we could vend the container definition in some utility library in this repo). That solves both of the above problems pretty cleanly but has a pretty big disadvantage in that we won't be able to get the logs until the container has actually started, which misses a lot of problems that can happen during boot+agent start.

My only other idea is to add parameters to the CreateVM api that define "reserved" vsock streams, which will allow us to reserve port numbers and tell users where to find the unix socket on which they can listen to services from within their VM, such as this journal forwarder. Curious to hear others' thoughts.

kzys:
I'm going to close the PR for now. We will check the overall performance first instead of just avoid using the console device.

Issue #343, opened 2019-11-26
Increase number of VMs tested in `TestMultipleVMs_Isolated`


sipsma:
`TestMultipleVMs_Isolated` has in the past been very helpful in finding rare race-conditions due to the fact that it (unsurprisingly) spins up multiple VMs concurrently. However, it was previously limited in how many VMs it could spin up seemingly due to the use of the naive snapshotter, which appeared to become a large bottleneck when many container snapshots were being made in parallel. It currently only spins up 5 VMs each with either 5 or 3 containers.

Now that we use devmapper in our CI system, it's worth seeing if we can increase the number of VMs+containers we spin up in the test as it could give us higher chances of catching rare race conditions. If devmapper does in fact improve the situation, updating this test could serve as a primitive form of a load test, though it would still focus on making functional assertions at this time.

This may require a little extra work as while it would be great for our CI hosts to spin up something like 100 VMs in parallel if feasible, we don't want that to be the default as most developers are not going to have a host with enough RAM to start up 100 VMs.

Zyqsempai:
@sipsma I would love to work on this one, if you don't mind!

nmeyerhans:
resolved via #356 

Thanks @Zyqsempai

PR #356, opened 2019-12-05
Added possibility to change number of vm's for MultipleVmsTest

Zyqsempai:

*Issue #343 *

*Description of changes:*
Added possibility to increase number of VM's through buildkite for TestMultipleVMs_Isolated.

### DIFF
--- a/runtime/service_integ_test.go
+++ b/runtime/service_integ_test.go
@@ -66,6 +66,9 @@ const (
 	defaultVMRootfsPath = "/var/lib/firecracker-containerd/runtime/default-rootfs.img"
 	defaultVMNetDevName = "eth0"
 	varRunDir           = "/run/firecracker-containerd"
+
+	numberOfVmsEnvName = "NUMBER_OF_VMS"
+	defaultNumberOfVms = 5
 )
 
 // Images are presumed by the isolated tests to have already been pulled
@@ -220,24 +223,20 @@ func TestMultipleVMs_Isolated(t *testing.T) {
 	netns, err := ns.GetCurrentNS()
 	require.NoError(t, err, "failed to get a namespace")
 
+	// numberOfVmsEnvName = NUMBER_OF_VMS ENV and is configurable from buildkite
+	numberOfVms, err := strconv.Atoi(os.Getenv(numberOfVmsEnvName))
+	require.NoError(t, err, "failed to get NUMBER_OF_VMS env")
+	if numberOfVms == 0 {
+		numberOfVms = defaultNumberOfVms
+	}
+	t.Logf("TestMultipleVMs_Isolated: will run %d vm's", numberOfVms)
+
 	cases := []struct {
 		MaxContainers int32
 		JailerConfig  *proto.JailerConfig
 	}{
 		{
 			MaxContainers: 5,
-		},
-		{
-			MaxContainers: 5,
-		},
-		{
-			MaxContainers: 5,
-		},
-		{
-			MaxContainers: 3,
-			JailerConfig: &proto.JailerConfig{
-				NetNS: netns.Path(),
-			},
 		},
 		{
 			MaxContainers: 3,
@@ -265,7 +264,9 @@ func TestMultipleVMs_Isolated(t *testing.T) {
 	// container ends up in the right VM by assigning each VM a network device with a unique mac address and having each container
 	// print the mac address it sees inside its VM.
 	var vmWg sync.WaitGroup
-	for vmID, c := range cases {
+	for i := 0; i < numberOfVms; i++ {
+		caseTypeNumber := i % len(cases)
+		c := cases[caseTypeNumber]
 		vmWg.Add(1)
 		go func(vmID int, containerCount int32, jailerConfig *proto.JailerConfig) {
 			defer vmWg.Done()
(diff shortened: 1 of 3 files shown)


Zyqsempai:
@sipsma PTAL

runtime/service_integ_test.go
```diff
@@ -220,25 +223,20 @@ func TestMultipleVMs_Isolated(t *testing.T) {
 	netns, err := ns.GetCurrentNS()
 	require.NoError(t, err, "failed to get a namespace")
 
+	// numberOfVmsEnvName = NUMBER_OF_VMS ENV and is configurable from buildkite
+	numberOfVms, err := strconv.Atoi(os.Getenv(numberOfVmsEnvName))
+	if err != nil || numberOfVms == 0 {
```
xibz: I would much prefer a `require.NoError` check on `err`

runtime/service_integ_test.go
```diff
@@ -265,7 +263,9 @@ func TestMultipleVMs_Isolated(t *testing.T) {
 	// container ends up in the right VM by assigning each VM a network device with a unique mac address and having each container
 	// print the mac address it sees inside its VM.
 	var vmWg sync.WaitGroup
-	for vmID, c := range cases {
+	for i := 0; i < numberOfVms; i++ {
+		caseTypeNumber := 1 % 2
```
xibz: this should be `caseTypeNumber := 1 % len(cases)`

runtime/service_integ_test.go
```diff
@@ -265,7 +263,9 @@ func TestMultipleVMs_Isolated(t *testing.T) {
...
+		c := cases[caseTypeNumber]
 		vmWg.Add(1)
 		go func(vmID int, containerCount int32, jailerConfig *proto.JailerConfig) {
```
xibz: I am also curious to how many are we actually running in parallel due to the `StopVM` call. We may need to call `StopVM` at a later time and have the VMs run some process to ensure we are getting accurate number of concurrency. Thoughts on this @sipsma

nmeyerhans: I don't think this is meant to be a true scale test, so we don't really need to max out the host capacity. AFAICT, 100 was chosen pretty much arbitrarily, and it is an upper bound on the number of VMs that will run concurrently, not necessarily a goal. Even if we're only running ~20 VMs at any one time, that's still more than we're running today. If we assume that more VMs means more likelihood of catching concurrency-related bugs, then a 3-4x increase in the number of running VMs sounds good to me.

At some point in the future we will need to be performing actual scale tests, but this PR doesn't need to be that.

sipsma: > I'm worried that a lot of the VMs are exiting before we can test the true bottleneck of what is parallel

This very well may be the case, but I agree with Noah that it's fine if so. We are primarily looking to find problems that occur as we actually perform actions on/in the VM (spin up/down containers, copy io, etc.). So even if we modified the test case to wait to call StopVM at the end, we would likely just end up with a bunch of VMs sitting around doing nothing for a while, which doesn't really improve coverage IMO. I agree though that if we want scale tests in the future we'd need to be more careful about issues like this.

runtime/service_integ_test.go
sipsma: If the intention is to switch back and forth between test cases (which I think is good), then I would have thought it should be `i % len(cases)`

Zyqsempai: Actually it doesn't matter in this particular case, coz we have only two cases, and it will be or 0 or 1, anyway, but I agree that in case we will have more than 2 cases, it make sense to use `i % len(cases)`

Zyqsempai:
@sipsma @xibz @nmeyerhans I made a number of vm's equal to 25, I think this is more than enough for this test.

runtime/service_integ_test.go
xibz: Sorry, yes, I meant i % len(cases). This is easier to maintain and doesnt require us to change anything if we add more cases in the future.

sipsma:
> @sipsma @xibz @nmeyerhans I made a number of vm's equal to 25, I think this is more than enough for this test.

@Zyqsempai there may have been a misunderstanding as to the suggestion around the number of VMs; I think it's good to leave `NUMBER_OF_VMS` as 100. My understanding of the discussion was that even when we set it to 100, there may in practice only be 25 running at a single time because VMs don't necessarily start/stop in perfect synchrony together.

I think leaving it at 100 is good as the test still only took about a minute to run in the CI system, which is reasonable while still increasing our potential for catching rare conditions.

runtime/service_integ_test.go
xibz: Ah, I was under the impression we wanted to actually test load. In that case, this looks fine

Zyqsempai:
@sipsma ok, got it, I tried to find compromise;) turn it back.

runtime/service_integ_test.go
```diff
@@ -265,7 +264,9 @@ func TestMultipleVMs_Isolated(t *testing.T) {
 	// container ends up in the right VM by assigning each VM a network device with a unique mac address and having each container
 	// print the mac address it sees inside its VM.
 	var vmWg sync.WaitGroup
-	for vmID, c := range cases {
+	for i := 0; i < numberOfVms; i++ {
+		caseTypeNumber := 1 % len(cases)
```
xibz: This needs to be `i % len(cases)`. I had a typo in my earlier comment!

Zyqsempai: Oh, yeah;)  fixed.

Zyqsempai:
@sipsma Done.

Issue #358, opened 2019-12-06
TestMultipleVMs failure - "exit status 148" and "race detected during execution of test"


sipsma:
There was an ephemeral failure of TestMultipleVMs here (run as part of #356 ): https://buildkite.com/firecracker-microvm/firecracker-containerd/builds/1315#1dd8d6e8-953c-417b-97f7-e029aea0a82f

I haven't looked in depth yet, but a few lines in the logs for vmID 78 (which got one of the "os: process already finished" errors), shows some odd looking lines:
```
time="2019-12-06T00:08:44.729560758Z" level=warning msg="firecracker exited: exit status 148" runtime=aws.firecracker vmID=78
...
time="2019-12-06T00:08:44.729847979Z" level=error msg="error returned from VM wait" error="1 error occurred:\n\t* exit status 148\n\n" runtime=aws.firecracker vmID=78
...
time="2019-12-06T00:08:44.729855148Z" level=error msg="the VM returns unknown error" error="1 error occurred:\n\t* exit status 148\n\n" runtime=aws.firecracker vmID=78
...
time="2019-12-06T00:08:44.730000243Z" level=debug msg="stopVMM(): sending sigterm to firecracker" runtime=aws.firecracker vmID=78
...
time="2019-12-06T00:08:44.730102402Z" level=error error="rpc error: code = Unknown desc = os: process already finished"
```

I'm not sure off the top of my head what caused the exit code to be 148. We also shouldn't need to send SIGTERM if the VM is already exited (but that may be unrelated).

sipsma:
@xibz indicated he previously encountered exit code 148 when Firecracker hit seccomp violations: https://github.com/firecracker-microvm/firecracker/issues/897

So that's something to look into as a possibility.

sipsma:
I can reproduce this locally. I saw there is an [unreleased fix for an issue that will cause a seccomp violation during a firecracker panic](https://github.com/firecracker-microvm/firecracker/blob/master/CHANGELOG.md#fixed), so I tried with 0.20 ([from here](https://github.com/serban300/firecracker/tree/release)), but still see VMs exiting with 148.

The exit 148 always happens right as the VM is supposed to shutdown (right after systemd logs say the system is about to reboot), and often (but not always) includes these logs about a bad syscall:
```
[...]

PR #362, opened 2019-12-10
Fix shutdown logic.

sipsma:
There's a few related cleanups here:
1. Fix an issue where the timeout used for Shutdown calls (as opposed to StopVM
   calls) was too long due to multiplying by time.Second twice.
2. Don't rely on the error type being exec.ExitError to determine if StopVMM
   needs to be called. This didn't work because the Go SDK wraps error types
   returned by machine.Wait(). The new logic eliminates the need for this check.
3. Expose monitorVMExit's knowledge of when the VM exits via a field, which
   allows shutdown to just read that field instead of making its own separate
   Wait call. This dedupes some logs and simplifies the shutdown function a bit.

Overall, the code related to shutdown is still pretty convoluted. The
complication is mostly due to the fact that we are given multiple contexts by
containerd shim code (shimCtx and requestCtx), which creates complicated
dependencies between those contexts and the lifecycle of the VM, our ttrpc
servers+connection and the shim process itself. This change doesn't really
help that situation, it's just focused on fixing some of the logical errors in
the shutdown procedure.

Ref: #358 

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -86,7 +86,7 @@ steps:
       - "runtime/logs/*"
     command:
       - make -C runtime integ-test FICD_SNAPSHOTTER=devmapper FICD_DM_POOL=build_${BUILDKITE_BUILD_NUMBER}_runtime
-    timeout_in_minutes: 10
+    timeout_in_minutes: 15
 
   - label: ":rotating_light: :exclamation: example tests (devmapper)"
     agents:
--- a/agent/main.go
+++ b/agent/main.go
@@ -119,7 +119,12 @@ func main() {
 	group.Go(func() error {
 		defer func() {
 			log.G(shimCtx).Info("stopping ttrpc server")
-			if err := server.Shutdown(shimCtx); err != nil {
+			// use context.Background() instead of the canceled shimCtx, which should allow shutdown to
+			// flush any pending responses before actually stopping the server. We're choosing to not
+			// use any timeouts here, deferring instead to the the higher-level runtime shim's shutdown
+			// timeouts. If we get blocked here, the runtime shim will eventually force the VM to shutdown
+			// after its timeouts.
+			if err := server.Shutdown(context.Background()); err != nil {
 				log.G(shimCtx).WithError(err).Errorf("failed to close ttrpc server")
 			}
 		}()
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -16,10 +16,10 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"math"
 	"net"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"runtime/debug"
 	"strconv"
@@ -122,7 +122,10 @@ type service struct {
 	config *config.Config
 
 	// vmReady is closed once CreateVM has been successfully called
-	vmReady                  chan struct{}
+	vmReady chan struct{}
+	// vmExitErr is set with the error returned by machine.Wait(). It should
+	// only be read after shimCtx.Done() is closed
+	vmExitErr                error
 	vmStartOnce              sync.Once
 	agentClient              taskAPI.TaskService
 	eventBridgeClient        eventbridge.Getter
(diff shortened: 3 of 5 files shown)


agent/main.go
```diff
@@ -119,7 +119,9 @@ func main() {
 	group.Go(func() error {
 		defer func() {
 			log.G(shimCtx).Info("stopping ttrpc server")
-			if err := server.Shutdown(shimCtx); err != nil {
+			// use context.Background() instead of the canceled shimCtx, which should allow shutdown to
+			// flush any pending responses before actually stopping the server
+			if err := server.Shutdown(context.Background()); err != nil {
```
mxpv: The [docs mention](https://golang.org/pkg/net/http/#Server.Shutdown) that `Shutdown` will: `... waiting indefinitely for connections to return to idle and then shut down`. Should we use context with timeout here?

sipsma: (This is a ttrpc server, not http, [but the same idea does apply](https://github.com/containerd/ttrpc/blob/master/server.go#L131))

I thought about that, but right now if we got blocked indefinitely on waiting for connections to become idle, then we'd just hit the shutdown timeout in the runtime shim and forcibly kill the VM. I think I'm okay with that because force-closing the server will often cause other errors to be returned to the runtime shim, which can result in the VM being forcibly closed anyways (i.e. `ttrpc: closed` errors to `agentClient.Shutdown`). Also, adding another timeout to a different part of the shutdown process is going to complicate it even further, which I'd really like to avoid unless we are sure we have to.

sipsma:
I saw an intermittent error in `TestMultipleVMs` that I haven't seen before: `failed to remove bundle path "/container/container-12-0": unlinkat /container/container-12-0: bad file descriptor` as part of task `Delete()` in agent.

I don't *think* this is related to this change as I didn't touch anything related to task deletion (it could just be another rare error found by increasing the number of VMs in `TestMultipleVMs`), so I'm just filing a separate issue for it for now: https://github.com/firecracker-microvm/firecracker-containerd/issues/363

runtime/service.go
```diff
@@ -1148,54 +1148,61 @@ func (s *service) Shutdown(requestCtx context.Context, req *taskAPI.ShutdownRequ
...
-		return err
-	}
-
-	shutdownCh := make(chan error)
+// shutdown will stop the VMM within the provided timeout. It attempts to shutdown gracefully by having
+// agent stop (which is presumed to cause the VM to begin a reboot) and then waiting for the VMM process
+// to exit (via the s.shimCtx.Done() channel). If that fails, StopVMM will be called to force a shutdown (currently
+// via sending SIGTERM). If that fails, the VMM will still be killed via SIGKILL when the shimCtx is canceled.
+func (s *service) shutdown(
+	requestCtx context.Context,
```
xibz: This request context isn't used anywhere but creating the timeout context. How do you feel about passing that context instead of the `requestCtx`?

sipsma: All the callers care about is that `shutdown` should have a timeout associated with it. If we have the callers create contexts, there's more duplicated code, more parameters to pass to `shutdown` (both a context and a cancel func) and more places you need to look to understand how the `shutdown` code works. I think it's simpler to just keep the fact that the timeout is currently enforced with a derived context inside `shutdown` rather than expose more to the callers.

.buildkite/pipeline.yml
```diff
@@ -86,7 +86,7 @@ steps:
       - "runtime/logs/*"
     command:
       - make -C runtime integ-test FICD_SNAPSHOTTER=devmapper FICD_DM_POOL=build_${BUILDKITE_BUILD_NUMBER}_runtime
-    timeout_in_minutes: 10
+    timeout_in_minutes: 15
```
xibz: Why did we increase the timeout?

sipsma: The commit for this change says:
```
We have enough tests now that they can take longer than the previous timeout of
10 minutes. A better long-term solution would be to run them all in parallel,
but in the meantime a 5 minute bump should suffice.
```

Most of the time it was under 10 minutes, but one time it went just slightly over, so I bumped it.

runtime/service.go
xibz: That's true with the `requestCtx` as well. You need to track what context is actually being passed in, so I don't think it is too much more to ask for. However, I don't feel too strongly on this.

.buildkite/pipeline.yml
xibz: Ah cool

PR #364, opened 2019-12-11
Add support for UID and GID when Jailing

xibz:
This adds UID and GID to the jailer config used in the CreateVMRequest
object. This has the caveat that UID and GID cannot be zero and if they
are an error will be returned

*Issue #, if available:*

*Description of changes:*


kzys:
The change looks good. Can you add some tests that use non-default UID/GID?

runtime/jailer.go
```diff
@@ -83,6 +79,14 @@ func newJailer(
 		return newNoopJailer(ctx, l, service.shimDir), nil
 	}
 
+	if request.JailerConfig.UID == 0 || request.JailerConfig.GID == 0 {
+		return nil, fmt.Errorf(
+			"attempting to run as %d:%d. 0 cannot be used for the UID or GID",
```
sipsma: I understand why we want this but it's sad in that it means the default values of the fields will result in an error. This, for instance, means that in the "demo" case where someone is trying to spin up a firecracker-containerd container with just `ctr` they will get an error if there is a jailer configuration specified in their runtime config file.

A couple ways we could potentially improve this (not necessarily mutually exclusive):
1. Change `UID` and `GID` to be strings instead of uint32 (then just validate they are convertable to uint32 when we receive them). This means we can distinguish between empty (`""`) and zero (`"0"`), which would allow us to provide a default in the case of `""` and explicitly block `"0"`.
   * This has a slightly nice side-effect of making it easier to support usernames+groupnames in the future if we wanted.
1. Update the ["demo" code path](https://github.com/firecracker-microvm/firecracker-containerd/blob/006b04a2496e6f1426e61c5c3a7b095ae45ce5aa/runtime/service.go#L316-L319) to specify a non-zero uid/gid if there is a jailer configuration to be applied

If we want to address this, I think my preference would be 1 above, but let me know what you think.

xibz: I think I prefer 2. I don't like `UID` and `GID` being strings, and I think in the future if we want usernames and groupnames, I think we should use a different field for that. I don't like having a field signify two different meanings, like UID or username.

Are you opposed to option 2?

sipsma: Ah okay, I thought the jailer config was like the other runtime configs in that it gets applied as the default configuration.

I don't think making the field be a string and potentially supporting names in addition to ids in the future would mean we are having the fields represent two different things. Allowing a user/group field to be either an id or a name is a pretty common practice in my experience (see the [chown command](http://www.man7.org/linux/man-pages/man1/chown.1.html) or `--user` in [docker run](https://docs.docker.com/engine/reference/commandline/run/)).

If the ctr demo path still works then I don't think this is blocking, but still something to consider. Besides the ability to support names in the future, making them a string would allow us to have default values that don't result in an error, which is nice even for the Go API case (as opposed to ctr). I don't really see any downside to doing it.

xibz:
> The change looks good. Can you add some tests that use non-default UID/GID?

@kzys - There is no longer a default UID and GID. However, I did update a test to use `300001` as the UID and GID if that was your concern.

kzys:
Ah, right. My concern was (iirc) that we are always using 300000/300000. In theory the test wouldn't catch any issues even if we hard-code them.

The concern is addressed by using 300001 on one of them. So LGTM!

Issue #367, opened 2019-12-11
Cleanup logs around io streams shutting down


sipsma:
Right now when a container exits you get some confusing error logs like this:
```
time="2019-12-11T20:03:38.436612731Z" level=error msg="error copying io" ExecID= TaskID=container-80-3 error="read unix @->firecracker.vsock: use of closed network connection" runtime=aws.firecracker stream=stdout vmID=80
```

or this:
```
time="2019-12-11T20:03:38.443037999Z" level=debug msg="[   81.326403] agent[848]: time=\"2019-12-11T20:03:38Z\" level=error msg=\"error copying io\" ExecID= TaskID=container-80-3 error=\"read /proc/self/fd/76: file already closed\" stream=stderr" jailer=noop runtime=aws.firecracker vmID=80 vmm_stream=stdout
```

I don't think these in particular should be logged at ERROR level and they should probably be clarified as they just indicate the stream got closed. Something more like `std<in,out,err> closed` at INFO level seems more appropriate. We can include the full actual error message at DEBUG still. We'll also want to make sure that if other unexpected errors occur, we continue to log those at ERROR.

Zyqsempai:
I can take this thingy;)

sipsma:
Sounds good, thanks @Zyqsempai !

FFMMM:
@kzys, can we close this given https://github.com/firecracker-microvm/firecracker-containerd/pull/369 ?

kzys:
I want to have some tests to prevent regressions.

PR #369, opened 2019-12-13
Change log level for copying io err

Zyqsempai:

Issue #367

*Description of changes:*
Changed error log level, to clean up logs.

### DIFF
--- a/internal/vm/ioproxy.go
+++ b/internal/vm/ioproxy.go
@@ -16,6 +16,7 @@ package vm
 import (
 	"context"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/firecracker-microvm/firecracker-containerd/internal"
@@ -129,7 +130,12 @@ func (connectorPair *IOConnectorPair) proxy(
 
 		_, err := io.CopyBuffer(writer, reader, make([]byte, internal.DefaultBufferSize))
 		if err != nil {
-			logger.WithError(err).Error("error copying io")
+			if strings.Contains(err.Error(), "use of closed network connection") ||
+				strings.Contains(err.Error(), "file already closed") {
+				logger.Infof("connection was closed: %v", err)
+			} else {
+				logger.WithError(err).Error("error copying io")
+			}
 			copyDone <- err
 		}
 	}()


internal/vm/ioproxy.go
```diff
@@ -129,7 +130,11 @@ func (connectorPair *IOConnectorPair) proxy(
 
 		_, err := io.CopyBuffer(writer, reader, make([]byte, internal.DefaultBufferSize))
 		if err != nil {
-			logger.WithError(err).Error("error copying io")
+			if strings.Contains(err.Error(), "use of closed network connection") {
```
kzys: While Go would [keep this string as a public interface](https://github.com/golang/go/blob/bdea352b96c6934ee49920b18c3b8aedffd27d08/src/internal/poll/fd.go#L20), checking the error by substring match doesn't seem right.

Can't we let `ioproxy` know that the status of the connection, rather than ignoring the error?

Zyqsempai: @sipsma What do you think?

sipsma: We are just operating on generic `io.Reader`s and `io.Writer`s, which is what allows this code to work equivalently across permutations of vsocks, unix sockets and fifos, so I'm not sure there's a straightforward way to just check the status of the connection as the "connection" could be any number of things.

I really don't like the whole method of doing error checking by string matching, but it's basically accepted as the way you have to do it in go at the moment (if I'm wrong or missing something please let me know). There's some hope with the recent error improvements in go 1.13, but to my knowledge that hasn't had enough time to soak into the various interfaces we use here, so for now string matching is okay with me.

sipsma: Another error we see logged sometimes when copying to/from a FIFO is:
```
time="2019-12-13T16:22:10.889084273Z" level=error msg="error copying io" ExecID= TaskID=TestShimExitsUponContainerDelete_Isolated-1576254128261884676 error="read /proc/self/fd/9: file already closed" runtime=aws.firecracker stream=stdin vmID=b5b59453-6630-4a72-96b7-e0d7a93600a7
```

Can we also handle that case similarly to `use of closed network connection`? Maybe just checking `file already closed` would be sufficient.

Zyqsempai:
@sipsma PTAL

internal/vm/ioproxy.go
```diff
@@ -129,7 +130,12 @@ func (connectorPair *IOConnectorPair) proxy(
 
 		_, err := io.CopyBuffer(writer, reader, make([]byte, internal.DefaultBufferSize))
 		if err != nil {
-			logger.WithError(err).Error("error copying io")
+			if strings.Contains(err.Error(), "use of closed network connection") ||
+				strings.Contains(err.Error(), "file already closed") {
+				logger.Infof("connection was closed: %s", err)
```
sipsma: nit: just noticed this, it's slightly better to use `%v` for errors since `error` is technically an interface. Otherwise LGTM, thanks!

Zyqsempai: Done

PR #379, opened 2020-01-16
Copy sparse files as spase files

kzys:
*Issue #, if available:*

NA

*Description of changes:*

Filesystem images are often big (multiple GBs) but sparse (have
zeros inside). At least mkfs.ext4 makes them as sparse files, but
io.Copy ignores the fact and copy all zeros as is, which is slow and
ineffecent.

This change adds sparse file support which makes runc_jailer's initial
file copy faster.

### DIFF
--- a/runtime/jailer_test.go
+++ b/runtime/jailer_test.go
@@ -15,11 +15,15 @@ package main
 
 import (
 	"context"
+	"io/ioutil"
 	"os"
+	"path/filepath"
+	"syscall"
 	"testing"
 
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/firecracker-microvm/firecracker-containerd/proto"
 )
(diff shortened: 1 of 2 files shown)


aaithal:
Is there additional test data that you can add here to make sure that this solution is as performant as `cp --sparse=always` ?

kzys:
It is faster than GNU cp! only if the sparse file is pretty much empty :(

I've added my benchmark commit, and here is the result on my EC2 box.

```
% go test -bench=BenchmarkCopy -benchtime=5s
...                
goos: linux
goarch: amd64
pkg: github.com/firecracker-microvm/firecracker-containerd/runtime
BenchmarkCopyGoEmptyFile-48               200000             46494 ns/op
BenchmarkCopyGoExt4-48                        20         579197267 ns/op
BenchmarkCopyGoNaiveEmptyFile-48              10         658009553 ns/op
BenchmarkCopyGoNaiveExt4-48                   10         682055838 ns/op
BenchmarkCopyExecEmptyFile-48              10000            975919 ns/op
BenchmarkCopyExecExt4-48                     300          18730759 ns/op
PASS
ok      github.com/firecracker-microvm/firecracker-containerd/runtime   62.182s
```

Apparently [GNU cp is using FIEMAP](https://lwn.net/Articles/789623/), whereas this Go implementation is using SEEK_HOLE and SEEK_DATA.

> But he did note that the way sparse files were handled using FIEMAP in cp was great. Ts'o pointed out that FIEMAP is a great example of how the process should work.

Regarding the benefits, `cp --sparse=always` is GNU only option and not supported by other cp's such as BusyBox. 

But I do agree that we should use cp when there is. Let me update the PR.

xibz:
@kzys - Are we ever going to be copying empty files though? I would much prefer `io.Copy` or the use of `cp` as they are much simpler implementations, and the difference between `io.Copy` and the sparse copy seems to be negligible in `ext4` files. Why are we optimizing this in the first place? Was there a bottle neck found?

nmeyerhans:
> @kzys - Are we ever going to be copying empty files though? I would much prefer `io.Copy` or the use of `cp` as they are much simpler implementations, and the difference between `io.Copy` and the sparse copy seems to be negligible in `ext4` files. Why are we optimizing this in the first place? Was there a bottle neck found?

We absolutely want to be optimizing this path. The time spent copying the rootfs image directly impacts VM launch latency. `cp` is probably enough of an optimization over direct handling of sparse files, and as @aaithal said, it's pretty well tested.

PR #380, opened 2020-01-16
Add support for running integ tests on bare metal

xibz:
This adds setupBareMetalTest helper function that will boostrap the test
with building firecracker-containerd and placing that in a random shim
directory. In addition, this function will also start
firecracker-containerd with the necessary configuration

*Issue #, if available:*

*Description of changes:*


.buildkite/pipeline.yml
```diff
@@ -35,7 +35,7 @@ steps:
   - label: ":lint-roller: loop device cleanup"
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
-      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      distro: "debian"
```
samuelkarp: Can you revert this?  We should be able to run all of the exact same steps on both distros without a different file.

Makefile
```diff
@@ -54,7 +54,7 @@ $(SUBDIRS):
 	$(MAKE) -C $@ EXTRAGOARGS=$(EXTRAGOARGS)
 
 %-in-docker:
-	docker run --rm -it \
+	docker run --rm \
```
samuelkarp: Do we no longer need an open stdin or a pty?

Makefile
```diff
@@ -325,7 +325,7 @@ tools/runc-builder-stamp: tools/docker/Dockerfile.runc-builder
 	touch $@
 
 $(RUNC_BIN): $(RUNC_DIR)/VERSION tools/runc-builder-stamp
-	docker run --rm -it --user $(UID) \
+	docker run --rm --user $(UID) \
```
samuelkarp: Same question.

runtime/integ_test.go
```diff
@@ -140,3 +145,159 @@ func runTask(ctx context.Context, c containerd.Container) (*commandResult, error
 		return nil, errors.New("context cancelled")
 	}
 }
+
+// setupBareMetalTest will run firecracker-containerd, the runtime shim, and
+// snapshotter. This will return the cleanup function needed to kill each
```
samuelkarp: The snapshotter is part of containerd at this point and shouldn't be run separately.  We've removed the external snapshotter implementation from this repository.

runtime/service_al2_integ_test.go
```diff
@@ -0,0 +1,54 @@
+// +build al2
```
samuelkarp: Why do we need a separate test for AL2?  We should be running the same tests everywhere.

tools/image-builder/Makefile
```diff
@@ -88,7 +88,6 @@ builder_stamp:
 %-in-docker: builder_stamp
 	docker run --rm \
 		--security-opt=apparmor=unconfined \
-		-it \
```
samuelkarp: Same question

runtime/service_al2_integ_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+const firecrackerTestBinary = "firecracker-v0.19.0"
+
+func TestMultipleVMs_AL2(t *testing.T) {
+	b := make([]byte, 16)
+	rand.Read(b)
+	uuid := base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(b)
+	baseDir := filepath.Join(os.TempDir(), "shim-base")
+	dir := filepath.Join(baseDir, uuid)
+	cleanupFns, err := setupBareMetalTest(dir, uuid, func(cfg *config.Config) {
```
xibz: Move this to be done on every test.

.buildkite/al2_pipeline.yml
```diff
@@ -0,0 +1,90 @@
...
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    env:
+      DOCKER_IMAGE_TAG: "$BUILDKITE_BUILD_NUMBER"
+      EXTRAGOARGS: "-race"
+    command:
+      - ./.buildkite/setup_al2.sh
+      - docker run --rm -v $PWD:/mnt debian:stretch-slim rm -rf /mnt/tools/image-builder/rootfs
```
kzys: Why do we need `docker run` for removing the directory which is mounted from the host?

.buildkite/al2_pipeline.yml
```diff
@@ -0,0 +1,90 @@
...
+      - sudo install -d -o root -g buildkite-agent -m 775 "/local/artifacts/$BUILDKITE_BUILD_NUMBER"
```
kzys: Does it run this agent?

.buildkite/al2_pipeline.yml
```diff
@@ -0,0 +1,90 @@
...
+    # With that we can now clean up loop devices before starting jobs to ensure
+    # we don't hit the loop device limit
+  - label: ":lint-roller: loop device cleanup"
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    command: 'sudo losetup -D'
+    concurrency: 1
+    concurrency_group: 'loop-device test'
```
kzys: We may be able to remove the lines since we don't have naive snapshotter anymore.

.buildkite/al2_pipeline.yml
```diff
@@ -0,0 +1,90 @@
...
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    env:
+      DOCKER_IMAGE_TAG: "$BUILDKITE_BUILD_NUMBER"
+    command:
+      - make tidy-in-docker
+      - test -z "$(git status --porcelain)" && exit 0 || git status && echo -e '\ngo.mod and/or go.sum differ from committed, please run "go mod tidy" and commit the updated files.\n' && exit 1
+    timeout_in_minutes: 1
```
kzys: Assuming that we will have another bare-metal pipelines in future, I'd like to keep this check only on our main (non-bare metal) pipeline since it won't be affected by the difference of distros. What do you think?

.buildkite/setup_al2.sh
```diff
@@ -0,0 +1,65 @@
...
+mkdir -p $devmapper_path
+mkdir -p $state_path
+
+echo "Resetting thinpool $uuid"
+./tools/thinpool.sh reset $uuid
+
+export INSTALLROOT=$dir
+export FIRECRACKER_CONTAINERD_RUNTIME_DIR=$dir
+echo "Running make"
+sudo make
```
kzys: make should work without sudo.

xibz:
Closing as this isn't ready for review yet. Will reopen once it is ready.

PR #381, opened 2020-01-21
Adding AL2 pipeline.yml

xibz:
This change adds an al2_pipeline.yml file that will be used to run tests
on AL2. Also, we have a series of scripts, one for setting up the AL2
environment to test against and the actual test runner itself.

### DIFF
--- a/.buildkite/al2_cleanup.sh
+++ b/.buildkite/al2_cleanup.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+source .buildkite/al2env.sh
+
+sudo rm -rf $dir
+./tools/thinpool.sh remove $unique_id
--- a/.buildkite/al2_pipeline.yml
+++ b/.buildkite/al2_pipeline.yml
@@ -1,0 +1,90 @@
+# Copyright 2018-2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may
+# not use this file except in compliance with the License. A copy of the
+# License is located at
+#
+# 	http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+steps:
+
+  - label: ":docker: Build"
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    env:
+      DOCKER_IMAGE_TAG: "$BUILDKITE_BUILD_NUMBER"
+      EXTRAGOARGS: "-race"
+    command:
+      - ./.buildkite/setup_al2.sh
+      - docker run --rm -v $PWD:/mnt debian:stretch-slim rm -rf /mnt/tools/image-builder/rootfs
+      - sudo install -d -o root -g buildkite-agent -m 775 "/local/artifacts/$BUILDKITE_BUILD_NUMBER"
+      - cp tools/image-builder/rootfs.img "/local/artifacts/$BUILDKITE_BUILD_NUMBER/"
+
+  - wait
+
+    # Global concurrency has been set to one in the website configuration.
+    # With that we can now clean up loop devices before starting jobs to ensure
+    # we don't hit the loop device limit
+  - label: ":lint-roller: loop device cleanup"
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    command: 'sudo losetup -D'
+    concurrency: 1
+    concurrency_group: 'loop-device test'
+
+  - wait
+
+  # Git history validation happens after the 'wait' step so it happens
+  # in parallel with the subsequent tests and does not prevent them
+  # from running in the event of a validation failure.
+  - label: 'git log validation'
+    command: './.buildkite/logcheck.sh'
+    # This should run in the same queue, but we don't care whether it runs on
+    # the same host.
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+
+  - label: ":protractor: verify proto"
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    env:
+      DOCKER_IMAGE_TAG: "$BUILDKITE_BUILD_NUMBER"
+    command:
+      - make proto
+      - test -z "$(git status --porcelain)" && exit 0 || git status && echo -e '\nGenerated protobuf code differs from committed, please run "make proto" and commit the updated files.\n' && exit 1
+    timeout_in_minutes: 1
+
+  - label: ":golang: go mod tidy"
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    env:
+      DOCKER_IMAGE_TAG: "$BUILDKITE_BUILD_NUMBER"
+    command:
+      - sudo make tidy-in-docker
+      - test -z "$(git status --porcelain)" && exit 0 || git status && echo -e '\ngo.mod and/or go.sum differ from committed, please run "go mod tidy" and commit the updated files.\n' && exit 1
+    timeout_in_minutes: 1
+    
+  - label: ":onion: al2 tests"
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    env:
+      NUMBER_OF_VMS: "5"
+    command:
+      - ./.buildkite/al2_test.sh
+    timeout_in_minutes: 10
+
--- a/.buildkite/al2_pipeline.yml
+++ b/.buildkite/al2_pipeline.yml
@@ -24,67 +24,28 @@ steps:
     command:
       - ./.buildkite/setup_al2.sh
       - docker run --rm -v $PWD:/mnt debian:stretch-slim rm -rf /mnt/tools/image-builder/rootfs
-      - sudo install -d -o root -g buildkite-agent -m 775 "/local/artifacts/$BUILDKITE_BUILD_NUMBER"
-      - cp tools/image-builder/rootfs.img "/local/artifacts/$BUILDKITE_BUILD_NUMBER/"
 
   - wait
 
-    # Global concurrency has been set to one in the website configuration.
-    # With that we can now clean up loop devices before starting jobs to ensure
-    # we don't hit the loop device limit
-  - label: ":lint-roller: loop device cleanup"
-    agents:
-      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
-      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
-      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
-    command: 'sudo losetup -D'
-    concurrency: 1
-    concurrency_group: 'loop-device test'
-
-  - wait
-
-  # Git history validation happens after the 'wait' step so it happens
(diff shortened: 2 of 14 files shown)


.buildkite/al2_pipeline.yml
```diff
@@ -0,0 +1,90 @@
...
+    # With that we can now clean up loop devices before starting jobs to ensure
+    # we don't hit the loop device limit
+  - label: ":lint-roller: loop device cleanup"
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    command: 'sudo losetup -D'
+    concurrency: 1
+    concurrency_group: 'loop-device test'
```
kzys: Are we still need to clean up them after removing naive snapshotter?

.buildkite/al2_pipeline.yml
```diff
@@ -0,0 +1,90 @@
...
+
+  - wait
+
+  # Git history validation happens after the 'wait' step so it happens
```
kzys: Assuming we are going to make some changes on these validation steps, I'd like to keep them only in the main pipeline, instead of running in here again.

.buildkite/al2env.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+shim_base=/tmp/shim-base
+uuid=$BUILDKITE_BUILD_NUMBER
+dir=$shim_base/$uuid
```
kzys: How about using $BUILDKITE_BUILD_NUMBER directly instead of putting that in $uuid? I would assume $uuid is hex-number-with-dashes, but this variable is not.

.buildkite/setup_al2.sh
```diff
@@ -0,0 +1,65 @@
...
+echo "Running make"
+sudo make
+echo "Running make install"
+sudo -E "INSTALLROOT=$INSTALLROOT" make install
+echo "Running make install-default-vmlinux"
+sudo make install-default-vmlinux
+echo "Running make image"
+sudo make image
+echo "Running make install-default-rootfs"
+sudo make install-default-rootfs
```
kzys: Can we remove `sudo` from non-install targets? They should work without being root.

.buildkite/al2_pipeline.yml
xibz: Yea, this was a copy pasta from our `pipeline.yml`. Looks like it wasn't removed from there for whatever reason.

.buildkite/al2env.sh
xibz: I changed it to `uid`. I much prefer that, since that signifies what it is suppose to be, a unique ID.

.buildkite/al2_pipeline.yml
xibz: Done

.buildkite/setup_al2.sh
xibz: Done

.buildkite/al2env.sh
samuelkarp: `uid`, in the context of Linux, would generally imply a User ID, not a unique ID.

xibz: Yea, there is some ambiguity there, perhaps `uniqueID`?

kzys: unique_id since other variables are snake_case?

kzys:
Aside from the lowerCamelCase variable and the fixup commit, looks good to me!

.buildkite/setup_al2.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+set -eu
+
+source ./.buildkite/al2env.sh
+
+echo "Creating necessary directories"
```
samuelkarp: I'd recommend removing all the `echo` statements and instead using `set -x` at the top.  This option will cause every command to be echoed out before it is run.

.buildkite/al2_cleanup.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+source .buildkite/al2env.sh
+
+sudo rm -rf $dir
+./tools/thinpool.sh remove $uniqueID
```
samuelkarp: Looks like [something is failing](https://buildkite.com/firecracker-microvm/firecracker-containerd-al2/builds/29#0c9f34ba-4a50-4839-9241-045c36b77718/34-36).

.buildkite/setup_al2.sh
```diff
@@ -0,0 +1,67 @@
...
+
+echo "Resetting thinpool $uniqueID"
+./tools/thinpool.sh reset $uniqueID
+
+export INSTALLROOT=$dir
+export FIRECRACKER_CONTAINERD_RUNTIME_DIR=$dir
+echo "Running make"
+make
+echo "Running make install"
+sudo -E "INSTALLROOT=$INSTALLROOT" make install
```
samuelkarp: Looks like [it can't find the go binary](https://buildkite.com/firecracker-microvm/firecracker-containerd-al2/builds/29#0c1d96de-cd1f-4ffd-903a-14ab8defe11c/50-94).

.buildkite/setup_al2.sh
```diff
@@ -0,0 +1,67 @@
...
+echo "Running make install-default-vmlinux"
+sudo make install-default-vmlinux
```
samuelkarp: We should already have the kernel binary on disk and don't need to download it, which is [what this target does](https://buildkite.com/firecracker-microvm/firecracker-containerd-al2/builds/29#0c1d96de-cd1f-4ffd-903a-14ab8defe11c/50-124).

.buildkite/setup_al2.sh
```diff
@@ -0,0 +1,67 @@
...
+echo "Running make image"
+make image
+echo "Running make install-default-rootfs"
+sudo make install-default-rootfs
```
samuelkarp: More [issues finding `go`](https://buildkite.com/firecracker-microvm/firecracker-containerd-al2/builds/29#0c1d96de-cd1f-4ffd-903a-14ab8defe11c/50-834).

.buildkite/al2_test.sh
```diff
@@ -0,0 +1,13 @@
...
+source ./.buildkite/al2env.sh
+
+export PATH=$bin_path:$PATH
+
+sudo -E "FIRECRACKER_CONTAINERD_RUNTIME_CONFIG_PATH=$runtime_config_path" $bin_path/firecracker-containerd --config $dir/config.toml &
+containerd_pid=$!
+sudo $bin_path/firecracker-ctr --address $dir/containerd.sock content fetch docker.io/library/alpine:3.10.1
+sudo -E env "PATH=$PATH" /usr/local/bin/go test -run TestMultipleVMs_Isolated ./...
+
+sudo kill -9 $containerd_pid
```
samuelkarp: That's certainly...forceful.  It would be nicer to send `SIGTERM` and let the process gracefully exit, but then you need to monitor it and make sure it actually does.  @nmeyerhans, do you have an opinion here?

.buildkite/al2env.sh
xibz: Fixed!

.buildkite/al2env.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+shim_base=/tmp/shim-base
```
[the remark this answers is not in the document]
xibz: When I use the buildkite checkout path env var, it'll silently fail on running install and not say why. So later when it tries to run the tests by starting up `firecracker-containerd` and `firecracker-ctr`, those binaries will not be present and cause the test to fail. So I went ahead and used `/tmp` instead.

.buildkite/hooks/pre-exit
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
-set -eu
+sudo make -C "$BUILDKITE_BUILD_CHECKOUT_PATH" clean
 
-make -C "$BUILDKITE_BUILD_CHECKOUT_PATH" clean
+# clean up ephemeral files since this will cause an error to build kite due to
+# these files being owned by root
+sudo make -C "$BUILDKITE_BUILD_CHECKOUT_PATH/tools/image-builder" distclean
```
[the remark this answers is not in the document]
xibz: This is due to sudo dropping `PATH` and using `secure_path` in `/etc/sudoers`. Unfortunately simply adding path does not fix this issue as well :(, and we need to update `/etc/sudoers` to get this to properly work.

.buildkite/al2_cleanup.sh
xibz: containerd cleans up the snap devices, but doesn't clean up the thinpool, since our script removes the snap devices and thin pool, we will always expect the snap device cleanup to fail unless containerd failed to clean it up.

.buildkite/setup_al2.sh
xibz: Fixed

.buildkite/setup_al2.sh
xibz: Done

.buildkite/al2_test.sh
xibz: I've wen't ahead and removed the `SIGKILL`.

.buildkite/setup_al2.sh
xibz: Fixed

.buildkite/al2_pipeline.yml
```diff
@@ -0,0 +1,53 @@
...
+
+  - wait
+
+  - label: ":onion: al2 tests"
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    env:
+      NUMBER_OF_VMS: "5"
```
kzys: How about changing the value from 5 to 100? Our main build is able to launch 100 VMs. So AL build should be able to launch even more, since the build doesn't have Docker.

xibz: Fixed

.buildkite/al2env.sh
```diff
@@ -0,0 +1,10 @@
...
+unique_id=$BUILDKITE_BUILD_NUMBER
+dir=$shim_base/$unique_id
+bin_path=$dir/bin
+devmapper_path=$dir/devmapper
+state_path=$dir/state
+runtime_config_path=$dir/firecracker-runtime.json
+firecracker_bin=firecracker-v0.19.0
```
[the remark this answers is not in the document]
xibz: My thinking was that these are more variables than constants. Bash standard is basically snake case for variables and upper casing for constants.

But I don't mind thinking that these are constants, either. Let me know what you think is best.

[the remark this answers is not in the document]
mxpv: right, we use (mostly) lower case for variables.

.buildkite/al2_pipeline.yml
```diff
@@ -0,0 +1,45 @@
...
+      distro: "${BUILDKITE_AGENT_META_DATA_DISTRO}"
+      hostname: "${BUILDKITE_AGENT_META_DATA_HOSTNAME}"
+    env:
+      DOCKER_IMAGE_TAG: "$BUILDKITE_BUILD_NUMBER"
+      EXTRAGOARGS: "-race"
+    command:
+      - ./.buildkite/setup_al2.sh
+      - docker run --rm -v $PWD:/mnt debian:stretch-slim rm -rf /mnt/tools/image-builder/rootfs
+      - sudo install -d -o root -g buildkite-agent -m 775 "/local/artifacts/$BUILDKITE_BUILD_NUMBER"
+      - cp tools/image-builder/rootfs.img "/local/artifacts/$BUILDKITE_BUILD_NUMBER/"
```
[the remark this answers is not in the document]
kzys: @xibz Is it answered?

runtime/service_integ_test.go
```diff
@@ -60,25 +61,33 @@ import (
...
 
-	containerdSockPath = "/run/containerd/containerd.sock"
-
 	firecrackerRuntime     = "aws.firecracker"
 	shimProcessName        = "containerd-shim-aws-firecracker"
 	firecrackerProcessName = "firecracker"
 	jailerProcessName      = "runc"
 
-	defaultVMRootfsPath = "/var/lib/firecracker-containerd/runtime/default-rootfs.img"
+	//defaultVMRootfsPath = "/var/lib/firecracker-containerd/runtime/default-rootfs.img"
```
kzys: nit: we should remove this line.

.buildkite/al2_pipeline.yml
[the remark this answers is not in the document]
xibz: I just removed them as we aren't using them for AL2 tests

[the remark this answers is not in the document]
kzys: Hmm, but you are using `/local/artifacts/$BUILDKITE_BUILD_NUMBER` here. Wasn't it Sam's question about?

xibz:
@kzys - Fixed the local files. Accidentally forgot to commit the `.buildkite` folder.

kzys:
I had retried the test since TestMultipleVMs_Isolated is a known issue. Strangely the re-run has been stuck for minutes... Is that something you can fix easily?

https://buildkite.com/firecracker-microvm/firecracker-containerd-al2/builds/168#2e89265c-829d-40b3-92a5-436d0100228b

If not, I'm fine merging this change as is.

xibz:
> I had retried the test since TestMultipleVMs_Isolated is a known issue. Strangely the re-run has been stuck for minutes... Is that something you can fix easily?

Weird. First time, I've seen that. I just rebuilt it and it seems to be fixed. However, the 148 exit code error is still occurring.

PR #386, opened 2020-02-07
Make integration tests work with loopback devices

kzys:
*Issue #, if available:*

On both "Getting Started" and "Quickstart", we explain the way to configure devmapper with loopback devices, but thinpool.sh assumes that a thinpool is backed by LVM.

*Description of changes:*

Make thinpool.sh work with loopback devices. Also remove snapshotter-related environment variables and if's.

### DIFF
--- a/runtime/integ_test.go
+++ b/runtime/integ_test.go
@@ -55,13 +55,8 @@ func shimBaseDir() string {
 	return defaultShimBaseDir
 }
 
-func defaultSnapshotterName() string {
-	if name := os.Getenv("FICD_SNAPSHOTTER"); name != "" {
-		return name
-	}
-
-	return "devmapper"
-}
+// devmapper is the only snapshotter we can use with Firecracker
+const defaultSnapshotterName = "devmapper"
 
 func prepareIntegTest(t *testing.T, options ...func(*config.Config)) {
 	t.Helper()
--- a/tools/thinpool.sh
+++ b/tools/thinpool.sh
@@ -1,4 +1,4 @@
-#! /bin/sh
+#! /bin/bash
 #
 # Copyright 2018-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
@@ -13,9 +13,9 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-set -eu
+FICD_DM_VOLUME_GROUP="$FICD_DM_VOLUME_GROUP"
 
-VOLUME_GROUP='fcci-vg'
+set -eu
 
 subcommand="$1"
 name="$2"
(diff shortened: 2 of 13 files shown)


tools/thinpool.sh
```diff
@@ -24,22 +24,51 @@ if [ -z "$name" ]; then
...
+             --poolmetadatasize 16GiB \
+             --size 1G \
+             -n "$name" "$FICD_DM_VOLUME_GROUP"
+    }
 
-pool_remove() {
-    sudo dmsetup remove "${dm_device}-snap-"* || true
-    sudo dmsetup remove \
+    pool_remove() {
+        sudo dmsetup remove "${dm_device}-snap-"* || true
```
mxpv: As far as I know, `dmsetup remove` does not support masks, so this will always fail.

tools/thinpool.sh
```diff
@@ -13,9 +13,9 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-set -eu
+FICD_DM_VOLUME_GROUP="$FICD_DM_VOLUME_GROUP"
```
mxpv: Is this assignment intentional ?

runtime/integ_test.go
```diff
@@ -45,13 +45,7 @@ var defaultRuntimeConfig = config.Config{
...
 }
 
-func defaultSnapshotterName() string {
-	if name := os.Getenv("FICD_SNAPSHOTTER"); name != "" {
-		return name
-	}
-
-	return "devmapper"
-}
+const defaultSnapshotterName = "devmapper"
```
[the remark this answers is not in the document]
kzys: How about adding some comments around this constant?

We can rename that to `defaultBlockDeviceSnapshotter` but it is leaking the implementation detail to the caller side. If callers need to be aware about the fact, shouldn't they pass "devmapper" directly?

tools/thinpool.sh
kzys: Sorry, which masks?

tools/thinpool.sh
kzys: Yes. If FICD_DM_VOLUME_GROUP is not defined, I'd like to treat that as an empty string.

tools/thinpool.sh
mxpv: `"${dm_device}-snap-"*`
As far as I understand this meant to be `dmsetup remove devicename-snap-*` ?

runtime/integ_test.go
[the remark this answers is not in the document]
mxpv: Comment works as well.

tools/thinpool.sh
kzys: Yeah, but `*` will be expanded by a shell. So dmsetup won't see `*`

PR #389, opened 2020-02-07
Make StopVM() synchronous

kzys:
*Issue #, if available:*

NA

*Description of changes:*

This PR makes StopVM() synchronous. We initially thought StopVM() should be asynchronous, and eventing should be used for synchronization, however

- CreateVM() is already synchronous. It waits the establishment of the vsock connection and guarantees that a VM is usable at the end of the request.
- StopVM() was somewhat synchronous even before this change. Since #362, StopVM() was waiting `<-s.shimCtx.Done()` before returning a response. So latency impact of this change is pretty minimal.
- It makes clients' life easier. Container orchestrators may create resources (e.g. filesystem images) before creating a VM. Then removing the resources must be happening after StopVM(). Doing synchronization inside StopVM() would make orchestrators's life easier.

So, now I think it would be better to make this call synchronous. In other words, StopVM() should guarantee the absence of the VMM process at the end of a request.

What do you think?

### DIFF
--- a/runtime/jailer.go
+++ b/runtime/jailer.go
@@ -56,7 +56,7 @@ type jailer interface {
 	// drive file
 	StubDrivesOptions() []FileOpt
 
-	// Stop the jailer as a way that the process can interpreted (e.g. SIGTERM).
+	// Stop the jailer as a way that is visible from the user-level process (e.g. SIGTERM).
 	Stop() error
 
 	// Close will do any necessary cleanup that the jailer has accrued.
--- a/runtime/jailer.go
+++ b/runtime/jailer.go
@@ -55,6 +55,10 @@ type jailer interface {
 	// StubDrivesOptions will return a set of options used to create a new stub
 	// drive file
 	StubDrivesOptions() []FileOpt
+
+	// Stop the jailer as a way that the process can interpreted (e.g. SIGTERM).
+	Stop() error
+
 	// Close will do any necessary cleanup that the jailer has accrued.
 	Close() error
 }
--- a/runtime/noop_jailer.go
+++ b/runtime/noop_jailer.go
@@ -15,8 +15,11 @@ package main
 
 import (
 	"context"
+	"os"
+	"syscall"
 
 	"github.com/firecracker-microvm/firecracker-go-sdk"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	"github.com/firecracker-microvm/firecracker-containerd/config"
@@ -29,17 +32,19 @@ type noopJailer struct {
 	logger  *logrus.Entry
 	shimDir vm.Dir
 	ctx     context.Context
+	pid     int
 }
 
-func newNoopJailer(ctx context.Context, logger *logrus.Entry, shimDir vm.Dir) noopJailer {
-	return noopJailer{
+func newNoopJailer(ctx context.Context, logger *logrus.Entry, shimDir vm.Dir) *noopJailer {
+	return &noopJailer{
 		logger:  logger,
 		shimDir: shimDir,
 		ctx:     ctx,
+		pid:     0,
 	}
 }
 
-func (j noopJailer) BuildJailedMachine(cfg *config.Config, machineConfig *firecracker.Config, vmID string) ([]firecracker.Opt, error) {
+func (j *noopJailer) BuildJailedMachine(cfg *config.Config, machineConfig *firecracker.Config, vmID string) ([]firecracker.Opt, error) {
 	if len(cfg.FirecrackerBinaryPath) == 0 {
 		return []firecracker.Opt{}, nil
 	}
@@ -59,25 +64,58 @@ func (j noopJailer) BuildJailedMachine(cfg *config.Config, machineConfig *firecr
 		cmd.Stderr = j.logger.WithField("vmm_stream", "stderr").WriterLevel(logrus.DebugLevel)
 	}
 
+	pidHandler := firecracker.Handler{
+		Name: "firecracker-containerd-jail-pid-handler",
+		Fn: func(ctx context.Context, m *firecracker.Machine) error {
+			pid, err := m.PID()
+			if err != nil {
+				return err
+			}
+			j.pid = pid
+			return nil
+		},
+	}
+
 	j.logger.Debug("noop operation for BuildJailedMachine")
 	return []firecracker.Opt{
 		firecracker.WithProcessRunner(cmd),
+		func(m *firecracker.Machine) {
+			m.Handlers.FcInit = m.Handlers.FcInit.Append(pidHandler)
+		},
 	}, nil
 }
 
-func (j noopJailer) JailPath() vm.Dir {
+func (j *noopJailer) JailPath() vm.Dir {
 	j.logger.Debug("noop operation returning shim dir for JailPath")
 	return j.shimDir
 }
 
-func (j noopJailer) ExposeFileToJail(path string) error {
+func (j *noopJailer) ExposeFileToJail(path string) error {
 	j.logger.Debug("noop operation for ExposeFileToJail")
 	return nil
 }
 
-func (j noopJailer) StubDrivesOptions() []FileOpt {
+func (j *noopJailer) StubDrivesOptions() []FileOpt {
 	j.logger.Debug("noop operation for StubDrivesOptions")
 	return []FileOpt{}
 }
 
-func (j noopJailer) Close() error { return nil }
+func (j *noopJailer) Stop() error {
+	if j.pid == 0 {
+		return errors.New("the machine hasn't been started")
+	}
+
+	j.logger.Debugf("sending SIGTERM to %d", j.pid)
+	p, err := os.FindProcess(j.pid)
+	if err != nil {
+		return err
+	}
+
+	err = p.Signal(syscall.SIGTERM)
+	if err == nil || err.Error() == "os: process already finished" {
+		return nil
+	}
+	return err
+}
+
+func (j *noopJailer) Close() error { return nil }
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -1187,7 +1187,7 @@ func (s *service) shutdown(
 	if err == nil {
 		return nil
 	}
-	return status.Error(codes.DeadlineExceeded, fmt.Sprintf("the VMM was killed forcibly: %v", err))
+	return status.Error(codes.Internal, fmt.Sprintf("the VMM was killed forcibly: %v", err))
 }
 
 // shutdownLoop sends multiple different shutdown requests to stop the VMM.
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -74,6 +74,7 @@ const (
 	firecrackerStartTimeout = 5 * time.Second
 	defaultStopVMTimeout    = 5 * time.Second
 	defaultShutdownTimeout  = 5 * time.Second
+	jailerStopTimeout       = 3 * time.Second
 
 	// StartEventName is the topic published to when a VM starts
 	StartEventName = "/firecracker-vm/start"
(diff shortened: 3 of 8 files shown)


kzys:
Here is the error I want to fix: https://buildkite.com/firecracker-microvm/firecracker-containerd/builds/1530#f278594b-e776-44f3-991f-8ffe29eb9e7b

kzys:
cc @haikuoliu

runtime/service.go
```diff
@@ -80,6 +80,8 @@ const (
 
 	// StopEventName is the topic published to when a VM stops
 	StopEventName = "/firecracker-vm/stop"
+
+	sigtermTimeout = 3 * time.Second
```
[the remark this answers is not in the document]
kzys: Sounds good. Will do.

runtime/service.go
```diff
@@ -1183,42 +1185,55 @@ func (s *service) shutdown(
 	shutdownCtx, cancel := context.WithTimeout(requestCtx, timeout)
 	defer cancel()
 	go func() {
-		// Once the shutdown procedure is done, the shim needs to shutdown too.
-		// This also ensures that if the VMM is still alive, it will receive a
-		// SIGKILL via exec.CommandContext
-		<-shutdownCtx.Done()
+		// Once the firecracker process is complete, the shim needs to shutdown too.
+		s.machine.Wait(context.Background())
```
mxpv: Do we want to log error here just in case?

runtime/service.go
```diff
@@ -1183,42 +1185,55 @@ func (s *service) shutdown(
...
 	}
 
-	return nil
+	// 3) If that fails, cancelling the context to send SIGKILL
+	s.shimCancel()
+	err = s.machine.Wait(context.Background())
+	if err != nil {
+		s.logger.WithError(err).Error("failed to wait the VMM")
+	}
+	return status.Error(codes.DeadlineExceeded, fmt.Sprintf("the VMM was killed by SIGKILL: %s", err))
```
mxpv: nit: `%v`

runtime/service.go
```diff
@@ -1183,42 +1185,55 @@ func (s *service) shutdown(
...
+	if waitCtx.Err() != context.DeadlineExceeded {
+		return status.Error(codes.DeadlineExceeded, fmt.Sprintf("the VMM was killed by SIGTERM: %s", err))
 	}
 
-	return nil
+	// 3) If that fails, cancelling the context to send SIGKILL
+	s.shimCancel()
+	err = s.machine.Wait(context.Background())
+	if err != nil {
+		s.logger.WithError(err).Error("failed to wait the VMM")
```
mxpv: `...after cancelling the context` to distinguish from `failed to wait the VMM` error message above?

runtime/service.go
```diff
@@ -1183,42 +1185,55 @@ func (s *service) shutdown(
...
+	if err != nil {
+		s.logger.WithError(err).Error("failed to stop the VM")
 	}
 
-	if s.vmExitErr != nil {
-		return status.Error(codes.Internal,
-			fmt.Sprintf("VMM exit errors: %v", s.vmExitErr))
+	err = s.machine.Wait(waitCtx)
+	if waitCtx.Err() != context.DeadlineExceeded {
+		return status.Error(codes.DeadlineExceeded, fmt.Sprintf("the VMM was killed by SIGTERM: %s", err))
```
mxpv: err nil be nil according to buildskite: https://buildkite.com/firecracker-microvm/firecracker-containerd/builds/1561#d92a6ab9-cc53-4f7d-bef5-deb392340bbc/370-396

kzys: This is strange, since we are killing the VMM.

runtime/service.go
kzys: It may double-log the error, but since err would be nil most of the case, I think it is better to double-log rather than ignoring. So will do.

runtime/service.go
kzys: Done

runtime/service.go
kzys: Done

runtime/service.go
```diff
@@ -1183,42 +1184,58 @@ func (s *service) shutdown(
...
-			s.logger.WithError(stopVMMErr).Error("failed to forcibly stop VMM")
-			cancel()
+	// 1) Tell agent to exit, causing the VM to begin a reboot.
+	_, err := s.agentClient.Shutdown(shutdownCtx, req)
+	if err != nil {
+		s.logger.WithError(err).Error("failed to shutdown VM agent")
+	} else {
+		// If the agent can take the request, wait a moment to gracefully shutdown.
+		err = s.machine.Wait(shutdownCtx)
+		if err != nil {
```
kzys: TODO: I probably should handle context.Cancelled as well. Since we are shutting down the agent, the shim context (== firecracker process's context?) would be cancelled by the termination of the firecracker process.

runtime/service_integ_test.go
```diff
@@ -1452,17 +1469,48 @@ func TestStopVM_Isolated(t *testing.T) {
...
+			// shim guarantees that the underlying Firecracker process is dead
+			// (either gracefully or forcibly) at the end of the request.
+			if test.withStopVM {
+				fcExists, err := process.PidExists(fcProcess.Pid)
+				require.NoError(err, "failed to find firecracker")
+				require.False(fcExists, "firecracker %s is still there", vmID)
 			}
 
+			// Since the shim is the one which is writing the response,
+			// it cannot guarantee that the itself is dead at the end of the request.
```
IRCody: nit: "that the [shim] itself" ?

runtime/service.go
```diff
@@ -1183,42 +1184,58 @@ func (s *service) shutdown(
...
+		return status.Error(codes.DeadlineExceeded, fmt.Sprintf("the VMM was killed by SIGTERM: %v", err))
+	}
+
+	// 3) If that fails, cancelling the context to send SIGKILL
+	s.shimCancel()
+	err = s.machine.Wait(context.Background())
+	if err != nil {
+		s.logger.WithError(err).Error("failed to wait the VMM after cancelling the context")
+	}
+	return status.Error(codes.DeadlineExceeded, fmt.Sprintf("the VMM was killed by SIGKILL: %v", err))
```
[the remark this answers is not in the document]
kzys: Yeah, I agree that DeadlineExceeded is a bit odd, since the specified action (stopping the VM) is executed anyway. What I wanted to tell were

- The shutdown process is exceeding the customer-specified timeout.
- Because of that, the firecracker process is killed by an undesirable way.

runtime/noop_jailer.go
```diff
@@ -80,4 +98,12 @@ func (j noopJailer) StubDrivesOptions() []FileOpt {
 	return []FileOpt{}
 }
 
+func (j noopJailer) Stop() error {
+	p, err := os.FindProcess(j.pid)
```
xibz: What should we do if `j.pid` has never been set and the value is `0`? Probably want to return some sort of error?

runtime/service.go
```diff
@@ -1182,43 +1183,39 @@ func (s *service) shutdown(
...
+			s.logger.WithError(err).Error("failed to stop the jailer")
+		} else {
+			time.Sleep(defaultSigtermTimeout)
+		}
 
-	if s.vmExitErr != nil {
-		return status.Error(codes.Internal,
-			fmt.Sprintf("VMM exit errors: %v", s.vmExitErr))
-	}
+		s.logger.Debug("SIGKILL")
```
xibz: This is a little odd in terms of logging. Users are going to see `SIGTERM`, then always followed by `SIGKILL`. While this will `kill` runc, it really isn't a `SIGKILL`, but a delete of a container. We may want to actually send a `SIGKILL` then follow up with the `Close` function, if `SIGTERM` failed.

runtime/service_integ_test.go
kzys: Thanks. Fixed.

runtime/service.go
```diff
@@ -1180,44 +1181,63 @@ func (s *service) shutdown(
...
+			timeout: defaultSigtermTimeout,
+		},
+	}
+
+	for _, action := range actions {
+		s.logger.Debug(action.name)
+		err := action.shutdown()
+		if err != nil {
+			s.logger.WithError(err).Errorf("failed to %s", action.name)
+		} else if action.timeout != 0 {
```
xibz: Can this `else if` be just an `if` statement separate from the one above. They aren't really related.

runtime/noop_jailer.go
kzys: I agree. Fixed.

runtime/service.go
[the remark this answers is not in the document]
IRCody: Looking at the other codes the only two that seemed like maybe-appropriate were Unknown and Internal. Do either of those seem appropriate here?

runtime/service.go
```diff
@@ -1180,44 +1181,69 @@ func (s *service) shutdown(
...
+			time.Sleep(action.timeout)
+		}
+		pid, err := s.machine.PID()
+		if pid == 0 && err != nil {
```
xibz: Should we move this to the top? In the event the machine hasn't started, this would just break the loop.

runtime/service.go
kzys: I'm going to remove SIGTERM and SIGKILL from service.go since it is hidden under jailers mostly.

runtime/service.go
kzys: It is related. If sending a request is not succeeding, I'd like to skip `time.Sleep()`

runtime/service.go
kzys: Sounds good. Will do.

runtime/service.go
IRCody: I think we could get rid of the if part and just let time.Sleep() be called since it's safe to call with a 0 duration.

runtime/service.go
```diff
@@ -1183,42 +1184,58 @@ func (s *service) shutdown(
...
-	// fails, try to forcibly stop the VMM. If that too fails, just cancel
-	// the shutdownCtx to fast-path to the VMM getting SIGKILL.
-	_, shutdownErr := s.agentClient.Shutdown(shutdownCtx, req)
-	if shutdownErr != nil {
-		s.logger.WithError(shutdownErr).Error("failed to shutdown VM agent")
-		stopVMMErr := s.machine.StopVMM()
-		if stopVMMErr != nil {
-			s.logger.WithError(stopVMMErr).Error("failed to forcibly stop VMM")
-			cancel()
+	// 1) Tell agent to exit, causing the VM to begin a reboot.
```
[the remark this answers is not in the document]
kzys: I have extracted the killing part as shutdownLoop() which is much shorter than the original implementation. What do you think?

runtime/service.go
kzys: I thought Go would do something but you are right. Will remove the if.

https://golang.org/src/runtime/time.go?h=timeSleep#L84

runtime/service.go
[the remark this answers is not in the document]
kzys: I'm fine with `Internal`. Let me update the implementation.

runtime/service.go
kzys: Done.

runtime/service.go
kzys: Done.

runtime/service.go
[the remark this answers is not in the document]
kzys: Done.

runtime/service.go
kzys: Done.

runtime/service.go
```diff
@@ -1171,53 +1172,80 @@ func (s *service) Shutdown(requestCtx context.Context, req *taskAPI.ShutdownRequ
...
-	}
+// shutdownLoop sends multiple different shutdown requests to stop the VMM.
+// 1) send a request to the in-VM agent, which is presumed to cause the VM to begin a reboot.
+// 2) stop the VM through jailer#Stop(). The signal should be visible from the VMM (e.g. SIGTERM)
+// 3) stop the VM through cancelling the associated context. The signal would not be visible from the VMM (e.g. SIGKILL)
+func (s *service) shutdownLoop(
+	requestCtx context.Context,
+	timeout time.Duration,
+	req *taskAPI.ShutdownRequest,
+) error {
```
IRCody: Will this ever return an error?

runtime/service.go
```diff
@@ -1171,53 +1172,80 @@ func (s *service) Shutdown(requestCtx context.Context, req *taskAPI.ShutdownRequ
...
+				s.shimCancel()
+				return nil
+			},
+		},
+	}
+
+	for _, action := range actions {
+		pid, err := s.machine.PID()
+		if pid == 0 && err != nil {
+			break // we have nothing to kill
```
IRCody: Is it worth logging something here? This is an unexpected state I think?

runtime/service.go
kzys: Good point. It won't.

runtime/service.go
kzys: On the first iteration of the lopp, yes. But after issuing shutdown commands, it would be `pid == 0` and then we can cancel the rest of shutdown commands.

runtime/jailer.go
```diff
@@ -55,6 +55,10 @@ type jailer interface {
 	// StubDrivesOptions will return a set of options used to create a new stub
 	// drive file
 	StubDrivesOptions() []FileOpt
+
+	// Stop the jailer as a way that is visible from the user-level process (e.g. SIGTERM).
+	Stop() error
```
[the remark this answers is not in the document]
kzys: runc has `kill` (called from our Stop()) and `delete` (called from our Close()). The former is sending SIGTERM and the latter is sending SIGKILL and removing some files.

https://github.com/opencontainers/runc/blob/e6555cc01a92b599bef90dbe8cb3b7bb74391da9/delete.go#L61

I'm fine renaming them, but combining them would be hard.

runtime/service.go
kzys: I haven't seen this case anymore.

runtime/service.go
kzys: s.machine.Wait() is now taking context.Background() which won't be cancelled.

runtime/service.go
```diff
@@ -1183,42 +1184,58 @@ func (s *service) shutdown(
...
 		}
 	}
 
-	// wait for the shimCtx to be done, which means the VM has exited and we're ready
-	// to shutdown
-	<-s.shimCtx.Done()
-	if shutdownCtx.Err() == context.DeadlineExceeded {
-		return status.Error(codes.DeadlineExceeded,
-			"timed out waiting for VM shutdown, VMM was sent SIGKILL")
+	// If Wait() returns before the deadline, the VMM might be killed by someone else
```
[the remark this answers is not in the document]
kzys: I have changed the structure.

runtime/jailer.go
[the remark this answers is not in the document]
IRCody: Ok. I was thinking that close() with the force flag was switching between SIGTERM and SIGKILL. My Mistake!

kzys:
@xibz @mxpv Can you take a look once again? The code is changed a lot after your reviews.

PR #390, opened 2020-02-07
Jailer leaves containers behind

xibz:
The jailer will leave containers behind if the jailing process was
killed with SIGKILL.

### DIFF
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -1278,8 +1278,16 @@ func (s *service) Cleanup(requestCtx context.Context) (*taskAPI.DeleteResponse, 
 }
 
 func (s *service) monitorVMExit() {
-	// once the VM shuts down, the shim should too
-	defer s.shimCancel()
+	defer func() {
+		// we ignore the error here due to cleanup will only succeed if the jailing
+		// process was killed via SIGKILL
+		if err := s.jailer.Close(); err != nil {
+			s.logger.WithError(err).Error("failed to close jailer")
+		}
+
+		// once the VM shuts down, the shim should too
+		s.shimCancel()
+	}()
 
 	// Block until the VM exits
 	s.vmExitErr = s.machine.Wait(s.shimCtx)
--- a/runtime/service_integ_test.go
+++ b/runtime/service_integ_test.go
@@ -35,6 +35,7 @@ import (
 	"github.com/containerd/containerd/oci"
 	"github.com/containerd/containerd/pkg/ttrpcutil"
 	"github.com/containerd/containerd/runtime"
+	"github.com/containerd/go-runc"
 	"github.com/containerd/typeurl"
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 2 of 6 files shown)


runtime/service.go
```diff
@@ -565,7 +565,12 @@ func (s *service) mountDrives(requestCtx context.Context) error {
 // If the VM has not been created yet and the timeout is hit waiting for it to exist, an error will be returned
 // but the shim will continue to shutdown.
 func (s *service) StopVM(requestCtx context.Context, request *proto.StopVMRequest) (_ *empty.Empty, err error) {
-	defer logPanicAndDie(s.logger)
+	defer func() {
+		// we ignore the error here due to cleanup will only succeed if the jailing
+		// process was killed via SIGKILL
+		s.jailer.Close()
+		logPanicAndDie(s.logger)
+	}()
```
kzys: Signals can be coming from anywhere. If calling Close() is necessary, how about putting the invocation in monitorVMExit()?

xibz: Done. Moved to `monitorVMExit`

runtime/service.go
```diff
@@ -1278,8 +1278,14 @@ func (s *service) Cleanup(requestCtx context.Context) (*taskAPI.DeleteResponse,
 }
 
 func (s *service) monitorVMExit() {
-	// once the VM shuts down, the shim should too
-	defer s.shimCancel()
+	defer func() {
+		// we ignore the error here due to cleanup will only succeed if the jailing
+		// process was killed via SIGKILL
+		s.jailer.Close()
```
kzys: So if the jailing process (firecracker) was not killed by SIGKILL, would `Close()` return an error? Can we specifically check the known, expected error and log other errors?

runtime/service_integ_test.go
```diff
@@ -1364,6 +1364,34 @@ func TestStopVM_Isolated(t *testing.T) {
...
+
+				// Sleep here to ensure runc finishes execution
+				time.Sleep(100 * time.Millisecond)
+
+				// ensure that the jailer has cleaned up all of the containers
+				b, err := exec.Command("runc", "list").Output()
+				require.NoError(err, "failed to run 'runc list'")
+				lines := strings.Split(string(b), "\n")
+				expectedLines := 2
+				assert.Equal(expectedLines, len(lines), string(b))
```
kzys: * You can use `runc list --quiet` to remove the header
* Shouldn't be 1 (only the header)?. Why do we have 2 lines?

xibz: Naw, two. For example the string "foo\n", will split into ["foo", ""]

runtime/service.go
xibz: Yea, it would return a generic "exit status 1" error. Which isn't too useful. We can do a string compare but it'd be pretty hacky. My idea would be to capture the `Stderr` from the command and do a string compare and return a better error instead.

What do you think?

kzys: How about using `runc delete -f`?

xibz: I've went ahead and added a `errContainerDoesNotExist` error to check against.

kzys:
Looks good to me. Please fix squash the commits.

xibz:
@kzys - Squashed!

runtime/service_integ_test.go
mxpv: Can we use `go-runc` to interact with runc (same applies for `Close` above) ? It provides high level interface to runc, does output parsing as needed as we already have it pulled in.

PR #393, opened 2020-02-19
Run TestMultipleVMs_Isolated on Docker-based build again

kzys:
*Issue #, if available:*

NA

*Description of changes:*

Seems like we accidentally disabled the test on the Docker-based build.

### DIFF
--- a/runtime/service_integ_test.go
+++ b/runtime/service_integ_test.go
@@ -17,7 +17,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


kzys:
@xibz I'd like to remove our custom runner rather than fixing the issue, because
- For this particular test, we can use `defer` instead
- Because of this comment: https://golang.org/src/testing/testing.go?s=30908:30955#L1074

xibz:
@kzys - the reason why I didnt do that is it seems it would change the test no? Now you arent concurrently running n number of containers.

kzys:
At the end of testMultipleExecs, it is killing and deleting tasks, and "tasks" in containerd's world refer running containers. So this change would not change the number of concurrently running containers.

PR #396, opened 2020-02-21
Added guidelines for scaling firecracker VMs count

ustiugov:
I have put together a set of guidelines to configure the host OS to scale the number of microVMs past one thousand (tested with 1500 microVMs on a host with 256GB memory). Feedback is welcome.

### DIFF
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -1,0 +1,50 @@
+# Scaling the number of Firecracker microVMs per host
+
+To scale the number of microVMs past one thousand, one needs to properly configure 
+the system constraints and provision enough networking resources for the microVMs.
+
+On Ubuntu 18.04 Linux, one needs to properly configure the number of processes, threads, memory, 
+open files that may simultaneously exist in a system as well as create enough 
+virtual bridges (one bridge can serve up to 1023 interfaces).
+
+To configure the system, one needs to set up the following parameters. 
+Note that the exact values depend on your setting.
+
+In `/etc/security/limits.conf`, set `nofile`, `nproc` and `stack` to the 
+appropriate values for both normal users and root:
+```
+* soft nofile 1000000
+* hard nofile 1000000
+root soft nofile 1000000
+root hard nofile 1000000
+* soft nproc 4000000
+* hard nproc 4000000
+root soft nproc 4000000
+root hard nproc 4000000
+* soft stack 65536
+* hard stack 65536
+root soft stack 65536
+root hard stack 65536
+```
+
+Additionally, one needs to provision the ARP cache to avoid garbage collection.
+```
+sudo sysctl -w net.ipv4.neigh.default.gc_thresh1=1024
+sudo sysctl -w net.ipv4.neigh.default.gc_thresh2=2048
+sudo sysctl -w net.ipv4.neigh.default.gc_thresh3=4096
+sudo sysctl -w net.ipv4.ip_local_port_range="32769 65535"
+```
+
+Also, configure the maximum number of processes and threads in the system.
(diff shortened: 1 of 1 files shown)


docs/scaling.md
```diff
@@ -0,0 +1,50 @@
+# Scaling the number of Firecracker microVMs per host
+
+To scale the number of microVMs past one thousand, one needs to properly configure 
+the system contraints and provision enough networking resources for the microVMs.
```
IRCody: constraints

docs/scaling.md
```diff
@@ -0,0 +1,50 @@
...
+
+On Ubuntu 18.04 Linux, one needs to the number of processes, threads, memory,
```
IRCody: "one needs to [know] the number"?

ustiugov: corrected

docs/scaling.md
ustiugov: corrected

ustiugov:
@IRCody Thank you for the corrections, please check.

docs/scaling.md
```diff
@@ -0,0 +1,50 @@
...
+sudo sh -c "echo \"root soft nofile 1000000\" >> /etc/security/limits.conf"
+sudo sh -c "echo \"root hard nofile 1000000\" >> /etc/security/limits.conf"
+sudo sh -c "echo \"* soft nproc 4000000\" >> /etc/security/limits.conf"
+sudo sh -c "echo \"* hard nproc 4000000\" >> /etc/security/limits.conf"
+sudo sh -c "echo \"root soft nproc 4000000\" >> /etc/security/limits.conf"
+sudo sh -c "echo \"root hard nproc 4000000\" >> /etc/security/limits.conf"
+sudo sh -c "echo \"* soft stack 65536\" >> /etc/security/limits.conf"
+sudo sh -c "echo \"* hard stack 65536\" >> /etc/security/limits.conf"
+sudo sh -c "echo \"root soft stack 65536\" >> /etc/security/limits.conf"
+sudo sh -c "echo \"root hard stack 65536\" >> /etc/security/limits.conf"
```
kzys: For this section, it would make sense to show that as a file, rather then the sequence of `echo` commands like https://github.com/firecracker-microvm/firecracker-containerd/blob/master/docs/getting-started.md#configure-firecracker-containerd-binary.

Regarding the configuration itself, I have a few questions.

- Do we need to have `root` and `*` separately? Doesn't `*` cover `root`?
- Assuming that Firecracker is running as a separate Linux user, not root, can we specifically target the user instead of essentially removing the limits for all of them?

ustiugov: @kzys Thank you for the feedback, I have applied it. 

- root and the rest of the users are managed separately in Ubuntu, please see [here](https://stackoverflow.com/questions/38991351/why-in-etc-security-limits-conf-doesnt-include-root-user) for details. 
- The orchestrator (like the one in `examples/taskworkflow.go`) runs with sudo (at least in the quickstart/getting-started example commands) and it creates a lot of threads and processes As a result, it can hit one of the system limits as well as VMs themselves.

IRCody:
Thanks for the update @ustiugov. The CI is failing on checking DCO/subject length. I think if you amend your commit and add a line between your commit subject and the Signed-off-by line it will pass.

docs/scaling.md
kzys: Thanks! I didn't know that `*` doesn't include `root`. 

I can understand why we need more `nofile` and `nproc`, but not so sure about `stack`. Does something consume more stack by having more VMs?

PR #398, opened 2020-02-29
Benchmark CreateVM and StopVM

kzys:
*Issue #, if available:*

NA

*Description of changes:*

Benchmark CreateVM and StopVM.

Things I'd like to discuss;
- I'm running both CreateVM and StopVM inside the loop, mainly because it makes running them under Go's benchmark library easier, but it is not ideal.
- Moreover, should we use Go's benchmark library? We may want to know p50, p90, p100, ...

### DIFF
--- a/runtime/benchmark_test.go
+++ b/runtime/benchmark_test.go
@@ -16,6 +16,7 @@ package main
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/namespaces"
@@ -26,7 +27,32 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func benchmarkCreateAndStopVM(b *testing.B, vcpuCount uint32, kernelArgs string) {
+func createAndStopVM(
+	ctx context.Context,
+	fcClient fccontrol.FirecrackerService,
+	request proto.CreateVMRequest,
+) error {
+	uuid, err := uuid.NewV4()
+	if err != nil {
+		return err
+	}
+
+	request.VMID = uuid.String()
+
+	_, err = fcClient.CreateVM(ctx, &request)
+	if err != nil {
+		return err
+	}
+
+	_, err = fcClient.StopVM(ctx, &proto.StopVMRequest{VMID: request.VMID})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func benchmarkCreateAndStopVM(b *testing.B, vcpuCount uint32, kernelArgs string, parallel int) {
 	require := require.New(b)
 
 	client, err := containerd.New(containerdSockPath, containerd.WithDefaultRuntime(firecrackerRuntime))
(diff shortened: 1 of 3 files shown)


kzys:
You can run that like below. Yes, the Makefile target is hacky...

```
% make -C runtime integ-test-TestStopVM 'EXTRAGOARGS=-bench=BenchmarkCreateAndStopVM -benchtime=5x'
make: Entering directory '/home/katokazu/ec2-ws/firecracker-containerd/runtime'
/home/katokazu/ec2-ws/firecracker-containerd/runtime/../tools/thinpool.sh reset ""
docker run --rm -it \
        --privileged \
        --ipc=host \
        --network=none \
        --volume /dev:/dev \
        --volume /run/udev/control:/run/udev/control \
        --volume /home/katokazu/ec2-ws/firecracker-containerd/runtime/logs:/var/log/firecracker-containerd-test \
        --volume /home/katokazu/ec2-ws/firecracker-containerd/runtime/..:/src \
        --volume gocache:/go \
        --env ENABLE_ISOLATED_TESTS=1 \
        --env FICD_DM_VOLUME_GROUP= \
        --env FICD_DM_POOL= \
        --env GOPROXY=direct \
        --env GOSUMDB=off \
        --env NUMBER_OF_VMS= \
        --workdir="/src/runtime" \
        --init \
        localhost/firecracker-containerd-test:latest \
        "go test -bench=BenchmarkCreateAndStopVM -benchtime=5x -run '^TestStopVM$'"
goos: linux
goarch: amd64
pkg: github.com/firecracker-microvm/firecracker-containerd/runtime
BenchmarkCreateAndStopVM_Vcpu1_Isolated-96             5        1640368469 ns/op
BenchmarkCreateAndStopVM_Vcpu5_Isolated-96             5        1697239668 ns/op
BenchmarkCreateAndStopVM_Quiet_Isolated-96             5        1132619387 ns/op
PASS
ok      github.com/firecracker-microvm/firecracker-containerd/runtime   26.745s
make: Leaving directory '/home/katokazu/ec2-ws/firecracker-containerd/runtime'
```

* Avoiding Firecracker's console device and making Linux generally quiet improves the boot latency
* Interestingly, adding vCPUs somehow negatively impacts the boot latency.

ustiugov:
I would like to suggest to add benchmarking parallel start and stop VMs in two parallelized loops (e.g., using goroutines), probably in addition to the start/stop latency benchmark suggested by @kzys. 

In my experience, there are problems that appear when issuing **many** start and stop VM requests simultaneously (e.g., sporadic timeouts, hitting host limits in the number of threads or processes, etc.). Conceptually, it would be close to the firecracker-demo scenario.

kzys:
@ustiugov Good suggestion! I think launching 2 to 10 VMs would be easy to add in this PR.

kzys:
Here is the last result:

```
% make -C runtime benchmark-BenchmarkCreateAndStopVM
make: Entering directory '/home/katokazu/ec2-ws/firecracker-containerd/runtime'
/home/katokazu/ec2-ws/firecracker-containerd/runtime/../tools/thinpool.sh reset ""
docker run --rm -it \
        --privileged \
        --ipc=host \
        --network=none \
        --volume /dev:/dev \
        --volume /run/udev/control:/run/udev/control \
        --volume /home/katokazu/ec2-ws/firecracker-containerd/runtime/logs:/var/log/firecracker-containerd-test \
        --volume /home/katokazu/ec2-ws/firecracker-containerd/runtime/..:/src \
        --volume gocache:/go \
        --env ENABLE_ISOLATED_TESTS=1 \
        --env FICD_DM_VOLUME_GROUP= \
        --env FICD_DM_POOL= \
        --env GOPROXY=direct \
        --env GOSUMDB=off \
        --env NUMBER_OF_VMS= \
        --workdir="/src/runtime" \
        --init \
        localhost/firecracker-containerd-test:latest \
        "go test  -run IGNORE -bench 'BenchmarkCreateAndStopVM'"
goos: linux
goarch: amd64
pkg: github.com/firecracker-microvm/firecracker-containerd/runtime
BenchmarkCreateAndStopVM_Vcpu1_Isolated-96                     1        1571609700 ns/op
BenchmarkCreateAndStopVM_Vcpu5_Isolated-96                     1        1570367740 ns/op
BenchmarkCreateAndStopVM_Quiet_Isolated-96                     1        1130995588 ns/op
BenchmarkCreateAndStopVM_Parallel10_Isolated-96                1        1766212328 ns/op
PASS
ok      github.com/firecracker-microvm/firecracker-containerd/runtime   6.087s
make: Leaving directory '/home/katokazu/ec2-ws/firecracker-containerd/runtime'
```

ustiugov:
@kzys should we raise a separate issue on improving the VM creation speed?

kzys:
@ustiugov "Improving the VM creation speed" is open-ended. We can have an issue to discuss about possible improvements, but I personally prefer to have more concrete issues we can eventually close.

There are a few approaches that would make the VM creation faster;

1. Reduce logging (like BenchmarkCreateAndStopVM_Quiet_Isolated)
1. Check what's going in a micro VM (probably with [systemd-analyze](https://www.freedesktop.org/software/systemd/man/systemd-analyze.html))
1. Pass more parameters upfront to reduce API requests to a Firecracker socket
1. Firecracker snapshots (https://github.com/firecracker-microvm/firecracker/pull/1537)
1. Firecracker block IO improvement (https://github.com/firecracker-microvm/firecracker/issues/1600)

IRCody:
> @ustiugov "Improving the VM creation speed" is open-ended. We can have an issue as a place to discuss about possible improvements. Personally I prefer to have more concrete issues we can eventually close. @mxpv @xibz @IRCody What do you think?

I think I'd also prefer the concrete issues. I can see how it might be useful to have a place to see all the concrete issues though. Would it make sense to have a "Performance improvement" project board or something like that to link issues related to performance?

runtime/benchmark_test.go
```diff
@@ -0,0 +1,119 @@
...
+					ch <- err
+					return
+				}
+			}
+		}()
+	}
+
+	// Make sure all goroutines are finished successfully.
+	for i := 0; i < parallel; i++ {
+		require.NoError(<-ch)
```
IRCody: If there are errors we will write more than parallel times to ch I think? Worst case is every CreateAndStopVM() call errors which would be (b.N+1)*parallel iiuc. I'm not sure if that matters here as we are failing on *any* error though. Can you confirm?

ustiugov:
@kzys Thank you, I agree that setting concrete goals is essential. 
However, I would suggest collecting the CreateVM delay breakdown (e.g., gRPC/connection, FS mounting, VM boot, etc.) as the first Issue. system-analyze can be one of the steps, as mentioned above.

runtime/benchmark_test.go
```diff
@@ -26,7 +26,32 @@ import (
...
+	request proto.CreateVMRequest,
+) error {
+	uuid, err := uuid.NewV4()
+	if err != nil {
+		return err
+	}
+
+	request.VMID = uuid.String()
+
+	_, err = fcClient.CreateVM(ctx, &request)
```
ustiugov: I would suggest adding a deadline because I experience sporadic timeouts. Probably 3-4 seconds can be reasonable. 
```suggestion
clientDeadline := time.Now().Add(time.Duration(*deadlineMs) * time.Millisecond)
ctx, cancel := context.WithDeadline(ctx, clientDeadline)
```

kzys: Thanks. Will do.

runtime/benchmark_test.go
kzys: I think all goroutines will write either 1 (because of `defer`) or 2 (because of `ch <- err` and `defer`). Yes. That is wrong.

kzys: Addressed in the last revision.

runtime/benchmark_test.go
kzys: Addressed in the last revision.

Issue #387, opened 2020-02-07
taskworkflow dies with snapshotter error


ustiugov:
Dear Team,

I installed and launched firecracker-containerd following the quickstart guide. When I execute `examples/taskworkflow.go`, I face the following error:

```
2020/02/07 13:14:51.112672 Creating containerd client
2020/02/07 13:14:51.113477 Created containerd client
2020/02/07 13:14:52.203320 creating container: unpack: failed to stat snapshot sha256:531743b7098cb2aaf615641007a129173f63ed86ca32fe7b5a246a1c47286028: snapshotter not loaded: devmapper: invalid argument
```

At the same time, I don't see any containerd output when running taskworkflow. Manual commands from the quickstart guide work just fine.

Host: Ubuntu 18.04, kernel 4.15.0-76-generic

ustiugov:
I've located the problem. In taskworkflow.go, the sock path is configured incorrectly:
```
-       containerdAddress      = "/run/containerd/containerd.sock"
+       containerdAddress      = "/run/firecracker-containerd/containerd.sock"
```

Also, taskworkflow.md is out of date a bit. For example, `-mask` argument has been removed. 

Shall I submit a PR fixing those two bugs?

ustiugov:
I tried to test taskworkflow with network configuration but there is the following problem:

```
ustiugov@iccluster027:~/staged-data-tiering/firecracker-containerd/examples$ sudo ./taskworkflow -ip 172.16.0.2 -gw 172.16.0.1
2020/02/07 16:55:04.203970 Creating containerd client
2020/02/07 16:55:04.204690 Created containerd client
2020/02/07 16:55:05.309439 failed to create VM: rpc error: code = Unknown desc = shim CreateVM returned error: rpc error: code = Unknown desc = failed to create VM: failed to build VM configuration: failed to convert network config StaticConfig:<MacAddress:"AA:FC:00:00:00:01" HostDevName:"tap0" IPConfig:<PrimaryAddr:"172.16.0.2" GatewayAddr:"172.16.0.1" > > : failed to parse CIDR from "172.16.0.2": invalid CIDR address: 172.16.0.2
```
Configuration:

```
sudo ip tuntap add tap0 mode tap
sudo ip addr add 172.16.0.1/24 dev tap0
sudo ip link set tap0 up
sudo iptables -t nat -A POSTROUTING -o ens786f0 -j MASQUERADE
[...]

PR #399, opened 2020-03-02
Updated taskworkflow.go: fixes in socket name and readme.

ustiugov:
*Issue #387 *

*Description of changes:*

Socket name was aligned with the quickstart guidelines. Removed the `-mask` argument from the readme as it is not supported anymore. Specified that `-ip` argument requires CIDR and not just an IP address.

Signed-off-by: Dmitrii Ustiugov <dmitrii.ustiugov@ed.ac.uk>

### DIFF
--- a/examples/taskworkflow.go
+++ b/examples/taskworkflow.go
@@ -36,7 +36,7 @@ import (
 )
 
 const (
-	containerdAddress      = "/run/containerd/containerd.sock"
+	containerdAddress      = "/run/firecracker-containerd/containerd.sock"
 	containerdTTRPCAddress = containerdAddress + ".ttrpc"
 	namespaceName          = "firecracker-containerd-example"
 	macAddress             = "AA:FC:00:00:00:01"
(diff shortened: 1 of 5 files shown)


kzys:
I think our Docker image is still running containerd on `/run/containerd/containerd.sock` and that's why the test is failing.

ustiugov:
@kzys shall we change `tools/docker/config.toml` to contain:

```
imports ["/etc/containerd/snapshotter/*.toml"]
[grpc]
  address = "/run/firecracker-containerd/containerd.sock"

```

and then add --config option to `tools/docker/entrypoint.sh` 

would it be the right way?

kzys:
@ustiugov That's the right way, but you don't need to pass `--config` since `/etc/containerd/config.toml` is the default configuration file.

https://github.com/firecracker-microvm/firecracker-containerd/blob/f7734c546bec3764be0dece8cc741ef78b3b830c/tools/docker/Dockerfile.integ-test#L33

ustiugov:
@kzys feedback applied, tests pass. Could you please check?

PR #404, opened 2020-03-19
Use GitHub Actions

kzys:
We are pretty happy about Travis CI, but GitHub Actions are

- Available as hosted and self-hosted. We could use GitHub Actions to
  replace both Travis CI and BuildKite.
- Well integrated with GitHub. We can reduce accounts/credentials we
  need to worry about.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,35 @@
+name: Build
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    env:
+      GO111MODULE: on
+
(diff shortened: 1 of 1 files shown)


xibz:
Hmm, how would this work running on bare metal for AL2?

kzys:
@xibz Right now, GItHub Actions are replacing only Travis CI.

But GitHub Actions have something similar to BuildKite's agent (https://help.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners). So we could install that on EC2 in future.

xibz:
@kzys - Ah okay. The bullet point confused me and I thought you were wanting to replace both. Okay, this LGTM!

PR #406, opened 2020-03-21
Upgrade Firecracker Go SDK to v0.21.0

kzys:
*Issue #, if available:*

NA

*Description of changes:*

Upgrade Firecracker Go SDK to v0.21.0

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #407, opened 2020-03-23
Build all variants like Travis CI

kzys:
*Issue #, if available:*

NA

*Description of changes:*

GitHub Actions cancel all in-progress jobs if any matrix job fails.
This behavior makes investigating version-specific issues (e.g. #406)
harder.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -18,6 +18,8 @@ jobs:
     strategy:
       matrix:
         go: ['1.11', '1.12', '1.13', '1.14']
+      # Build all variants regardless of failures
+      fail-fast: false
     name: Go ${{ matrix.go }}
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -18,14 +18,23 @@ jobs:
...
+          # By default, the checkout action only fetches the last commit,
+          # but we want to run DCO check against all commit messages.
+          # Assuming that a pull request may have multiple commits,
+          # fetching the single commit doesn't work for us.
+          #
+          # Since the checkout action automatically creates a merge commit
+          # for merging "master" into a pull request branch, we need to
+          # have 21 commits (including the merge commit) to have
+          # 20 actual commits.
+          fetch-depth: 21
```
xibz: Can we also state that the 20 comes from the `make lint/git-validation` rule? Just to be more clear

kzys: Sure. Will do.

PR #412, opened 2020-05-03
Fixes bug with fifo files

xibz:
This commit addresses an issue with the fifo paths specified in the
machine config not being relative to the jailer's root path. This change
fixes this by making the path relative and running chown on those fifo
files

*Issue #, if available:*

*Description of changes:*


kzys:
Can you write some tests to prevent regression?

xibz:
@kzys - this is actually already tested. How this was discovered was our tests were failing when we updated the SDK

kzys:
Thanks. I didn't know that. LGTM.

PR #414, opened 2020-05-06
Remove a shim directory on StopVM() correctly and synchronously

kzys:
Previously the cleanup logic is in firecracker-control, but it doesn't
work correctly with a runc-based jailer, since the jailer is making
files without namespaces.

Morever the cleanup logic is running after StopVM() that makes our
contracts unclear. We should guarantee all resource are deleted at
the end of StopVM().

*Issue #, if available:*

#415

*Description of changes:*

### DIFF
--- a/runtime/jailer_integ_test.go
+++ b/runtime/jailer_integ_test.go
@@ -99,15 +99,13 @@ func testJailer(t *testing.T, jailerConfig *proto.JailerConfig) {
 func TestJailerCPUSet_Isolated(t *testing.T) {
 	prepareIntegTest(t)
 
-	t.Run("TestJailerCPUSet_Isolated", func(t *testing.T) {
-		b := cpuset.Builder{}
-		cset := b.AddCPU(0).AddMem(0).Build()
-		config := &proto.JailerConfig{
-			CPUs: cset.CPUs(),
-			Mems: cset.Mems(),
-			UID:  300000,
-			GID:  300000,
-		}
-		testJailer(t, config)
-	})
+	b := cpuset.Builder{}
+	cset := b.AddCPU(0).AddMem(0).Build()
+	config := &proto.JailerConfig{
+		CPUs: cset.CPUs(),
+		Mems: cset.Mems(),
+		UID:  300000,
+		GID:  300000,
+	}
+	testJailer(t, config)
 }
--- a/runtime/jailer_integ_test.go
+++ b/runtime/jailer_integ_test.go
@@ -15,12 +15,15 @@ package main
 
 import (
 	"context"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/namespaces"
 	"github.com/containerd/containerd/oci"
 	"github.com/containerd/containerd/pkg/ttrpcutil"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	_ "github.com/firecracker-microvm/firecracker-containerd/firecracker-control"
@@ -78,10 +81,19 @@ func testJailer(t *testing.T, jailerConfig *proto.JailerConfig) {
 	stdout := startAndWaitTask(ctx, t, c)
 	require.Equal("hello", stdout)
 
-	defer func() {
-		err := c.Delete(ctx, containerd.WithSnapshotCleanup)
-		require.NoError(err, "failed to delete a container")
-	}()
+	stat, err := os.Stat(filepath.Join(shimBaseDir(), "default", vmID))
+	require.NoError(err)
+	assert.True(t, stat.IsDir())
+
+	err = c.Delete(ctx, containerd.WithSnapshotCleanup)
+	require.NoError(err, "failed to delete a container")
+
+	_, err = fcClient.StopVM(ctx, &proto.StopVMRequest{VMID: vmID})
+	require.NoError(err)
+
+	_, err = os.Stat(filepath.Join(shimBaseDir(), "default", vmID))
+	assert.Error(t, err)
+	assert.True(t, os.IsNotExist(err))
 }
 
 func TestJailerCPUSet_Isolated(t *testing.T) {
--- a/runtime/runc_jailer.go
+++ b/runtime/runc_jailer.go
@@ -156,7 +156,7 @@ func (j *runcJailer) BuildJailedMachine(cfg *config.Config, machineConfig *firec
 func (j *runcJailer) BuildJailedRootHandler(cfg *config.Config, machineConfig *firecracker.Config, vmID string) firecracker.Handler {
 	ociBundlePath := j.OCIBundlePath()
 	rootPath := j.RootPath()
-	machineConfig.SocketPath = filepath.Join(rootPath, "api.socket")
+	machineConfig.SocketPath = filepath.Join(rootfsFolder, "api.socket")
 
 	return firecracker.Handler{
 		Name: jailerHandlerName,
--- a/runtime/runc_jailer.go
+++ b/runtime/runc_jailer.go
@@ -23,9 +23,11 @@ import (
 	"path/filepath"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/containerd/go-runc"
 	"github.com/firecracker-microvm/firecracker-go-sdk"
+	"github.com/hashicorp/go-multierror"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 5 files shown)


kzys:
"runc did not terminate sucessfully" is coming from https://github.com/containerd/go-runc/blob/7016d3ce2328dd2cb1192b2076ebd565c4e8df0c/runc.go#L179

runtime/service.go
```diff
@@ -1183,11 +1192,18 @@ func (s *service) shutdown(
...
 
-	err := s.machine.Wait(context.Background())
-	if err == nil {
-		return nil
+	var result *multierror.Error
+	if err := s.machine.Wait(context.Background()); err != nil {
+		result = multierror.Append(result, err)
+	}
+	if err := s.jailer.Close(); err != nil {
+		result = multierror.Append(result, err)
```
IRCody, line 1200: Is this racing with monitorVmExit()'s call to jailer.Close() as well as some of the actions in shutdownLoop() which can exit the shim while this is in process?

kzys: Good point. It might be. Let me double check.

kzys: shutdownLoop() should be fine. Since `jailer.Close()` is always called after `machine.Wait()`, we can gurantee the goroutine is not there at the time we are calling jailer.Close.

Regarding monitorVMExit(), there is a chance that two jailer.Close() in happening in the same time. I'm not so sure how bad it would be.

I do agree our shutdown logic is complicated.

kzys: How about putting jailer.Close() and publishVMStop() inside sync.Once and put the once inside service as like vmStartOnce?

IRCody: I think moving all the cleanup stuff to one func() and putting that inside a sync.Once could make sense. We'd need a way to also show when that function is done that it's safe to call shimCancel I think.

runtime/jailer_integ_test.go
```diff
@@ -78,24 +81,31 @@ func testJailer(t *testing.T, jailerConfig *proto.JailerConfig) {
...
+	require.NoError(err)
+	assert.True(t, stat.IsDir())
+
+	err = c.Delete(ctx, containerd.WithSnapshotCleanup)
+	require.NoError(err, "failed to delete a container")
+
+	_, err = fcClient.StopVM(ctx, &proto.StopVMRequest{VMID: vmID})
+	require.NoError(err)
+
+	_, err = os.Stat(filepath.Join(shimBaseDir(), "default", vmID))
```
IRCody, line 94: So this will leave the ShimBaseDir()/default dir?

What happens if we are using multiple namespaces? Do those shimBaseDir/Namespace dirs ever get cleaned up?

kzys: Yes. The namespace directory will be kept. One thing we could do for workaround the issue is creating the directory like `${namespace}-${vmid}`, not '${namespace}/${vmid}', but not so sure the value it will add, assuming all directories are empty anyway.

Interestingly containerd seems having the same problem.

```
% sudo ls /var/lib/containerd/io.containerd.runtime.v2.task
default
% sudo ctr -n ns1 run --tty --rm docker.io/library/debian:latest debian
root@ip-172-31-25-63:/# exit
% sudo ls /var/lib/containerd/io.containerd.runtime.v2.task            
default  ns1
% sudo ls /var/lib/containerd/io.containerd.runtime.v2.task/ns1 
%
```

The empty ns1 directory is kept, while there are no containers inside.

runtime/runc_jailer.go
```diff
@@ -460,9 +460,19 @@ func (j *runcJailer) setDefaultConfigValues(cfg *config.Config, socketPath strin
 // Close will cleanup the container that may be left behind if the jailing
 // process was killed via SIGKILL.
 func (j *runcJailer) Close() error {
-	return j.runcClient.Delete(j.ctx, j.vmID, &runc.DeleteOpts{
-		Force: true,
-	})
+	// Even the jailer's associated context is cancelled,
+	// we'd like to do the cleanups below just in case.
+	ctx := context.Background()
```
xibz, line 465: I think we probably still want some sort of deadline here. Can we use a timeout?

runtime/service.go
```diff
@@ -1304,26 +1316,35 @@ func (s *service) Cleanup(requestCtx context.Context) (*taskAPI.DeleteResponse,
...
 
-func (s *service) monitorVMExit() {
-	defer func() {
+// cleanup resources
+func (s *service) cleanup() error {
+	s.cleanupOnce.Do(func() {
 		// we ignore the error here due to cleanup will only succeed if the jailing
 		// process was killed via SIGKILL
 		if err := s.jailer.Close(); err != nil {
+			s.vmExitErr = err
```
xibz, line 1325: We potentially could be stomping over the error here. How do you feel about just using an error in `cleanup`'s scope?

runtime/service.go
```diff
@@ -484,6 +485,10 @@ func (s *service) publishVMStop() error {
 	return s.eventExchange.Publish(s.shimCtx, StopEventName, &proto.VMStop{VMID: s.vmID})
 }
 
+func jailerRootDir(shimBaseDir, namespace, vmID string) string {
```
xibz, line 488: I wonder if we should move this to `internal/vm` package similarly to all the other directory stuff.

kzys: I'd like to keep it here. Moving that to internal/vm means that we need to expose the function, which in unnecessary.

runtime/service.go
kzys: Let me check. This function might be the only one who sets vmExitErr.

runtime/runc_jailer.go
kzys: Okay. How about 5 seconds?

runtime/service.go
xibz: That's the point of internal. In Go, no one can use that package but the package itself.

kzys: Yeah. But moving there means anybody in the repo (e.g. agent) can use that. Do we want to?

xibz: Yea, I mean I don't see why not since it is in internal and it'll be with the rest of the directory stuff, but I don't feel too strongly on it.

runtime/runc_jailer.go
xibz: That sounds good to me

runtime/service.go
```diff
@@ -574,7 +584,6 @@ func (s *service) StopVM(requestCtx context.Context, request *proto.StopVMReques
 		timeout = time.Duration(request.TimeoutSeconds) * time.Second
 	}
 
-	defer s.shimCancel()
```
IRCody, line 577: Do we also want to remove the s.shimCancel() in the case of CreateVM errors and move that to a cleanup call?

runtime/service.go
```diff
@@ -1183,11 +1192,21 @@ func (s *service) shutdown(
 		s.shutdownLoop(requestCtx, timeout, req)
 	}()
 
-	err := s.machine.Wait(context.Background())
-	if err == nil {
+	var result *multierror.Error
+	if err := s.machine.Wait(context.Background()); err != nil {
+		result = multierror.Append(result, err)
+	}
+	if err := s.cleanup(); err != nil {
```
IRCody, line 1199: Is the jailer.Close() here racing with the Jailer.Stop() from ShutdownLoop? Is there an expected order for them to be called in?

runtime/service.go
kzys: Yes. We should reduce the number of places where we call shimCancel.

runtime/service.go
kzys: s.machine.Wait() will return once Firecracker is terminated. cleanup() won't be called while shutdownLoop() is running.

IRCody: Good call. I missed that.

runtime/service.go
kzys: Oh, there is ShimDir in the internal package which does exactly the same 🤦

runtime/service.go
kzys: This function is the only one who uses (read/write) the variable.

kzys: I've renamed the variable from vmExitErr to cleanupErr, since this is only used by cleanup. We still need to keep it in `*service` to make sure calling cleanup() always returns the same value.

runtime/service.go
```diff
@@ -1183,11 +1191,21 @@ func (s *service) shutdown(
...
+		result = multierror.Append(result, err)
+	}
+	if result == nil {
```
xibz, line 1201: Do you need this? Wont the `ErrorOrNil` handle it being nil?

https://github.com/hashicorp/go-multierror/blob/master/multierror.go#L29

kzys: That was my Java thinking... Let me fix.

runtime/runc_jailer.go
```diff
@@ -466,9 +467,20 @@ func (j *runcJailer) setDefaultConfigValues(cfg *config.Config, socketPath strin
...
-	})
+	// Even the jailer's associated context is cancelled,
+	// we'd like to do the cleanups below just in case.
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	_, stateErr := j.runcClient.State(ctx, j.vmID)
+	if stateErr == nil {
+		// Delete the container, if it is still running.
+		err := j.runcClient.Delete(ctx, j.vmID, &runc.DeleteOpts{Force: true})
```
xibz, line 478: I do have a little concern here about potentially leaking containers, since `State` can fail with any number of reasons. How do you feel about always trying to delete even if `stateErr` is `nil`

runtime/service_integ_test.go
```diff
@@ -580,35 +583,69 @@ func TestLongUnixSocketPath_Isolated(t *testing.T) {
...
-	})
-	require.NoError(t, err, "failed to create VM")
+	subtests := []struct {
+		name    string
+		request proto.CreateVMRequest
+	}{
+		{
+			name: "Without Jailer",
+			request: proto.CreateVMRequest{
+				VMID:              vmID + "1",
```
xibz, line 600: This is a little hacky. Im guessing you are doing this to see which ones which, but I'd prefer adding `-jailer` or something to the jailed one.

kzys: Thanks. Will do.

runtime/runc_jailer.go
kzys: I couldn't recall the reason of adding this check. Let me remove State() here and see how it works.

PR #418, opened 2020-05-08
Adding Cgroup customization via CreateVM

### LINKED ISSUE #417: Make cgroup path configurable from CreateVM

xibz:
This commit allows for configuration of the cgroup path. If no value is
provided the default value of /firecracker-containerd will be used
making the full cgroup path of
/sys/fs/cgroup/cpu/firecracker-containerd/<vmID>

Fixes #417 


proto/firecracker.proto
```diff
@@ -102,5 +102,11 @@ message JailerConfig {
     string Mems = 3;
     uint32 UID = 4;
     uint32 GID = 5;
+	// CgroupPath is used to dictate where the cgroup should be located
+	// relative to the cgroup directory which is
+	// /sys/fs/cgroup/cpu/<CgroupPath>/<vmID>
```
kzys, line 107: @ofiliz Now clients can inject an arbitrary string, but still cannot remove the last vmID part. Does it work?

runtime/jailer_test.go
```diff
@@ -96,3 +100,39 @@ func TestJailer_invalidUIDGID(t *testing.T) {
...
 	assert.Error(t, err, "expected invalid uid and gid error, but received none")
 }
+
+func TestNewJailer_Isolated(t *testing.T) {
+	prepareIntegTest(t)
+
+	ociBundlePath, err := ioutil.TempDir("", t.Name())
+	require.NoError(t, err)
+	defer os.RemoveAll(ociBundlePath)
+	err = exec.Command("cp", "firecracker-runc-config.json.example", filepath.Join(ociBundlePath, "config.json")).Run()
```
kzys, line 110: How about using CombinedOutput? If the cp doesn't work (e.g. disk is full), the err variable may not contain the information.

runtime/service_integ_test.go
```diff
@@ -470,7 +479,7 @@ func testMultipleExecs(
 		require.NoError(t, err, "failed to stat root path of jailer")
 		_, err = os.Stat(filepath.Join("/sys/fs/cgroup/cpu", cgroupPath))
 		require.NoError(t, err, "failed to stat cgroup path of jailer")
-		assert.Equal(t, filepath.Join("/firecracker-containerd", vmIDStr), cgroupPath)
+		assert.Regexp(t, fmt.Sprintf(".?.*/%s", vmIDStr), cgroupPath)
```
kzys, line 482: Would it be `fmt.Sprintf(".+/%s", vmIDStr)`?

xibz: Ah yea, that simplifies it. Forgot about the + meaning one or more.

runtime/jailer_test.go
xibz: That's fair. Will change.

proto/firecracker.proto
ofiliz: That's the same VMID in the CreateVMRequest, right? Yes, that should be fine.

PR #421, opened 2020-05-12
Fix Exec()'s timeout issue

kzys:

*Issue #, if available:*

NA

*Description of changes:*

* The first commit reproduces the issue
* The second commit fixes the issue. Opening a FIFO with O_RDWR is unnecessary and makes EOF invisible.
* The third commit fixes a timing issue. Closing all connectors by the cancellation of the passed context seems too early. We need to copy bytes even after the end of the process, since a pipe can buffer some bytes.


kzys:
cc @ddyzhang

PR #424, opened 2020-05-20
Adds debug helper

xibz:
This adds the debug package and debug.Helper which allows for easier
management of debug log levels. This also gets rid of the Debug field in
Config as it was really confusing on what logging that enabled, and also
firecracker-runtime.json. The log_level/LogLevel field in Config and the
firecracker-runtime.json has been renamed to log_levels and LogLevels.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/pkg/errors"
 
+	"github.com/firecracker-microvm/firecracker-containerd/internal/debug"
 	"github.com/firecracker-microvm/firecracker-containerd/proto"
 	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
 )
@@ -40,14 +41,13 @@ const (
 
 // Config represents runtime configuration parameters
 type Config struct {
-	FirecrackerBinaryPath string `json:"firecracker_binary_path"`
-	KernelImagePath       string `json:"kernel_image_path"`
-	KernelArgs            string `json:"kernel_args"`
-	RootDrive             string `json:"root_drive"`
-	CPUTemplate           string `json:"cpu_template"`
-	LogLevel              string `json:"log_level"`
-	HtEnabled             bool   `json:"ht_enabled"`
-	Debug                 bool   `json:"debug"`
+	FirecrackerBinaryPath string   `json:"firecracker_binary_path"`
+	KernelImagePath       string   `json:"kernel_image_path"`
+	KernelArgs            string   `json:"kernel_args"`
+	RootDrive             string   `json:"root_drive"`
+	CPUTemplate           string   `json:"cpu_template"`
+	LogLevels             []string `json:"log_levels"`
+	HtEnabled             bool     `json:"ht_enabled"`
 	// If a CreateVM call specifies no network interfaces and DefaultNetworkInterfaces is non-empty,
 	// the VM will default to using the network interfaces as specified here. This is especially
 	// useful when a CNI-based network interface is provided in DefaultNetworkInterfaces.
@@ -57,6 +57,8 @@ type Config struct {
 	// directory.
 	ShimBaseDir  string       `json:"shim_base_dir"`
 	JailerConfig JailerConfig `json:"jailer"`
+
+	DebugHelper *debug.Helper `json:"-"`
 }
 
 // JailerConfig houses a set of configurable values for jailing
@@ -90,6 +92,11 @@ func LoadConfig(path string) (*Config, error) {
 		JailerConfig: JailerConfig{
 			RuncConfigPath: runcConfigPath,
 		},
+	}
+
+	cfg.DebugHelper, err = debug.New(cfg.LogLevels...)
+	if err != nil {
+		return nil, err
 	}
 
 	if err := json.Unmarshal(data, cfg); err != nil {
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -1,0 +1,65 @@
+## Logging Congiguration
+--
+
+firecracker-containerd allows for users to specify per application/library
+logging. We do this by setting the `log_levels` field in the
+firecracker-runtime.json file.
+
+```json
+{
+  "firecracker_binary_path": "/usr/local/bin/firecracker",
+  "kernel_image_path": "/var/lib/firecracker-containerd/runtime/default-vmlinux.bin",
+  "kernel_args": "ro console=ttyS0 noapic reboot=k panic=1 pci=off nomodules systemd.journald.forward_to_console systemd.unit=firecracker.target init=/sbin/overlay-init",
+  "root_drive": "/var/lib/firecracker-containerd/runtime/default-rootfs.img",
+  "cpu_count": 1,
+  "cpu_template": "T2",
+  "log_levels": ["debug"],
+  "jailer": {
+    "runc_binary_path": "/usr/local/bin/runc"
+  }
+}
+```
+
+| log levels                     | description                                                                     |
+| :---------------------------:  | :-----------------------------------------------------------------------------: |
+| error                          | This will set all log levels to error                                           |
+| warning                        | This will set all log levels to warning                                         |
+| info                           | This will set all log levels to info                                            |
+| debug                          | This will set all log levels to debug                                           |
+| firecracker:error              | Logs any error information in Firecracker's lifecycle                           |
+| firecracker:warning            | Logs any error and warning information in Firecracker's lifecycle               |
+| firecracker:info               | Logs any error, warning, info information in Firecracker's lifecycle            |
+| firecracker:debug              | Most verbose log level for firecracker                                          |
+| firecracker:output             | Logs Firecracker's stdout and stderr. Can be used with other firecracker levels |
+| firecracker-go-sdk:error       | Logs any errors information in firecracker-go-sdk                               |
+| firecracker-go-sdk:warning     | Logs any errors or warnings in firracker-go-sdk                                 |
+| firecracker-go-sdk:info        | Logs any errors, warnings, or infos in firecracker-go-sdk                       |
+| firecracker-go-sdk:debug       | Most verbose logging for firecracker-go-sdk                                     |
+| firecracker-containerd:error   | Logs any error information during the container/vm lifecycle                    |
+| firecracker-containerd:warning | Logs any error level logs along with any warn level logs                        |
+| firecracker-containerd:info    | Logs any error, warn, and info level logs                                       |
+| firecracker-containerd:debug   | Most verbose logging for firecracker-containerd                                 |
+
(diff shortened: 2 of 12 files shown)


config/config.go
```diff
@@ -45,9 +46,8 @@ type Config struct {
 	KernelArgs            string `json:"kernel_args"`
 	RootDrive             string `json:"root_drive"`
 	CPUTemplate           string `json:"cpu_template"`
-	LogLevel              string `json:"log_level"`
+	LogLevels             string `json:"log_levels"`
```
kzys, line 49: How about using a list (`["debug", "firecracker_error"]`) or  a map (`{"firecracker":"error"}`)? It is odd to use CSV in JSON.

xibz: @kzys Yea, I'll use the list. Don't know why I defaulted to CSV :x

docs/logging.md
```diff
@@ -0,0 +1,52 @@
...
+  }
+}
+```
+
+| log levels                     | description                                                               |
+| :---------------------------:  | :-----------------------------------------------------------------------: |
+| debug                          | Enables default values for logging. This will set all log levels to debug |
+| firecracker_error              | Logs entries that come from Firecracker's error macro                     |
+| firecracker_warning            | Logs entries that come from Firecracker's error and warn macro            |
+| firecracker_info               | Logs entries that come from Firecracker's error, warn, and info macro     |
```
kzys, line 28: It is better to not mention `macro`. That's Firecracker's implementation details.

docs/logging.md
```diff
@@ -0,0 +1,52 @@
...
+| firecracker_debug              | Most verbose log level for firecracker                                    |
+| firecracker_output             | Logs Firecracker's stdout and stderr                                      |
+| firecracker-go-sdk             | Enables genrally logging for the firecracker-go-sdk                       |
+| firecracker-containerd_error   | Logs any error information during the container/vm lifecycle              |
+| firecracker-containerd_warning | Logs any error level logs along with any warn level logs                  |
+| firecracker-containerd_info    | Logs any error, warn, and info level logs                                 |
+| firecracker-containerd_debug   | Most verbose logging for firecracker-containerd                           |
+
+`log_levels` is a case insensitive list of comma seperated log levels.
```
kzys, line 37: Does it have to be case-insensitive? Most of stuff we have is case-sensitive.

internal/debug/debug.go
```diff
@@ -0,0 +1,206 @@
...
+	LogLevelFirecrackerInfo = "firecracker_info"
+	// LogLevelFirecrackerWarning enables warning level logging on Firecracker
+	// MicroVMs
+	LogLevelFirecrackerWarning = "firecracker_warning"
+	// LogLevelFirecrackerOutput will redirect stdout and stderr of Firecracker
+	// to logging. By default, stdout and stderr is ignored
+	LogLevelFirecrackerOutput = "firecracker_output"
+	// LogLevelFirecrackerSDK enables logging on firecracker-go-sdk. This will
+	// show what handlers and APIs are being called to Firecracker.
+	LogLevelFirecrackerSDK = "firecracker-go-sdk"
```
kzys, line 44: The SDK uses Logrus and have log levels like others. Why don't we support that?

xibz: Will add. Didn't realize that it also had log levels along with a debug boolean

xibz: Ah nvm, the log level is for firecracker and to enable logging on the sdk it is just a boolean.

docs/logging.md
xibz: It doesn't have to be, but again this was to support backwards compatibility, but since that is no longer an option I can make it case sensitive

docs/logging.md
xibz: Yea, I copied this straight from the docs of Firecracker, but I will go ahead and reword

internal/debug/debug.go
kzys: For example, startVMM() has Printf, Errorf and Debugf. Is there a way to log SDK's Errof on firecracker-containerd's logs?

https://github.com/firecracker-microvm/firecracker-go-sdk/blob/18c53a97c81711965327fe00349a7f99dcd6e5d5/machine.go#L482

xibz: Yea, we unfortunately don't expose a way of configuring that to the user D:

kzys: Hmm. It is complicated than we thought. The SDK has

1. Debug flag
1. LogLevel
1. The way to pass Logrus's logger

The first one is only used by [client_transport.go](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/a3f9b2f83cf947e6b3fc8da19a59c78b8e0c945e/client_transports.go#L44). [NewClient is taking the flag](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/a3f9b2f83cf947e6b3fc8da19a59c78b8e0c945e/firecracker.go#L65). For clients that use Machine abstraction. That is passed from [cfg.Debug](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/18c53a97c81711965327fe00349a7f99dcd6e5d5/machine.go#L338).

The second one is [Firecracker's level](https://github.com/firecracker-microvm/firecracker/blob/b3869fa29b500c08a021d161bb1cc669ec6a4ebf/src/api_server/swagger/firecracker.yaml#L661).

The third one is [used by our runtime](https://github.com/firecracker-microvm/firecracker-containerd/blob/1e7b50ed4b41575f88afe7c3200a5fe698b773ab/runtime/service.go#L515).

kzys: So setting firecracker-containerd_info affects the SDK since the runtime is passing the logger to SDK as is. I'm unsure what is the right way to fix that. Having a separate logger as we wrote in the doc?

xibz: Ah, yea, completely forgot you could pass the logger in yourself. Makes this way more complicated than it needs to be ><. Hm, I will go ahead and add the `firecracker-go-sdk_XX` levels and figure out a good way to pass the logger. I think I'll just have the `Helper` construct one and a getter to retrieve it.

I will spend a little time thinking about it and see if I can come up with something better

docs/logging.md
```diff
@@ -0,0 +1,52 @@
...
+  "log_levels": "debug",
+  "jailer": {
+    "runc_binary_path": "/usr/local/bin/runc"
+  }
+}
+```
+
+| log levels                     | description                                                               |
+| :---------------------------:  | :-----------------------------------------------------------------------: |
+| debug                          | Enables default values for logging. This will set all log levels to debug |
```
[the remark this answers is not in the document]
xibz: Wouldn't that be `firecracker-containerd_XX`?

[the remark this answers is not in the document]
IRCody: You can set all levels to debug, but why not warning/info/error also?

docs/logging.md
```diff
@@ -0,0 +1,56 @@
...
+  "jailer": {
+    "runc_binary_path": "/usr/local/bin/runc"
+  }
+}
+```
+
+| log levels                     | description                                                               |
+| :---------------------------:  | :-----------------------------------------------------------------------: |
+| debug                          | This will set all log levels to debug                                     |
+| firecracker_error              | Logs any error information in Firecracker's lifecycle                     |
```
kzys, line 26: How about using `=` or `:` as the separator, instead of `_`?

xibz: Yea, I thinking about using `:` but ultimately went just with `_`. But since you mentioned it, I'll go ahead and switch it to that.

runtime/service.go
```diff
@@ -183,9 +181,10 @@ func NewService(shimCtx context.Context, id string, remotePublisher shim.Publish
...
 	}
 
-	if cfg.Debug {
-		logrus.SetLevel(logrus.DebugLevel)
-		logger.Logger.SetLevel(logrus.DebugLevel)
+	logrusLevel, ok := cfg.DebugHelper.GetFirecrackerContainerdLogLevel()
+	if ok {
+		logrus.SetLevel(logrusLevel)
+		logger.Logger.SetLevel(logrusLevel)
 	}
```
kzys, line 188: How about returning the default log level, rather than using `ok`? Unlike SDK, clients cannot pass a Logrus logger to the shim. So we don't have to honor the existing log level.

xibz: Well, users may want to disable logging which that would prevent

runtime/service.go
```diff
@@ -503,10 +502,14 @@ func (s *service) createVM(requestCtx context.Context, request *proto.CreateVMRe
...
 
-	opts := []firecracker.Opt{
-		firecracker.WithLogger(s.logger),
-	}
+	opts := []firecracker.Opt{}
 
+	if v, ok := s.config.DebugHelper.GetFirecrackerSDKLogLevel(); ok {
+		logger := log.G(s.shimCtx)
+		logrus.SetLevel(v)
```
kzys, line 509: Since we allow SDK and firecracker-containerd to have different log levels, we shouldn't call logrus.SetLevel() to modify the global logger. The fact SDK's log level impacts the global logger and overrides firecracker-containerd's log level regarding the global logger is confusing.

xibz: Ah yea, that was copy pasta

internal/debug/debug.go
```diff
@@ -92,7 +110,7 @@ type Helper struct {
 // New will return a new Helper in the event an error does not occur. This will
 // parse the logLevel provided to figure out what logging is enabled. New will
 // also validate that the log level is a valid value along with any mutually
-// exclusive values.
+// ekclusive values.
```
IRCody, line 113: exclusive*

docs/logging.md
```diff
@@ -0,0 +1,65 @@
+## Logging Congiguration
+--
+
+firecracker-containerd allows for users to specify per application/library
+logging. We do this by setting the `log_level` field in the
```
kzys, line 5: log_levels

docs/logging.md
```diff
@@ -0,0 +1,65 @@
...
+
+```json
+{
+  "firecracker_binary_path": "/usr/local/bin/firecracker",
+  "kernel_image_path": "/var/lib/firecracker-containerd/runtime/default-vmlinux.bin",
+  "kernel_args": "ro console=ttyS0 noapic reboot=k panic=1 pci=off nomodules systemd.journald.forward_to_console systemd.unit=firecracker.target init=/sbin/overlay-init",
+  "root_drive": "/var/lib/firecracker-containerd/runtime/default-rootfs.img",
+  "cpu_count": 1,
+  "cpu_template": "T2",
+  "log_levels": "debug",
```
kzys, line 16: `["debug"]`

docs/logging.md
xibz: Fixed!

docs/logging.md
xibz: Same here, fixed.

Issue #423, opened 2020-05-20
Make CreateVM's timeout configurable


kzys:
Currently CreateVM doesn't have timeout (unlike StopVM) and we have some hard-coded timeouts around CreateVM.

These timeouts should be configurable to accommodate other rootfs images.

kzys:
I'm going to add TimeoutSeconds on CreateVMRequest below.

https://github.com/firecracker-microvm/firecracker-containerd/blob/ca88e0fd3ef941b2632b474f4e9adf722f2e13e0/proto/firecracker.proto#L8

And remove the timeout below, since CreateVM's timeout can be used to cancel the context.

https://github.com/firecracker-microvm/firecracker-containerd/blob/1e7b50ed4b41575f88afe7c3200a5fe698b773ab/internal/vm/vsock.go#L41

@aaithal @IRCody @xibz What do you folks think?

aaithal:
This approach makes sense to me. I'm not sure if this is required, but I had to bump this one as well: https://github.com/firecracker-microvm/firecracker-containerd/blob/1e7b50ed4b41575f88afe7c3200a5fe698b773ab/runtime/service.go#L73

PR #425, opened 2020-05-20
Make CreateVM's timeout configurable

kzys:
*Issue #, if available:* #423

*Description of changes:*

As like StopVM, CreateVM's timeout should be configurable to accomodate
other root fs images and slower environments.


runtime/service.go
```diff
@@ -68,12 +68,18 @@ func init() {
...
+
+	// vmReadyTimeout is used to control the time all requests wait a Go channel (vmReady) before calling
+	// Firecracker's API server. The channel is closed once the VM starts.
+	vmReadyTimeout = 5 * time.Second
+
+	defaultCreateVMTimeout = 20 * time.Second
+	defaultStopVMTimeout   = 5 * time.Second
+	defaultShutdownTimeout = 5 * time.Second
+
+	jailerStopTimeout = 3 * time.Second
```
IRCody, line 82: Does it make sense to make the rest of these configurable also?

runtime/service.go
```diff
@@ -435,14 +441,21 @@ func (s *service) waitVMReady() error {
 func (s *service) CreateVM(requestCtx context.Context, request *proto.CreateVMRequest) (*proto.CreateVMResponse, error) {
 	defer logPanicAndDie(s.logger)
 
+	timeout := defaultCreateVMTimeout
+	if request.TimeoutSeconds > 0 {
+		timeout = time.Duration(request.TimeoutSeconds) * time.Second
```
IRCody, line 446: Is there a sensible maximum we'd want to enforce for TimeoutSeconds?

runtime/service.go
kzys: - StopVM's timeout is configurable and defaultStopVMTimeout is the default value. We don't have to make it configurable.
- Shutdown() is only exposed to containerd. We cannot have a request-level timeout. We could have a configuration variable on firecracker-runtime.json, but I'd like to keep it as is, to keep our configuration simpleler.
- jailerStopTimeout is not doing compute-intensive stuff, compared to CreateVM. 3 seconds may be enough for everyone.

So, right now I don't think we need to expose these knobs to clients.

runtime/service.go
kzys: I assume container clients are containerd operators. People won't and shouldn't expose containerd to the internet. Because of that, if clients want to wait 30 minutes or even one hour, it is their choice.

PR #433, opened 2020-06-05
Populate response for CreateVM in Runtime shim

olibearo:
*Issue #, if available:*

N/A

*Description of changes:*

Populate the response fields in the Runtime Shim's CreateVM method and refactor redundant calls to determine the shim directory and VM namespace. 

### DIFF
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -336,7 +336,6 @@ func (s *service) StartShim(shimCtx context.Context, containerID, containerdBina
 		if err != nil {
 			return "", errors.Wrap(err, "failed to generate UUID for VMID")
 		}
-
 		s.vmID = uuid.String()
 
 		// This request is handled by a short-lived shim process to find its control socket.
@@ -457,7 +456,6 @@ func (s *service) CreateVM(requestCtx context.Context, request *proto.CreateVMRe
 		err = s.createVM(ctxWithTimeout, request)
 		createRan = true
 	})
-
 	if !createRan {
 		return nil, status.Error(codes.AlreadyExists, "shim cannot create VM more than once")
 	}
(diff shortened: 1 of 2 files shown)


olibearo:
@kzys I've tested this locally with a VM with a simple VM that has a predefined id and the VM metadata is populating correctly on the response. Thanks for taking a look at this

kzys:
Thanks @olibearo! Can you sign-off your commits?

https://github.com/firecracker-microvm/firecracker-containerd/blob/master/CONTRIBUTING.md#contributing-via-pull-requests

runtime/service.go
```diff
@@ -454,10 +456,19 @@ func (s *service) CreateVM(requestCtx context.Context, request *proto.CreateVMRe
 	)
 
 	s.vmStartOnce.Do(func() {
-		err = s.createVM(ctxWithTimeout, request)
+		if err = s.createVM(ctxWithTimeout, request); err != nil {
```
kzys, line 459: Could you add some tests to validate the response?

olibearo: I did a small refactor to make the construction of the response a separate function in order to make the test easier, but then realized that constructing the response is just copying values over from the `service` struct. 

What would you like to see asserted?

olibearo:
@kzys I squashed and adding the self signature

runtime/service.go
kzys: How about checking VMID in https://github.com/firecracker-microvm/firecracker-containerd/blob/master/runtime/service_integ_test.go#L1792?

olibearo: @kzys Done. Getting CI green, will ping when done

PR #442, opened 2020-07-28
Make Firecracker's instance ID configurable

kzys:
*Issue #, if available:*

NA

*Description of changes:*

Firecracker's instance ID is appeared in Firecracker's log
file. Setting the ID makes the log easily trackable from clients.

### DIFF
--- a/runtime/integ_test.go
+++ b/runtime/integ_test.go
@@ -93,7 +93,7 @@ func writeRuntimeConfig(options ...func(*config.Config)) error {
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


kzys:
The corresponding change I made for Go SDK is not required 🤦 

https://github.com/firecracker-microvm/firecracker-go-sdk/pull/253

xibz:
> The corresponding change I made for Go SDK is not required 
It doesn't hurt to make it configurable in the config though

PR #443, opened 2020-08-11
Make runc_jailer's behavior configurable regarding drive files

kzys:

*Issue #, if available:*

*Description of changes:*

runc_jailer implicitly copies all drive files to a jail directory before calling runc.

While copying provides a stronger guarantee of isolation, some clients would want have more control on the drive files (e.g. use shred(1) for removing files).

DriveExposePolicy is designed to use bind-mount instead of copying.


proto/firecracker.proto
```diff
@@ -110,5 +110,11 @@ message JailerConfig {
 	// if no value was provided, then /firecracker-containerd will be used as
 	// the default value
     string CgroupPath = 6;
+
+    // DriveExposePolicy is used to configure the method to expose drive files.
+    // "copy" is copying the files to the jail, which is the default behavior.
+    // "bind" is bind-mounting the files on the jail, assuming a caller pre-configures the permissions of
+    // the files appropriately.
+    string DriveExposePolicy = 7;
```
IRCody, line 118: Would this make sense as an enum since we don't allow arbitrary strings? Maybe something like:

```proto
enum DrivePolicy {
    COPY = 1;
    BIND = 2;
}
DrivePolicy DriveExposePolicy = 7 [default = BIND];
```

kzys: I didn't know proto3 can have enums (https://developers.google.com/protocol-buffers/docs/proto3#enum). Yes it does make sense.

runtime/runc_jailer.go
```diff
@@ -63,8 +63,14 @@ type runcJailerConfig struct {
 	CPUs           string
 	Mems           string
 	CgroupPath     string
+
+	// DriveExposePolicy defines how the jailer exposes files.
+	DriveExposePolicy string
 }
 
+const driveExposePolicyCopy = "copy"
+const driveExposePolicyBind = "bind"
```
IRCody, line 72: If we can use iotas here and match to enum in the proto we won't need to do the string validation. Would that make sense here?

kzys: Yes. Removed on the last revision.

proto/firecracker.proto
kzys: The default syntax might be proto2, whereas proto3's default value is always zero.

PR #450, opened 2020-10-06
Make LogPath and MetricsPath configurable

kzys:
Before this change, firecracker-containerd was returning the paths of
the files as a part of CreateVM's response. However the files are also
useful to investage when CreateVM doesn't finish successfully.

In addition to that, "FIFO" is removed from a few fields, since
Firecracker renamed the fields.

*Issue #, if available:*

*Description of changes:*


kzys:
Let me replace the AL2 instance.

kzys:
Synced-up offline with @IRCody. I'm going to make this PR backward-compatible. We will discuss the policy around backward compatibility in general, but not in this PR.

IRCody (approved):

PR #453, opened 2020-10-24
Rename the files inside a jail to avoid conflicts

kzys:
Previously the basename part of a customer-supplied path is used
inside a jail directory.

However there is no guarantee that the supplied-path doesn't cause
a conflict against other paths (e.g. LogFifo vs. MetricsFifo) and/or
important system-reserved files (e.g. "kernel-image").

It is better to rename the files by ourselves rather than relying on
the customer-supplied paths.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/runtime/runc_jailer.go
+++ b/runtime/runc_jailer.go
@@ -270,10 +270,14 @@ func (j *runcJailer) BuildJailedRootHandler(cfg *config.Config, machineConfig *f
 	}
 }
 
-func (j *runcJailer) makeLinkInJail(src string) (string, error) {
+// makeLinkInJail creates a hard link to `src` inside the jail directory.
+func (j *runcJailer) makeLinkInJail(src, base string) (string, error) {
 	root := j.RootPath()
 
-	base := filepath.Base(src)
+	if strings.ContainsRune(base, os.PathSeparator) {
+		return "", fmt.Errorf("%q must not contain %q", base, os.PathSeparator)
+	}
+
 	dst := filepath.Join(root, base)
 
 	// Since Firecracker is unaware that we are in a jailed environment and
@@ -294,13 +298,13 @@ func (j *runcJailer) BuildLinkFifoHandler() firecracker.Handler {
 	return firecracker.Handler{
 		Name: jailerFifoHandlerName,
 		Fn: func(ctx context.Context, m *firecracker.Machine) error {
-			logFifo, err := j.makeLinkInJail(m.Cfg.LogPath)
+			logFifo, err := j.makeLinkInJail(m.Cfg.LogPath, internal.FirecrackerLogFifoName)
 			if err != nil {
 				return err
 			}
 			m.Cfg.LogFifo = logFifo
 
-			metricsFifo, err := j.makeLinkInJail(m.Cfg.MetricsPath)
+			metricsFifo, err := j.makeLinkInJail(m.Cfg.MetricsPath, internal.FirecrackerMetricsFifoName)
 			if err != nil {
 				return err
 			}
--- a/runtime/runc_jailer_test.go
+++ b/runtime/runc_jailer_test.go
@@ -171,3 +171,66 @@ func TestBindMountToJail_Isolated(t *testing.T) {
 	)
 	require.Error(t, err)
 }
+
+func TestFifoHandler_Isolated(t *testing.T) {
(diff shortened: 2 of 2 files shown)


runtime/runc_jailer.go
```diff
@@ -270,10 +270,10 @@ func (j *runcJailer) BuildJailedRootHandler(cfg *config.Config, machineConfig *f
 	}
 }
 
-func (j *runcJailer) makeLinkInJail(src string) (string, error) {
+// makeLinkInJail creates a symlink to `src` inside the jail directory.
+func (j *runcJailer) makeLinkInJail(src, base string) (string, error) {
 	root := j.RootPath()
```
IRCody, line 276: Is it worth leaving the filepath.Base call but doing it on base instead of src to ensure we don't accidentally provide a path instead in the future here?

runtime/runc_jailer.go
```diff
@@ -294,13 +294,13 @@ func (j *runcJailer) BuildLinkFifoHandler() firecracker.Handler {
 	return firecracker.Handler{
 		Name: jailerFifoHandlerName,
 		Fn: func(ctx context.Context, m *firecracker.Machine) error {
-			logFifo, err := j.makeLinkInJail(m.Cfg.LogPath)
+			logFifo, err := j.makeLinkInJail(m.Cfg.LogPath, internal.FirecrackerLogFifoName)
```
IRCody, line 297: I want to phrase this change to ensure I'm understanding it correctly:

 What's happening here is that previously we were accepting a LogPath, and using the customer provided filename and now we will be using our own filename to avoid customers accidentally using a name that conflicts with other important names? 

Under what circumstances would this change be visible to current users?

runtime/runc_jailer.go
kzys: Good point. How about returning an error when `filepath.Base(base) != base`? I agree that `base` shouldn't be a path here. But correcting that automatically may hide the issue rather than let developers know and fix that.

runtime/runc_jailer.go
kzys: I think the change won't be visible to users, except for the one who are affected by the issue and workaround that by providing "safe" paths that don't conflict.

xibz, line 297: Does this mean `LogPath` will always be a directory and not a path to a file? Cause if that is the intention I dont know if I like the change.

xibz:
Actually the more I think about this the less I think it is needed. Like why not just allow the customer to provide their own path? Customers will know what's on their environment so I think it makes sense that they can provide a given path. Also using our internal log fifo names doesn't guarantee no collisions.

Did we actually run into an instance where a customer was actually overwriting files? A better solution is to create the file and fail if it already exists. We can do that in an atomic operation and we can just return an error.

runtime/runc_jailer.go
kzys: No. LogPath is still the path to Firecracker's log file. Previously the basename part of LogPath is used, but it restricts the value LogPath can take. For example, using `/var/log/fc/log/[VM ID].log` as LogPath while using `/var/log/fc/metrics/[VM ID].log` doesn't work since both of them become `[VM ID].log` inside a jail directory.

kzys:
Inside a jail directory, we have the following files;

```
[OCIBundlePath]/
  config.json
  rootfs/
    kernel-image
    firecracker
    [drive image 1]
    [drive image 2]
    ...
    [drive image N]
    [log fifo]
    [metrics fifo]
```

While we can document the restrictions such as "don't use `kernel-image`", the names of the files don't matter much for customers as long as they are correctly copied and/or symlinked from the original files. I think renaming all of them to avoid conflicts makes customers' life easier.

xibz:
Ahhhh, okay, that wasn't clear to me. I get what you are saying now. It wasn't clear to me that the mertics and log fifo could potentially collide with one another. With that being said I agree, just renaming makes it a lot easier than trying to do something like recreate the directory structure to prevent collisions. LGTM!

xibz (approved):

runtime/runc_jailer.go
kzys: Added `strings.ContainsRune(base, os.PathSeparator)` as a check.

IRCody (approved):

PR #459, opened 2020-11-17
Override Process.Args regardless of the existing value

kzys:
Previously firecracker-containerd keeps
/etc/containerd/firecracker-runc-config.json's Process.Args as is
if the value exists.

However passing VM ID (39e0475) doesn't work when it uses the exising
fixed value.

This change removes this confusing behavior and always overrides
Process.Args.

*Issue #, if available:*

*Description of changes:*


xibz (approved):
LGTM! I think one minor comment of documenting that the process args in the runc config will always be stomped over

kzys:
Thanks! I will have a follow-up PR regarding the documentation.

PR #460, opened 2020-11-20
Add PauseVM and ResumeVM support

plamenmpetrov:
The goal of this PR is to implement support for PauseVM and ResumeVM.

Authored by @plamenmpetrov and @ustiugov 

We are open to feedback and would be glad to finalize and contribute this code to upstream.

*Issue #, if available:*
#448 
*Description of changes:*
* Update firecracker-go-sdk to a version which supports PauseVM and ResumeVM.
* Update build procedure for `tc-redirect-tap`, which is no longer part of the Go SDK
* Implement calls to PauseVM and ResumeVM


kzys:
Looks good to me! Can you add an integration test that pauses and resumes a VM?

ustiugov:
it seems that the buildkite failure is not related to the patch. how can I re-run the test?

kzys:
Let me take a look. Right now only maintainers can restart tests.

runtime/service_integ_test.go
```diff
@@ -2032,6 +2032,93 @@ func TestCreateVM_Isolated(t *testing.T) {
 	}
 }
 
+func TestPauseResumeVM_Isolated(t *testing.T) {
```
kzys, line 2035: I'd like to have some other cases here.

- If we pause a VM multiple times, does it work? Does Firecracker itself returns an error that callers can understand? Should firecracker-containerd handle the case specifically by having some flags?
- If we resume the VM without pausing, does it work? The same questions above would be applied here as well.

runtime/cni_integ_test.go
```diff
@@ -24,20 +24,19 @@ import (
...
-	"github.com/stretchr/testify/require"
-
 	"github.com/firecracker-microvm/firecracker-containerd/config"
 	"github.com/firecracker-microvm/firecracker-containerd/internal"
 	"github.com/firecracker-microvm/firecracker-containerd/proto"
 	fccontrol "github.com/firecracker-microvm/firecracker-containerd/proto/service/fccontrol/ttrpc"
 	"github.com/firecracker-microvm/firecracker-containerd/runtime/firecrackeroci"
+	"github.com/shirou/gopsutil/cpu"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
kzys, line 39: Generally speaking, it is better to not make any style changes if you don't have to.

kzys (changes requested):

runtime/cni_integ_test.go
ustiugov: this change was automatic by GoFmt. However, I reverted it now.

runtime/service_integ_test.go
ustiugov: sorry for the delay, we will try to add the tests shortly

ustiugov:
@kzys we added the necessary tests, could you please take another look?

kzys:
Looks good to me. I think the build errors are related to #475, not your changes.

kzys:
I've merged #475. Can you rebase this PR against master? This should fix GitHub Actions failures at least.

kzys (approved):

PR #462, opened 2020-11-30
Upgrade containerd to 1.3.9

kzys:
- The version has a CVE fix, while we may not be affected.
- It is better to use relesed versions if possible.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/firecracker-control/local.go
+++ b/firecracker-control/local.go
@@ -20,7 +20,6 @@ import (
 	"os"
 	"os/exec"
 	"strconv"
-	"strings"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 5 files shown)


firecracker-control/local.go
```diff
@@ -223,14 +223,14 @@ func (s *local) shimFirecrackerClient(requestCtx context.Context, vmID string) (
...
-	socketAddr, err := fcShim.FCControlSocketAddress(requestCtx, vmID)
+	socketAddr, err := fcShim.FCControlSocketAddress(requestCtx, s.containerdAddress, vmID)
 	if err != nil {
 		err = errors.Wrap(err, "failed to get shim's fccontrol socket address")
 		s.logger.WithError(err).Error()
 		return nil, err
 	}
 
-	return fcclient.New("\x00" + socketAddr)
+	return fcclient.New(socketAddr) //
```
samuelkarp, line 233: Looks like a stray `//` got left in.

samuelkarp (approved):
LGTM! Just to confirm: CVE-2020-15257 does not affect firecracker-containerd as the abstract sockets exist outside the microVM guest; a process inside the microVM guest regardless of what namespaces and privileges it has cannot access an abstract socket on the host.

runtime/service.go
```diff
@@ -354,6 +355,15 @@ func (s *service) StartShim(shimCtx context.Context, containerID, containerdBina
...
 	}
 
+	// If there is a live socket, reuse the socket without calling CreateVM.
+	address, err := shim.SocketAddress(shimCtx, containerdAddress, s.vmID)
+	if err != nil {
+		return "", err
+	}
+	if shim.CanConnect(address) {
+		return address, nil
+	}
```
kzys, line 365: I would double-check this part. While this code itself makes sense, I don't know why it *was* working when the sockets were abstract.

kzys: This change is not needed. So I've removed from the last revision.

kzys:
We may skip this version and use 1.4.x. See #467.

PR #464, opened 2020-12-03
Adding default vsock connection timeout

xibz:
Prior to this commit, io would be proxied from the VM to the host, but
no timeout is provided to the proxy. So if vsock has an issue, the io
would forever try to connect to the vsock leading to an extraneous
amount of errors. This commit adds a simple default timeout, 5 second,
which will eliminate the extraneous amount of logs

*Issue #, if available:*

*Description of changes:*


IRCody (commented):
This LGTM. Do we know why AL2 is failing?

IRCody (approved):
Test failure looks unrelated.

PR #465, opened 2020-12-11
Fix cio.NewAttach

kzys:
cio.NewAttach didn't work with with firecracker-containerd. The operation is doing the following;

- Calls State() to find FIFO files that are connected to the container
- Passes the FIFO files to the client
- The client can open the FIFO files to interact with the containerd's stdio streams.

However firecracker-containerd was proxying the request to in-VM agent as is. As a result the FIFO files on the response are pointing the files inside the VM, which the client couldn't access.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -32,6 +32,7 @@ import (
 	// secure randomness
 	"math/rand" // #nosec
 
+	"github.com/containerd/containerd/cio"
 	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/events/exchange"
 	"github.com/containerd/containerd/log"
@@ -87,6 +88,10 @@ const (
 
 	// StopEventName is the topic published to when a VM stops
 	StopEventName = "/firecracker-vm/stop"
+
+	// taskExecID is a special exec ID that is pointing its task itself.
+	// While the constant is defined here, the convention is coming from containerd.
+	taskExecID = ""
 )
 
 var (
@@ -145,6 +150,10 @@ type service struct {
 	machineConfig    *firecracker.Config
 	vsockIOPortCount uint32
 	vsockPortMu      sync.Mutex
+
+	// fifos have stdio FIFOs containerd passed to the shim. The key is [taskID][execID].
+	fifos   map[string]map[string]cio.Config
+	fifosMu sync.Mutex
 }
 
 func shimOpts(shimCtx context.Context) (*shim.Opts, error) {
@@ -209,6 +218,7 @@ func NewService(shimCtx context.Context, id string, remotePublisher shim.Publish
 
 		vmReady: make(chan struct{}),
 		jailer:  newNoopJailer(shimCtx, logger, shimDir),
+		fifos:   make(map[string]map[string]cio.Config),
 	}
 
 	s.startEventForwarders(remotePublisher)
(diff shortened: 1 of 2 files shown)


runtime/service.go
```diff
@@ -884,6 +891,39 @@ func (s *service) newIOProxy(logger *logrus.Entry, stdin, stdout, stderr string,
...
+	defer s.fifosMu.Unlock()
+
+	_, exists := s.fifos[taskID]
+	if !exists {
+		s.fifos[taskID] = make(map[string]cio.Config)
+	}
+
+	value, exists := s.fifos[taskID][execID]
+	if exists {
+		return fmt.Errorf("(task=%q, exec=%q) must not have FIFOs, but there was %+v", taskID, execID, value)
```
IRCody, line 905: I think this could be reworded to be more clear. Why must we not have FIFOs already set here? (because this would overwrite them?)

runtime/service.go
```diff
@@ -884,6 +891,39 @@ func (s *service) newIOProxy(logger *logrus.Entry, stdin, stdout, stderr string,
...
+	return nil
+}
+
+func (s *service) deleteFIFOs(taskID, execID string) error {
+	s.fifosMu.Lock()
+	defer s.fifosMu.Unlock()
+
+	_, exists := s.fifos[taskID][execID]
+	if !exists {
+		return fmt.Errorf("(task=%q, exec=%q) must have FIFOs", taskID, execID)
```
IRCody, line 917: Could this be changed to something like "unable to delete fifo: Fifo does not exist"?

runtime/service.go
```diff
@@ -1088,6 +1152,20 @@ func (s *service) State(requestCtx context.Context, req *taskAPI.StateRequest) (
 		return nil, err
 	}
 
+	// If the fields aren't "", they are pointing in-VM FIFO files.
+	// Replacing them with the corresponding FIFO files on the host.
```
IRCody, line 1156: Is there ever a case where the fields are empty but we still want to return something here?

runtime/service.go
kzys: Sounds good. Will do.

runtime/service.go
kzys: Let me add test cases around the check. IIRC, /dev/null is converted to "".

runtime/service.go
kzys: > Why must we not have FIFOs already set here? (because this would overwrite them?)

Yeah. I don't want to overwrite the FIFO files.

runtime/service.go
kzys: `cio.NullIO` sets them `""`, but in that case, s.fifos will be also `""`. So I don't need the ifs.

runtime/service_integ_test.go
```diff
@@ -2012,3 +2012,97 @@ func TestCreateVM_Isolated(t *testing.T) {
...
+		{
+			name: "null io",
+
+			// firecracker-containerd doesn't create IO Proxy objects in this case.
+			newIO: func(ctx context.Context, id string) (cio.IO, error) {
+				return cio.NullIO(id)
+			},
+
+			// So, it doesn't work.
+			expected: "",
```
kzys, line 2054: I'm unsure this is the right behavior. Let me check that on contained's normal shim.

kzys: Looks like containerd's v2 shim is behaving the same. Let me update the comment.

runtime/service.go
kzys: I'm going to update this comment.

xibz (changes requested):
Thanks for the updated description. I had a couple comments.

runtime/service.go
```diff
@@ -87,6 +88,8 @@ const (
 
 	// StopEventName is the topic published to when a VM stops
 	StopEventName = "/firecracker-vm/stop"
+
+	taskExecID = ""
```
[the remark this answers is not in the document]
kzys: containerd's task by default has one exec with ID = `""`. I will document that on the comment.

runtime/service.go
```diff
@@ -145,6 +148,10 @@ type service struct {
 	machineConfig    *firecracker.Config
 	vsockIOPortCount uint32
 	vsockPortMu      sync.Mutex
+
+	// fifos have stdio FIFOs containerd passed to the shim. The key is [taskID][execID].
+	fifos   map[string]map[string]cio.Config
```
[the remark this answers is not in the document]
kzys: Since the service struct is in the shim binary and only handling one container, I'm unsure we have a lock contention here.

https://golang.org/pkg/sync/#Map

>  The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content. 

I'd like to keep this as is. We can benchmark later to see whether there is a bottleneck or not.

IRCody (approved):
LGTM minus this minor comment.

runtime/service.go
```diff
@@ -884,6 +892,39 @@ func (s *service) newIOProxy(logger *logrus.Entry, stdin, stdout, stderr string,
...
+	return nil
+}
+
+func (s *service) deleteFIFOs(taskID, execID string) error {
+	s.fifosMu.Lock()
+	defer s.fifosMu.Unlock()
+
+	_, exists := s.fifos[taskID][execID]
+	if !exists {
+		return fmt.Errorf("unable to delete the FIFO files for task %q (exec=%q)", taskID, execID)
```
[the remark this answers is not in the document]
kzys: How about `task %q (exec=%q) doesn't have corresponding FIFOs to delete`?

runtime/service.go
```diff
@@ -1088,6 +1155,14 @@ func (s *service) State(requestCtx context.Context, req *taskAPI.StateRequest) (
 		return nil, err
 	}
 
+	// These fields are pointing files inside the VM.
+	// Replace them with the corresponding files on the host, so clients can access.
+	s.fifosMu.Lock()
+	defer s.fifosMu.Unlock()
+	resp.Stdin = s.fifos[req.ID][req.ExecID].Stdin
```
xibz, line 1162: Is it possible that the `req.ID` or the `req.ExecID` to not be present?

kzys: Since this code is running after `s.agentClient.State(requestCtx, req)`, these IDs must be valid and there. If either of them is invalid, this function returns before reaching there.

xibz: Okay, just double checking!

xibz (approved):
LGTM

PR #466, opened 2020-12-15
Fix leak of VMs

xibz:
This commit fixes a leak that occurs when an error has been encountered
in CreateVM

### DIFF
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -465,9 +465,7 @@ func (s *service) CreateVM(requestCtx context.Context, request *proto.CreateVMRe
 	// If we failed to create the VM, we have no point in existing anymore, so shutdown
 	if err != nil {
 		s.shimCancel()
-
 		s.logger.WithError(err).Error("failed to create VM")
-
 		if errors.Cause(err) == context.DeadlineExceeded {
 			return nil, status.Errorf(codes.DeadlineExceeded, "VM %q didn't start within %s: %s", request.VMID, timeout, err)
 		}
@@ -481,7 +479,6 @@ func (s *service) CreateVM(requestCtx context.Context, request *proto.CreateVMRe
 	}
 
 	go s.monitorVMExit()
-
 	// let all the other methods know that the VM is ready for tasks
 	close(s.vmReady)
 
@@ -527,6 +524,15 @@ func (s *service) createVM(requestCtx context.Context, request *proto.CreateVMRe
 	if err != nil {
 		return errors.Wrap(err, "failed to create jailer")
 	}
+
+	defer func() {
+		// in the event of an error, we should stop the VM
+		if err != nil {
+			if e := s.jailer.Stop(true); e != nil {
+				s.logger.WithError(e).Debug("failed to stop firecracker")
+			}
+		}
+	}()
 
 	s.machineConfig, err = s.buildVMConfiguration(request)
 	if err != nil {
--- a/runtime/service_integ_test.go
+++ b/runtime/service_integ_test.go
@@ -1909,7 +1909,6 @@ func requireNonEmptyFifo(t testing.TB, path string) {
 
 func TestCreateVM_Isolated(t *testing.T) {
 	prepareIntegTest(t)
-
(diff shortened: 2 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -103,7 +103,7 @@ steps:
       - "runtime/logs/*"
     command:
       - make -C runtime integ-test FICD_DM_POOL=build_${BUILDKITE_BUILD_NUMBER}_runtime
-    timeout_in_minutes: 15
+    timeout_in_minutes: 20
```
kzys, line 106: Why do you want to change the timeout? Doesn't it work with 15 minutes timeout?

eventbridge/eventbridge.go
```diff
@@ -133,6 +133,9 @@ func Attach(ctx context.Context, source Getter, sink events.Forwarder) <-chan er
 				errChan <- ctx.Err()
 				return
 			default:
+				if source == nil {
```
kzys, line 136: Would it be non-nil in the beginning of Attach but become nil in this select?

runtime/service.go
```diff
@@ -497,10 +494,16 @@ func (s *service) CreateVM(requestCtx context.Context, request *proto.CreateVMRe
 }
 
 func (s *service) publishVMStart() error {
+	if s.eventExchange == nil {
```
kzys, line 497: This function will be panicked without this check. Do we want to have an error instead?

runtime/service.go
```diff
@@ -564,6 +567,15 @@ func (s *service) createVM(requestCtx context.Context, request *proto.CreateVMRe
 		return errors.Wrapf(err, "failed to start the VM")
 	}
 
+	defer func() {
```
kzys, line 570: How about adding this defer right after calling newJailer()?

runtime/service.go
```diff
@@ -1248,6 +1260,9 @@ func (s *service) shutdownLoop(
 		{
 			name: "send a request to the in-VM agent",
 			shutdown: func() error {
+				if s.agentClient == nil {
+					return nil
```
kzys, line 1264: Shouldn't be an error?

runtime/service_integ_test.go
```diff
@@ -1981,34 +2003,32 @@ func TestCreateVM_Isolated(t *testing.T) {
 		requireNonEmptyFifo(t, metricsFile)
 
 		// Some test cases are expected to have an error, some are not.
-		validate(t, createVMErr)
+		s.validate(t, createVMErr)
 
 		// No VM to stop.
```
kzys, line 2008: You should update this "No VM to stop" comment.

kzys (changes requested):

eventbridge/eventbridge.go
xibz: If there was an error during CreateVM, this will error out because we never set the `Getter` resulting in a panic

runtime/service.go
xibz: Hmm, I think I like this at `machine.Start` because that's when the jailer logic would be ran. If we added this after `newJailer` and `machine.Start` wasn't called then this would always error out.

runtime/service.go
xibz: I don't necessarily think so as this would just mean the `agentClient` wasn't even set. But I am indifferent on returning an error or not.

eventbridge/eventbridge.go
kzys: How about handling the case on the caller side? It is odd to see that we are calling eventbridge.Attach() with nil.

runtime/service.go
kzys: I see. I prefer the simplicity of having `defer` right after instantiating an object (like os.Open) rather than remembering which methods are leaving files/processes. But I'm fine with both ways.

runtime/service.go
kzys: Let's return an error then. We log all errors in the loop below. So this would be helpful for operators.

xibz: Turns out all these nil checks were due to a previous change I had and is no longer needed. So I have removed them

kzys (approved):
LGTM!

PR #468, opened 2020-12-18
Extract IO-related functionalities as io.Manager

kzys:
The shim's service struct is getting bigger and bigger. This change
extracts IO-related functionalities as io.Manager to make the service
struct slightly smaller.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/runtime/io/manager.go
+++ b/runtime/io/manager.go
@@ -1,0 +1,167 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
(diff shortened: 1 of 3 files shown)


runtime/io/manager.go
```diff
@@ -0,0 +1,156 @@
...
+const (
+	// taskExecID is a special exec ID that is pointing its task itself.
+	// While the constant is defined here, the convention is coming from containerd.
+	TaskExecID = ""
+
+	minVsockIOPort             = uint32(11000)
+	defaultVSockConnectTimeout = 5 * time.Second
+)
+
+type Manager struct {
```
kzys, line 38: I don't like the name, but not having any better alternatives...

runtime/io/manager.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	return nil
+}
+
+// Find returns FIFO paths regarding the given exec.
+func (m *Manager) Find(taskID, execID string) cio.Config {
+	m.fifosMu.Lock()
+	defer m.fifosMu.Unlock()
+
+	return m.fifos[taskID][execID]
```
IRCody, line 126: Is it not possible here for taskId or execId to not be found?

kzys: Good point. I think it wouldn't be but I'd like to have the check just in case.

kzys:
#483 changed a lot of stuff around IO handling. I'd like to step back and think about the issue more.

PR #473, opened 2021-02-12
Wrap vsock-related errors to explain the context

kzys:
tryConnect() has multiple error cases which would be difficult to
understand. This change adds some context on each of them.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/internal/vm/vsock.go
+++ b/internal/vm/vsock.go
@@ -194,7 +194,7 @@ func vsockConnectMsg(port uint32) string {
 func tryConnect(logger *logrus.Entry, udsPath string, port uint32) (net.Conn, error) {
 	conn, err := net.DialTimeout("unix", udsPath, unixDialTimeout)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "failed to dial %q within %s", udsPath, unixDialTimeout)
 	}
 
 	defer func() {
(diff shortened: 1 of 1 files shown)


internal/vm/vsock.go
```diff
@@ -207,14 +207,19 @@ func tryConnect(logger *logrus.Entry, udsPath string, port uint32) (net.Conn, er
...
+		return nil, vsockConnectMsgError{
+			cause: errors.Wrapf(err, `failed to write %q within %s`, msg, vsockConnectMsgTimeout),
+		}
 	}
 
 	line, err := tryConnReadUntil(conn, '\n', vsockAckMsgTimeout)
 	if err != nil {
-		return nil, vsockAckError{cause: err}
+		return nil, vsockAckError{
+			cause: errors.Wrapf(err, `failed to read "OK <port>" within %s`, vsockAckMsgTimeout),
```
FFMMM, line 221: is there a way to log the `port` number as well? or would we not want to do that?

FFMMM (approved):
thanks for opening this @kzys ! LGTM!

internal/vm/vsock.go
kzys: Let me take a look. I think the port here is assigned by Firecracker.

PR #475, opened 2021-02-12
Fetch 30 commits just in case to correctly check 20 commits

kzys:
git-validation doesn't check Signed-off-by: line (aka DCO) in
merge commits. However, apparently Git cannot recognize merge
commits when one of the parents is missing.

This change fetches 30 commites instead of 21 just in case to
make sure we fetch merge commits with the associated parents.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -33,9 +33,11 @@ jobs:
           # While DCO check (on Makefile) checks latest 20 commits,
           # the checkout action automatically creates a merge commit
           # for merging "master" into a pull request branch.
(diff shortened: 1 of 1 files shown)


alakesh (commented):
How did you come up with the number 30?

kzys:
Both 20 and 30 are arbitrary. It is assuming the pull request has >=20 commits at most and merge commits won't merge very old commits. But we have no guarantees.

alakesh (approved):
Looks good for now.

IRCody (approved):

PR #476, opened 2021-02-18
Increase vsock timeouts and retry intervals.

alakesh:
If host is overloaded, the current retry interval and timeouts may not
be enough and would cause unnecessary retries for request that would
have succeeded otherwise. Increasing it to 150 milliseconds.

*Issue #, if available:*
Current 100 millisecond timeouts may not be enough on loaded systems that can result in unnecessary retries.

*Description of changes:*
This increases the timeouts to 150 milliseconds. This is arbitrary increase but it should give enough window. Ideal approach would be to make this value adaptive based on system load factors.


internal/vm/vsock.go
```diff
@@ -28,9 +28,9 @@ import (
 
 const (
 	vsockRetryTimeout      = 20 * time.Second
-	vsockRetryInterval     = 100 * time.Millisecond
-	unixDialTimeout        = 100 * time.Millisecond
-	vsockConnectMsgTimeout = 100 * time.Millisecond
+	vsockRetryInterval     = 150 * time.Millisecond
```
kzys, line 31: I'm fine about the timeout changes. It is arbitrary, but most customers wouldn't hit the timeouts.

The interval change is different. It would add +50ms launch latency in the worse case by checking the vsock less frequently. Do you really want to do that?

kzys:
Can you rebase the PR against master? This should fix GitHub Actions' failures.

internal/vm/vsock.go
alakesh: It will totally depend on the timing when the VM is ready and when the last unsuccessful retry was made. For example in an experiments run on default VM, we spend 8 retries ( 1200645113 nanoseconds) with 150ms retry interval and 13 retries ( 1300653403 nanoseconds ) with 100ms retry interval. 

The benefit of having a higher retry interval will be to not lose one retry due to the interval not being enough to read the response from firecracker. +50ms in worst case does not seem to be concerning.

kzys:
It would be better to rebase your branch instead of merging changes from master to this branch.

internal/vm/vsock.go
kzys: > The benefit of having a higher retry interval will be to not lose one retry due to the interval not being enough to read the response from firecracker. +50ms in worst case does not seem to be concerning.

We are setting a short read timeout in below. Changing the interval wouldn't affect the timeout I think.

https://github.com/firecracker-microvm/firecracker-containerd/blob/6d5ee656590308dad79d9e88d0bad8ebc0b85dfa/internal/vm/vsock.go#L256

FFMMM:
hey @alakesh what's the status of this CR?

kzys:
Closing due to inactivity.

Issue #477, opened 2021-02-19
Disable serial-getty on our Debian-based microVM root filesystem


kzys:
Our Debian-based micro VM root filesystem has serial-getty inside, and it logs errors like below;

(Originally reported in https://github.com/firecracker-microvm/firecracker-containerd/issues/472)

```
DEBU[2021-02-19T10:07:03.428545928Z] [    1.471698] systemd[756]: serial-getty@ttyS0.service: Executable /sbin/agetty missing, skipping: No such file or directory  jailer=noop runtime=aws.firecracker vmID=f7ac9a76-637a-4fb0-ac9c-8d933fabc4ab vmm_stream=stdout
```

I do think we don't need serial-getty here. We don't assume customers access the microVM's serial console for logging in.

PR #479, opened 2021-02-23
Disable systemd's getty services for debain based rootfs.

alakesh:
*Issue #477  :*

*Description of changes: In the Debian root filesystem getty service is masked, but getty target is not masked and still pointing to systemd binaries. Masking it by pointing it to /dev/null. With this we no longer see the messages like this

DEBU[2021-02-19T10:07:03.428545928Z] [    1.471698] systemd[756]: serial-getty@ttyS0.service: Executable /sbin/agetty missing, skipping: No such file or directory  jailer=noop runtime=aws.firecracker vmID=f7ac9a76-637a-4fb0-ac9c-8d933fabc4ab vmm_stream=stdout

This also saves 0.1 second for the VM to start up, as systemd repeatedly tries to enable serial consoles.

 *

### DIFF
--- a/tools/image-builder/files_debootstrap/etc/systemd/system/firecracker.target.wants/getty.target
+++ b/tools/image-builder/files_debootstrap/etc/systemd/system/firecracker.target.wants/getty.target
@@ -1,1 +1,1 @@
-/lib/systemd/system/getty.target
+/dev/null


kzys (approved):
LGTM. Thanks!

PR #483, opened 2021-02-25
Re-establish a connection when vsock breaks

### LINKED ISSUE #482: Re-establish a pipe -> vsock -> unix domain socket -> pipe connection when it breaks
Right now, detaching a container's stdio doesn't work correctly on firecracker-containerd.

While the firecracker-ctr command below finishes successfully;

```
% sudo =firecracker-ctr -a /run/firecracker-containerd/containerd.sock run \
     --snapshotter devmapper \
     --runtime aws.firecracker \
     --rm --tty --net-host --detach \
     docker.io/library/debian:latest \
     yes-forever6 yes
%
```

firecracker-containerd itself is having the following errors.

```
DEBU[2021-02-25T19:59:31.473222427Z] begin copying io                              ExecID= TaskID=yes-forever6 runtime=aws.firecracker stream=stdout vmID=eec0e852-534b-4182-9b41-a81ca62fd354
DEBU[2021-02-25T19:59:31.473240558Z] begin copying io                              ExecID= TaskID=yes-forever6 runtime=aws.firecracker stream=stdin vmID=eec0e852-534b-4182-9b41-a81ca62fd354
DEBU[2021-02-25T19:59:31.479095080Z] event forwarded                               ns=default topic=/tasks/create type=containerd.events.TaskCreate
DEBU[2021-02-25T19:59:31.479310390Z] event forwarded                               ns=default topic=/tasks/create type=containerd.events.TaskCreate
DEBU[2021-02-25T19:59:31.480545410Z] [    1.079132] agent[705]: time="2021-02-25T19:59:31Z" level=info msg="successfully created task" ExecID= TaskID=yes-forever6 pid_in_vm=728  jailer=noop runtime=aws.firecracker vmID=eec0e852-534b-4182-9b41-a81ca62fd354 vmm_stream=stdout
INFO[2021-02-25T19:59:31.480776524Z] successfully created task                     ExecID= TaskID=yes-forever6 pid_in_vm=728 runtime=aws.firecracker vmID=eec0e852-534b-4182-9b41-a81ca62fd354
DEBU[2021-02-25T19:59:31.481285446Z] start                                         exec_id= runtime=aws.firecracker task_id=yes-forever6
DEBU[2021-02-25T19:59:31.487834549Z] event forwarded                               ns=default topic=/tasks/start type=containerd.events.TaskStart
DEBU[2021-02-25T19:59:31.487930047Z] state                                         exec_id= runtime=aws.firecracker task_id=yes-forever6
DEBU[2021-02-25T19:59:31.487966644Z] event forwarded                               ns=default topic=/tasks/start type=containerd.events.TaskStart
DEBU[2021-02-25T19:59:31.490430495Z] copied 0                                      ExecID= TaskID=yes-forever6 runtime=aws.firecracker stream=stdin vmID=eec0e852-534b-4182-9b41-a81ca62fd354
DEBU[2021-02-25T19:59:31.490491753Z] end copying io                                ExecID= TaskID=yes-forever6 runtime=aws.firecracker stream=stdin vmID=eec0e852-534b-4182-9b41-a81ca62fd354
DEBU[2021-02-25T19:59:31.497902916Z] copied 0                                      ExecID= TaskID=yes-forever6 runtime=aws.firecracker stream=stdout vmID=eec0e852-534b-4182-9b41-a81ca62fd354
ERRO[2021-02-25T19:59:31.497916295Z] error copying io                              ExecID= TaskID=yes-forever6 error="write /proc/self/fd/13: broken pipe" runtime=aws.firecracker stream=stdout vmID=eec0e852-534b-4182-9b41-a81ca62fd354
DEBU[2021-02-25T19:59:31.497944335Z] end copying io                                ExecID= TaskID=yes-forever6 runtime=aws.firecracker stream=stdout vmID=eec0e852-534b-4182-9b41-a81ca62fd354
```

While Firecracker v0.24.2's SIGPIPE fix keeps Firecracker running, `end copying io` here means that the FIFO on the host side is no longer connected to Firecracker.

Thus, attaching to this container doesn't work. There are no explicit errors though.

```
% sudo =firecracker-ctr -a /run/firecracker-containerd/containerd.sock t attach yes-forever6

```

kzys:
IOProxy struct connects the container's stdio inside a micro VM to the FIFO files on its host. containerd clients consume the FIFO files to access the container's stdio.

However, IOProxy was stopping itself when there was an error from the streams, such as EPIPE. In that case, reading from the FIFO files blocks indefinitely since there were no writers.

This change adds ioProxyHandler and uses the handler on service#State() to make sure IOProxy is running. If not, ioProxyHandler#Attach() creates a new IOProxy instance.

*Issue #, if available:*

*Description of changes:*


runtime/service.go
```diff
@@ -1201,10 +1206,54 @@ func (s *service) State(requestCtx context.Context, req *taskAPI.StateRequest) (
...
+		logger.Debug("task is no longer running")
+		return resp, nil
+	}
+
+	resp.Stdin = host.Stdin
+	resp.Stdout = host.Stdout
+	resp.Stderr = host.Stderr
+
+	state, err := s.ioProxyClient.State(requestCtx, &ioproxy.StateRequest{ID: req.ID, ExecID: req.ExecID})
+	if err != nil {
```
xibz, line 1222: What's the correct behaviour here? In the event that we have an error, do we not want to create a proxy? Just want to make sure we aren't missing anything.

kzys: If we cannot call State(), then the chance of having success in the later stage is low, so it would be better to not create a new proxy, I think.

runtime/service.go
```diff
@@ -1201,10 +1206,54 @@ func (s *service) State(requestCtx context.Context, req *taskAPI.StateRequest) (
...
+		StderrPort: attach.StderrPort,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	err = s.taskManager.AttachIO(requestCtx, req.ID, req.ExecID, proxy)
+	if err != nil {
+		return nil, err
+	}
```
kzys, line 1256: Let me extract this one as a new function.

runtime/service.go
xibz: What kind of errors do we expect to see here in the event of an error?

xibz:
Do we also know what's causing the connections to be broken?

runtime/service.go
kzys: There is a chance that a process was running when we call`s.agentClient.State(requestCtx, req)`, but exited before calling `s.ioProxyClient.State(requestCtx, &ioproxy.StateRequest{ID: req.ID, ExecID: req.ExecID})`. Then this state call would have codes.NotFound.

kzys:
Yes. we have FIFO files on the host and FIFO files on the guest. The host-side FIFO files are consumed by containerd clients (including `ctr`). In #482, ctr process was exiting so that FIFO readers were not there anymore that resulted EPIPE.

https://linux.die.net/man/2/write

> fd is connected to a pipe or socket whose reading end is closed. When this happens the writing process will also receive a SIGPIPE signal. (Thus, the write return value is seen only if the program catches, blocks or ignores this signal.) 

Assuming that a in-container process is handling SIGPIPE, the process could handle EPIPE and keep writing. So that new readers could read the FIFO files later.

However before this PR's changes, EPIPE entirely stops IOProxy, so that new readers could read pipes, but wouldn't see new bytes. Moreover it would block indefinitely since these FIFO files wouldn't have writers.

xibz (approved):
LGTM

kzys:
Thanks! Let me add the issue number on the last commit.

PR #489, opened 2021-03-19
docker build and run needs root privilege

alakesh:
On a fresh checkout and build of firecracker containerd as non-root user
would fail as docker build and run need root privilege.

*Issue #, if available:*

*Description of changes:*

To prevent checkout and build using root account, we can just run these two steps in Makefile with root privilege.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -291,14 +291,14 @@ $(FIRECRACKER_DIR)/Cargo.toml:
 	git submodule update --init --recursive $(FIRECRACKER_DIR)
 
 tools/firecracker-builder-stamp: tools/docker/Dockerfile.firecracker-builder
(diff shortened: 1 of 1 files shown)


kzys:
I think asking a user to add themselves on docker group is better. https://github.com/firecracker-microvm/firecracker-containerd/blob/master/docs/quickstart.md has `sudo usermod -aG docker $(whoami)`

alakesh:
> I think asking a user to add themselves on docker group is better. https://github.com/firecracker-microvm/firecracker-containerd/blob/master/docs/quickstart.md has `sudo usermod -aG docker $(whoami)`

Agreed. Testing that and it works fine. Closing this PR.

PR #490, opened 2021-03-22
extend quickstart script to support rpm based distro

alakesh:
Most of the steps for installation, configuration and running of
firecracker containerd on rpm based linux distro is similar to debian.
Added an alternate quickstart script for Amazon Linux 2. Running on
a similar rpm based distro e.g. centos would need minimal changes.
For example instead of installing docker-ce using amazon-linux-extras,
it can be installed using yum install.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -20,8 +20,8 @@ files into `/usr/local/bin`.
    on [this page](https://wiki.debian.org/Cloud/AmazonEC2Image/Buster).
    If you need help launching an EC2 instance, see the
    [EC2 getting started guide](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EC2_GetStarted.html).
-3. Run the script below to download and install all the required components.
-   This script expects to be run from your `$HOME` directory.
+3. Run the script below to download and install all the required dependencies on a debian based instance.
+   Alternate steps for rpm based instance is provided as well. This script expects to be run from your `$HOME` directory.
 
 ```bash
 #!/bin/bash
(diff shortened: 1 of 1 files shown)


docs/quickstart.md
```diff
@@ -174,6 +176,152 @@ sudo tee /etc/containerd/firecracker-runtime.json <<EOF
...
+# * a Debian-based root filesystem configured as read-only with a read-write
+#   overlay
+# * firecracker-containerd, an alternative containerd binary that includes the
+#   firecracker VM lifecycle plugin and API
+# * tc-redirect-tap and other CNI dependencies that enable VMs to start with
+#   access to networks available on the host
+git clone https://github.com/firecracker-microvm/firecracker-containerd.git
+cd firecracker-containerd
+sudo yum -y install device-mapper
+sg docker -c 'make all image firecracker'
```
kzys, line 227: Does it need sg since the user is in docker group?

docs/quickstart.md
```diff
@@ -174,6 +176,152 @@ sudo tee /etc/containerd/firecracker-runtime.json <<EOF
...
+  address = "/run/firecracker-containerd/containerd.sock"
+[plugins]
+  [plugins.devmapper]
+    pool_name = "fc-dev-thinpool"
+    base_image_size = "10GB"
+    root_path = "/var/lib/firecracker-containerd/snapshotter/devmapper"
+
+[debug]
+  level = "debug"
+EOF
```
kzys, line 256: Are these steps distro-agnostic? If so, how about having a section for per-distro commands (apt-get vs. yum) and a section for distro-agnostic commands?

kzys (changes requested):

docs/quickstart.md
alakesh: Even though we add the user to docker group, without running with sg the Gid of a process run by the user is still its primary group. This can be verified by reading /proc/<pid>/status  But if we run with sg the Gid is seen to be of docker group. So i think its better to run with sg.

docs/quickstart.md
```diff
@@ -62,6 +62,51 @@ sudo DEBIAN_FRONTEND=noninteractive apt-get \
...
+
+A similar script to install dependencies for rpm based linux distro e.g. Amazon Linux 2 can be found here
+
+<details>
+
+```bash
+#!/bin/bash
+
+cd ~
+
```
kzys, line 77: Do you need `cd` here and below? All commands on the section won't read/write files in the current working directory.

kzys (approved):

PR #491, opened 2021-03-23
Enables jailer to copy cache topology

xibz:
unfortunately due to differences in arm, arm does not setup a cache
topology. This commit fixes that copying the system's cache topology
into the jailed rootfs

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/runtime/runc_jailer.go
+++ b/runtime/runc_jailer.go
@@ -41,7 +41,27 @@ import (
 
 const (
 	networkNamespaceRuncName = "network"
+	cacheTopologyPath        = "/sys/devices/system/cpu/cpu0/cache"
+	cacheFolderPrefix        = "index"
 )
+
+var cacheTopologyPaths = []string{
+	"sys",
+	"devices",
+	"system",
+	"cpu",
+	"cpu0",
+	"cache",
+}
+
+var cacheTopologyFiles = []string{
+	"level",
+	"type",
+	"size",
+	"number_of_sets",
+	"shared_cpu_map",
+	"coherency_line_size",
+}
 
 // runcJailer uses runc to set up a jailed environment for the Firecracker VM.
 type runcJailer struct {
@@ -253,6 +273,10 @@ func (j *runcJailer) BuildJailedRootHandler(cfg *config.Config, machineConfig *f
 				filename := filepath.Base(v.Path)
 				v.Path = filepath.Join("/", filename)
 				m.Cfg.VsockDevices[i] = v
+			}
+
+			if err := j.setupCacheTopology(rootPath); err != nil {
+				return err
 			}
 
 			j.logger.Debugf("Writing %q for runc", rootPathToConfig)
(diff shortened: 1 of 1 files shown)


kzys:
Firecacker's jailer is doing something similar. Should it be owned by Go SDK?

https://github.com/firecracker-microvm/firecracker/blob/646ec96733d6aa72586e69dc9bcd0b38dab2df7d/src/jailer/src/env.rs#L282

xibz:
@kzys the SDK has no knowledge of jailing. So no, it probably doesn't belong there imo

kzys:
Oh sorry. This is for runc...

runtime/runc_jailer.go
```diff
@@ -41,8 +41,19 @@ import (
...
 )
 
+var cacheTopologyPaths = []string{
+	"sys",
+	"devices",
+	"system",
+	"cpu",
+	"cpu0",
+	"cache",
+}
```
kzys, line 55: - Does the directory has `cpu0` under `cpu0/cache`?
- What we copy is different from what Firecracker's jailer copies in https://github.com/firecracker-microvm/firecracker/pull/2334. Do you know the reason? (cc @dianpopa)

runtime/runc_jailer.go
```diff
@@ -393,6 +408,25 @@ func (j *runcJailer) copyFileToJail(src, dst string, mode os.FileMode) error {
...
+// copyDirectoryToJail will copy all of the files and the parent directory to
+// the given destination folder, dst. This will also set the proper ownership
+// to the files and folders.
+func (j *runcJailer) copyDirectoryToJail(src, dst string) error {
+	output, err := exec.Command("cp", "-rf", src, dst).CombinedOutput()
+	if err != nil {
+		return errors.Wrapf(err, "failed to copy directory %q to %q: %s", src, dst, output)
+	}
+
+	filepath.Walk(dst, func(path string, info os.FileInfo, err error) error {
```
kzys, line 420: I think we need to check the error from filepath.Walk.

runtime/runc_jailer.go
```diff
@@ -640,3 +674,43 @@ func (j runcJailer) Stop(force bool) error {
...
+
+	// builds the cache topology path from the root directory
+	for _, p := range cacheTopologyPaths {
+		path = filepath.Join(path, p)
+		if err := mkdirAndChown(path, mode, j.Config.UID, j.Config.GID); err != nil {
+			return err
+		}
+	}
+
+	filepath.Walk(cacheTopologyPath, func(cachePath string, info os.FileInfo, err error) error {
```
kzys, line 693: Check the error from `filepath.Walk`

kzys (changes requested):

runtime/runc_jailer.go
xibz: Yea, I asked Diana and she told me today what files to copy. And yes, all these directories will exist.

runtime/runc_jailer.go
xibz: Will fix

xibz:
@kzys Updated to copy the required files for the cache topology and the `filepath.Walk` error

runtime/runc_jailer.go
```diff
@@ -640,3 +664,52 @@ func (j runcJailer) Stop(force bool) error {
...
+			return err
+		}
+		if !info.IsDir() {
+			return err
+		}
+
+		folder := filepath.Base(cachePath)
+		if !strings.HasPrefix(folder, cacheFolderPrefix) {
+			return err
+		}
```
kzys, line 694: How about returning nil explicitly here?

kzys (approved):

runtime/runc_jailer.go
xibz: Cool, will fix this. Forgot that I returned if `err` was not `nil` at the top of the closure.

PR #493, opened 2021-04-06
Memory ballooning support on firecracker-containerd

RoyceDavison:

*Issue #, if available:*

*Description of changes:*
Implement the memory ballooning APIs on firecracker-containerd. The swagger definition can be found [HERE](https://github.com/firecracker-microvm/firecracker/blob/master/src/api_server/swagger/firecracker.yaml#L63-L128).


proto/firecracker.proto
```diff
@@ -92,6 +95,40 @@ message GetVMMetadataResponse {
...
+    string VMID = 1;
+    int64 StatsPollingIntervals = 2;
+}
+
+message GetBalloonStatsRequest {
+    string VMID = 1;
+}
+
+message GetBalloonStatsResponse {
+    int64 ActualMb = 1; // Actual amount of memory (in MB) the device is holding.
```
FFMMM, line 113: nit indentation.

proto/types.proto
```diff
@@ -167,3 +167,9 @@ message FirecrackerTokenBucket {
 	int64 RefillTime = 2; // Specifies the amount of millis it takes for the bucket to fill
 	int64 Capacity = 3; // Specifies the number of tokens this bucket can hold
 }
+
+message FirecrackerBalloonDevice {
+    int64 AmountMb = 1; //Target balloon size in MB.
+    bool DeflateOnOom = 2; // Whether the balloon should deflate when the guest has memory pressure.
+    int64 StatsPollingIntervals = 3; // Interval in seconds between refreshing statistics.
+}
```
FFMMM, line 175: nit indentation

runtime/drive_handler_test.go
```diff
@@ -95,7 +95,7 @@ func TestContainerStubs(t *testing.T) {
 		mockMachine, err := firecracker.NewMachine(ctx, firecracker.Config{}, firecracker.WithClient(
 			firecracker.NewClient("/path/to/socket", nil, false, firecracker.WithOpsClient(&fctesting.MockClient{
 				PatchGuestDriveByIDFn: func(params *ops.PatchGuestDriveByIDParams) (*ops.PatchGuestDriveByIDNoContent, error) {
-					assert.Equal(t, hostPath, firecracker.StringValue(params.Body.PathOnHost))
+					assert.Equal(t, hostPath, firecracker.StringValue(&params.Body.PathOnHost))
```
FFMMM, line 98: why the `&` here and on line `243` ?

runtime/service.go
```diff
@@ -769,6 +791,115 @@ func (s *service) GetVMMetadata(requestCtx context.Context, request *proto.GetVM
...
+	balloonStats, err := s.machine.GetBalloonStats(requestCtx)
+	if err != nil {
+		err = errors.Wrap(err, "failed to get balloon statistics")
+		s.logger.WithError(err).Error()
+		return nil, err
+	}
+
+	// TODO: decide a better way for safely dereferencing?
+	if balloonStats.ActualMb == nil {
+		s.logger.Errorf("Failed to dereference nil ptr: alloonStats.ActualMb")
```
FFMMM, line 873: Just a suggestion but instead of saying failed to dereference, maybe we can change the err message to saying that "X setting has not passed in the config".

---

Getting a bit more philosophical here, why do we want to catch these errs in the first place? Should we even let the init continue if these haven't been set by the client/ callers?

go.mod
```diff
@@ -54,4 +54,6 @@ require (
 // Workaround for github.com/containerd/containerd issue #3031
 replace github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
 
+replace github.com/firecracker-microvm/firecracker-go-sdk => github.com/RoyceDavison/firecracker-go-sdk v0.22.1-0.20210402232445-4bb6cbb0e7f8
```
FFMMM, line 57: hm should this be here?

RoyceDavison: Yeah, this shouldn't be here. As my commit message suggested "add go.mod go.sum for integ tests, will remove it later"

runtime/drive_handler_test.go
RoyceDavison: This is because the original definition is changed. As you can find [HERE](https://github.com/firecracker-microvm/firecracker-go-sdk/blob/45eff16db190c59d1dd04d9ffc402b06317a2970/pointer_helpers.go). This definition changed is not made by me. To use the new StringValue(), I have to pass an address.

proto/firecracker.proto
RoyceDavison: Yeah, I noticed these. But I have no idea how to format firecracker.proto. Is there a tool we used like "gofmt -w" to help format protobuf? I saw there was some on github wrote by others but not sure if I can use it for this project.

runtime/service.go
RoyceDavison: Because these pointers should have some values associated with them if the client made a call, iirc

runtime/service.go
```diff
@@ -769,6 +785,142 @@ func (s *service) GetVMMetadata(requestCtx context.Context, request *proto.GetVM
...
+		s.logger.WithError(err).Error()
+		return nil, err
+	}
+
+	if balloonStats.ActualMb == nil ||
+		balloonStats.ActualPages == nil ||
+		balloonStats.TargetMb == nil ||
+		balloonStats.TargetPages == nil {
+		s.logger.Errorf("One of BalloonStats properties is nil, please check ActualMb, ActualPages, TargetMb, TargetPages")
+	}
```
FFMMM, line 880: Could we actually point out which stat is `nil`? 

I think it could be worth it but others can argue against that.

FFMMM:
overall this LGTM!! Quick nit

* Do we want to include instructions on how to use/ configure the mem ballon device in one of our `.md`s ?

RoyceDavison:
> overall this LGTM!! Quick nit
> 
> * Do we want to include instructions on how to use/ configure the mem ballon device in one of our `.md`s ?

Good catch.
I think we may want to add some details about balloon's behaviors? But all of these can be found in [Firecracker balloon docs](https://github.com/firecracker-microvm/firecracker/blob/main/docs/ballooning.md). 
I am not sure if it is worth to repeat again?

runtime/service.go
RoyceDavison: I just feel it is a good practice to check before deference a pointer. Open to discuss.

proto/firecracker.proto
```diff
@@ -135,3 +137,44 @@ message JailerConfig {
...
+    string VMID = 1;
+    int64 AmountMb = 2;
+}
+
+message GetBalloonConfigRequest{
+    string VMID = 1;
+}
+
+message GetBalloonConfigResponse{
+    FirecrackerBalloonDevice BalloonConfig = 6;
```
kzys, line 150: The number on the right hand side is the order of the fields, to be backward-compatible while adding/removing fields. So you should start from 1 here.

runtime/service.go
```diff
@@ -769,6 +785,142 @@ func (s *service) GetVMMetadata(requestCtx context.Context, request *proto.GetVM
...
+	amountMB := request.BalloonDevice.AmountMb
+	deflateOnOom := request.BalloonDevice.DeflateOnOom
+	statsPollingIntervals := request.BalloonDevice.StatsPollingIntervals
+
+	balloonDevice := firecracker.NewBalloonDevice(amountMB, deflateOnOom, firecracker.WithStatsPollingIntervals(statsPollingIntervals))
+	balloon := balloonDevice.Build()
+
+	if balloon.AmountMb == nil || balloon.DeflateOnOom == nil {
+		return balloon, errors.Errorf("One of balloon properties is nil, please check AmountMb, DeflateOnOom")
+	}
```
kzys, line 798: If these checks are coming from Firecracker, we probably should have them on the Go SDK.

runtime/service.go
kzys: Are they nil-able on Firecracker's schema?

I agree it is better to check that before de-referencing, but this code only logs error after that. You must not de-reference them if they are nil.

runtime/service_integ_test.go
```diff
@@ -247,6 +250,86 @@ func createTapDevice(ctx context.Context, tapName string) error {
...
+func TestMemoryBalloon_Isolated(t *testing.T) {
+	prepareIntegTest(t)
+	ctx := namespaces.WithNamespace(context.Background(), defaultNamespace)
+
+	pluginClient, err := ttrpcutil.NewClient(containerdSockPath + ".ttrpc")
+	require.NoError(t, err, "failed to create ttrpc client")
+
+	numVMs := 5
+	var vmGroup sync.WaitGroup
+	for i := 0; i < numVMs; i++ {
```
kzys, line 262: Why do you want to have 5 VMs?

kzys (changes requested):

proto/firecracker.proto
RoyceDavison: Good catch, thanks. Fixed it.

runtime/service.go
RoyceDavison: Fixed it.

runtime/service_integ_test.go
RoyceDavison: > Are they nil-able on Firecracker's schema? I agree it is better to check that before de-referencing, but this code only logs error after that. You must not de-reference them if they are nil."

Nope, they aren't nil-able. But I still think it is a good practice to leave this safe deference check, I just changed the log to return an error. 

> Why do you want to have 5 VMs?
It is an arbitrary number. I just wanted to test if it works with more than 1 VMs.

proto/firecracker.proto
```diff
@@ -135,3 +137,44 @@ message JailerConfig {
...
+    FirecrackerBalloonDevice BalloonConfig = 1;
+}
+
+message GetBalloonStatsRequest {
+    string VMID = 1;
+}
+
+message GetBalloonStatsResponse{
+    int64 ActualMb = 1; // Actual amount of memory (in MB) the device is holding.
+	int64 ActualPages = 2; // Actual number of pages the device is holding.
```
kzys, line 159: The indent here looks off.

runtime/service.go
```diff
@@ -769,6 +785,140 @@ func (s *service) GetVMMetadata(requestCtx context.Context, request *proto.GetVM
...
+	statsPollingIntervals := request.BalloonDevice.StatsPollingIntervals
+
+	balloonDevice := firecracker.NewBalloonDevice(amountMB, deflateOnOom, firecracker.WithStatsPollingIntervals(statsPollingIntervals))
+	balloon := balloonDevice.Build()
+
+	s.logger.Infof("Creating a balloon device: AmountMb=%d, DeflateOnOom=%t and statsPollingIntervals=%d ", *balloon.AmountMb, *balloon.DeflateOnOom, balloon.StatsPollingIntervals)
+	return balloon, nil
+}
+
+func (s *service) BuildBalloonDeviceOpt(balloon models.Balloon) ([]firecracker.Opt, error) {
```
kzys, line 800: Does it have to be exported to other packages?

runtime/service.go
```diff
@@ -769,6 +785,140 @@ func (s *service) GetVMMetadata(requestCtx context.Context, request *proto.GetVM
...
+		err = errors.Wrap(err, "failed to get balloon statistics")
+		s.logger.WithError(err).Error()
+		return nil, err
+	}
+
+	if balloonStats.ActualMb == nil ||
+		balloonStats.ActualPages == nil ||
+		balloonStats.TargetMb == nil ||
+		balloonStats.TargetPages == nil {
+		err = errors.Wrap(err, "One of BalloonStats properties is nil, please check ActualMb, ActualPages, TargetMb, TargetPages")
```
kzys, line 875: - As @FFMMM pointed out, this function knows which properties are nil, we should specifically mention the properties or include balloonStats in the error message.
- `errors.Wrap(err)` - but err here is nil.

kzys (commented):
Almost looks good to me.

proto/firecracker.proto
RoyceDavison: Yeah, I noticed it. I don't know how to fix it, since it look correctly in firecracker.proto. I tried various ways to fix it such as  leave a blank above the line 158, move the comment from line 158 to line 157, but none of them can fix it.

runtime/service.go
RoyceDavison: Nope. Fixed it.

runtime/service.go
RoyceDavison: Fixed it.

proto/firecracker.proto
RoyceDavison: Fixed it.

runtime/service.go
RoyceDavison: Fixed it.

runtime/service.go
RoyceDavison: Fixed it.

kzys (approved):

PR #498, opened 2021-04-20
Restructure docs/architecture.md and add a new diagram

kzys:
This change restructures architecture.md and add a new high-level
diagram.

The new diagram is highlighting components we own, clarify process
boundaries and omit details which we don't touch such as content
store.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -3,53 +3,42 @@
 firecracker-containerd aims to make it easier to run containers with virtual
 machine isolation provided by the [Firecracker virtual machine monitor
 (VMM)](https://github.com/firecracker-microvm/firecracker).
-firecracker-containerd integrates with [containerd](https://containerd.io) by
-implementing its interfaces and extending its APIs.
 
-firecracker-containerd implements the following containerd interfaces:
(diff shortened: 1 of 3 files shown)


RoyceDavison:
LGTM 🚀

kzys:
@samuelkarp @IRCody @xibz [The new diagram](https://github.com/kzys/firecracker-containerd/blob/architecture-diagram/docs/img/overview.svg) is simpler than [what we have in main](https://github.com/firecracker-microvm/firecracker-containerd/blob/main/docs/img/architecture-diagram.png).

What do you think? Are there any details we'd like to keep?

samuelkarp:
> What do you folks think about the omission of the details here? Are there any details you'd like to keep?

I think omitting the VM disk image and kernel are reasonable.  I'd like to see an example network configuration continue to be represented (at a minimum the tap device, but showing how a CNI plugin can be used is useful too).

It's also useful to include the detail that the communication from the shim to the agent happens over a vsock rather than a network socket.

kzys:
Let me redo this one.

Issue #492, opened 2021-04-01
StopVM must not stuck when the VM itself is paused


kzys:
Right now StopVM calls an agent inside the VM. However after having pause/resume, the VM could be paused. In that case, StopVM will stuck since it cannot establish the connection to the VM.

We may need to either

- Implicitly resume the VM
- Kill the VM without resuming

Or consider 
- returning an error when the VM is in that state.

In addition to that, StopVM has "force kill" which SIGKILL the Firecracker process. Why didn't it work on #487?

alakesh:
We can do the following:
1. Detect if the VM is in paused
2. If the VM is paused, then resume and remount its FS readonly ( to prevent data corruption )
3. Force kill the VM ( if that works ).

Meanwhile to be able to do that we need api to obtain VM's running state.  Following PR would facilitate this if approved
https://github.com/firecracker-microvm/firecracker-go-sdk/pull/338

RoyceDavison:
Isn't the vm running state information stored in machineConfig? https://github.com/firecracker-microvm/firecracker-containerd/blob/f320d3636aee41661eb525b284ce6213f6c7a3d5/runtime/service.go#L679

alakesh:
> Isn't the vm running state information stored in machineConfig? How about call GetVMInfo()?
> 
> https://github.com/firecracker-microvm/firecracker-containerd/blob/f320d3636aee41661eb525b284ce6213f6c7a3d5/runtime/service.go#L679

It does not return the VM state [ running/paused/resumed ].

alakesh:
Was thinking, if this issue can only happen with StopVM. Ideally any call that depends on the agent running inside the VM would need to check if the VM is running. @kzys? If that is the case, this issue would be a good place to address all such cases.

alakesh:
With PR #500 PR #502  and PR #519  merged, this issue can be closed.

PR #502, opened 2021-04-27
Fix for StopVM to force shutdown VM is in paused state in order to avoid hang

alakesh:
*Issue #, if available:*
https://github.com/firecracker-microvm/firecracker-containerd/issues/492

*Description of changes:*
a. Update go-sdk to have changes that implement DescribeInstanceInfo in sdk 
b. Modify StopVM to first check if instance is stopped and return error if it is, rather than hang trying to connect to agent..

### DIFF
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -636,7 +636,8 @@ func (s *service) mountDrives(requestCtx context.Context) error {
 
 // StopVM will shutdown the VMM. Unlike Shutdown, this method is exposed to containerd clients.
 // If the VM has not been created yet and the timeout is hit waiting for it to exist, an error will be returned
-// but the shim will continue to shutdown.
+// but the shim will continue to shutdown. Similarly if we detect that the VM is in pause state, then
+// we are unable to communicate to the in-VM agent. In this case, we do a forceful shutdown.
 func (s *service) StopVM(requestCtx context.Context, request *proto.StopVMRequest) (_ *empty.Empty, err error) {
 	defer logPanicAndDie(s.logger)
 	s.logger.WithFields(logrus.Fields{"timeout_seconds": request.TimeoutSeconds}).Debug("StopVM")
@@ -644,6 +645,20 @@ func (s *service) StopVM(requestCtx context.Context, request *proto.StopVMReques
 	timeout := defaultStopVMTimeout
 	if request.TimeoutSeconds > 0 {
 		timeout = time.Duration(request.TimeoutSeconds) * time.Second
+	}
+
+	info, err := s.machine.DescribeInstanceInfo(requestCtx)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get instance info %v", info)
+	}
+
+	if *info.State == models.InstanceInfoStatePaused {
+		s.logger.Debug("Instance is in Paused state, force shutdown in progress")
+		err = s.jailer.Stop(true)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to stop VM in paused State")
+		}
+		return &empty.Empty{}, nil
 	}
 
 	err = s.waitVMReady()
--- a/runtime/service_integ_test.go
+++ b/runtime/service_integ_test.go
@@ -1537,6 +1537,7 @@ func TestStopVM_Isolated(t *testing.T) {
 		stopFunc        func(ctx context.Context, fcClient fccontrol.FirecrackerService, req proto.CreateVMRequest)
 		withStopVM      bool
 	}{
+
 		{
 			name:       "Successful",
 			withStopVM: true,
@@ -1624,6 +1625,22 @@ func TestStopVM_Isolated(t *testing.T) {
 				} else {
 					assert.Contains(err.Error(), "signal: killed")
 				}
+			},
+		},
+
+		// Test that StopVM returns success if the VM is in paused state, instead of hanging forever.
+		// VM is force shutdown in this case, so we expect no Error or hang.
+		{
+			name:       "PauseStop",
+			withStopVM: true,
+
+			createVMRequest: proto.CreateVMRequest{},
+			stopFunc: func(ctx context.Context, fcClient fccontrol.FirecrackerService, req proto.CreateVMRequest) {
+				_, err = fcClient.PauseVM(ctx, &proto.PauseVMRequest{VMID: req.VMID})
+				require.Equal(status.Code(err), codes.OK)
+
+				_, err = fcClient.StopVM(ctx, &proto.StopVMRequest{VMID: req.VMID})
+				require.NoError(err)
 			},
 		},
 	}


proto/firecracker.proto
```diff
@@ -92,6 +92,17 @@ message GetVMMetadataResponse {
     string Metadata = 1;
 }
 
+message DescribeInstanceInfoRequest {
```
kzys, line 95: I prefer to not expose the API to customers right now. Removing/deprecating APIs is hard and we should do that once we really know the use-case. We can fix #492 by calling DescribeInstance by ourselves.

kzys (changes requested):

proto/firecracker.proto
alakesh: Not sure what you meant by "calling DescribeInstance by ourselves".  Do you mean we should internally call the function from StopVM call?

alakesh: Anyway, implemented the fix for StopVM without needing to add another API in firecracker containerd. This should serve the purpose for now.

go.mod
```diff
@@ -54,4 +54,6 @@ require (
 // Workaround for github.com/containerd/containerd issue #3031
 replace github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
 
+replace github.com/golang/protobuf => github.com/golang/protobuf v1.3.1
```
kzys, line 57: Can you explain "why" as like the replace directive above?

runtime/service.go
```diff
@@ -627,6 +627,13 @@ func (s *service) StopVM(requestCtx context.Context, request *proto.StopVMReques
 		timeout = time.Duration(request.TimeoutSeconds) * time.Second
 	}
 
+	if info, err := s.machine.DescribeInstanceInfo(requestCtx); err != nil {
+		return nil, err
+	} else if *info.State == models.VMStatePaused {
+		s.logger.Debugf("Instance is paused, can't stop VM")
+		return nil, errors.New("VMPaused")
```
kzys, line 634: - You don't have to use Debugf if the string doesn't have any format directives such as `%s`.
- Can you check other error messages we have? I think we are returning a full sentence rather than something like `VMPaused`

kzys (changes requested):
Please add some integration tests, or update existing integration tests to trigger the condition.

go.mod
RoyceDavison: gogo/protobuf is not compatible with APIv2 golang/protobuf 1.4, the same issue is described in [contaierd/ttrpc#62](https://github.com/containerd/ttrpc/issues/62). Also if we ever use go mod tidy, it gets automatically upgrade to v1.4 so we need to manually downgrade it w/ replace.

runtime/drive_handler_test.go
```diff
@@ -95,7 +95,7 @@ func TestContainerStubs(t *testing.T) {
 		mockMachine, err := firecracker.NewMachine(ctx, firecracker.Config{}, firecracker.WithClient(
 			firecracker.NewClient("/path/to/socket", nil, false, firecracker.WithOpsClient(&fctesting.MockClient{
 				PatchGuestDriveByIDFn: func(params *ops.PatchGuestDriveByIDParams) (*ops.PatchGuestDriveByIDNoContent, error) {
-					assert.Equal(t, hostPath, firecracker.StringValue(params.Body.PathOnHost))
+					assert.Equal(t, hostPath, firecracker.StringValue(&params.Body.PathOnHost))
```
RoyceDavison, line 98: How about create a new PR for this change. What do you think @kzys ?

runtime/service.go
```diff
@@ -627,6 +627,13 @@ func (s *service) StopVM(requestCtx context.Context, request *proto.StopVMReques
 		timeout = time.Duration(request.TimeoutSeconds) * time.Second
 	}
 
+	if info, err := s.machine.DescribeInstanceInfo(requestCtx); err != nil {
+		return nil, err
```
RoyceDavison, line 631: How about wrap additional context with this error, such as errors.Wrap(err, "failed to get instance info")?

runtime/service.go
```diff
@@ -627,6 +627,13 @@ func (s *service) StopVM(requestCtx context.Context, request *proto.StopVMReques
...
+	} else if *info.State == models.VMStatePaused {
```
RoyceDavison, line 632: Should we check for safe dereferencing?

runtime/service.go
alakesh: Added in most recent commit

runtime/drive_handler_test.go
alakesh: This error shows up after updating to most recent go-sdk.  Yes we can make a separate PR but not sure how different it will be. Right now keeping it as a separate commit.

RoyceDavison: Yeah, I just saw your separate commit for this. It is enough though imo.

go.mod
kzys: Thanks. @alakesh Can you add that on go.mod file? People could find Royce's comment from `git annotate` though.

runtime/service.go
kzys: You need to return the wrapped error. `errors.Wrap()` won't mutate the err :)

runtime/service_integ_test.go
```diff
@@ -1546,6 +1547,35 @@ func TestStopVM_Isolated(t *testing.T) {
 				}
 			},
 		},
+
+		{
+			name:       "Unsuccessful",
```
FFMMM, line 1552: nit: indents

runtime/service_integ_test.go
```diff
@@ -1546,6 +1547,35 @@ func TestStopVM_Isolated(t *testing.T) {
...
+					err := firecrackerProcess.KillWithContext(ctx)
+					require.NoError(err, "failed to kill firecracker process")
+				}()
+
+				_, err = fcClient.StopVM(ctx, &proto.StopVMRequest{
+					VMID:           req.VMID,
+					TimeoutSeconds: 10,
+				})
+				require.Error(err)
+				//assert.Contains(err.Error(), "cannot stop VM in paused state")
```
FFMMM, line 1576: do we need this cmtd out?

runtime/service_integ_test.go
```diff
@@ -1546,6 +1547,35 @@ func TestStopVM_Isolated(t *testing.T) {
 				}
 			},
 		},
+
+		{
```
FFMMM, line 1551: Please add some high level indication on what this is testing.

runtime/service_integ_test.go
```diff
@@ -1546,6 +1547,35 @@ func TestStopVM_Isolated(t *testing.T) {
...
+			name:       "Unsuccessful",
+			withStopVM: true,
+
+			createVMRequest: proto.CreateVMRequest{},
+			stopFunc: func(ctx context.Context, fcClient fccontrol.FirecrackerService, req proto.CreateVMRequest) {
```
FFMMM, line 1556: more of an open-ended q but curious if we wanted to start assigning the funcs a var name.

imo, it would be useful for readability but unless we can eke out some reuse i am not too partial to any option.

runtime/service.go
kzys: @alakesh Please fix this one.

runtime/service.go
```diff
@@ -646,6 +647,21 @@ func (s *service) StopVM(requestCtx context.Context, request *proto.StopVMReques
...
 	}
 
+	info, err := s.machine.DescribeInstanceInfo(requestCtx)
+	if err != nil {
+		errors.Wrapf(err, "failed to get instance info %v", info)
+		return nil, err
+	}
+
+	if *info.State == models.InstanceInfoStatePaused {
+		s.logger.Debugf("Instance is paused state: %v, force shutdown in progress", *info.State)
```
kzys, line 657: Wouldn't it print `Instance is paused state: Paused, force shutdown in progress`? Does it have to say "Paused" twice?

runtime/service_integ_test.go
alakesh: removed. Thanks

runtime/service_integ_test.go
alakesh: Done

kzys (approved):

PR #495, opened 2021-04-16
Update vmlinux kernel

RoyceDavison:
Signed-off-by: Royce Zhao <qiqinzha@amazon.com>
Update download link to fetch vmlinux kernel and its sha256sum. The new vmlinux kernel will support memory ballooning feature. More information about the prerequisites for the kernel to support memory ballooning can be found [HERE](https://github.com/firecracker-microvm/firecracker/blob/main/docs/ballooning.md#prerequisites)

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #503, opened 2021-04-30
Update outdated slack invitation & kernel download links

RoyceDavison:
This PR has two things changed. 
1. There is a merged PR #495 for updating kernel download link on Makefile. I think we should also update the kernel download link on docs to keep them consistent. The new kernel download link comes from https://github.com/firecracker-microvm/firecracker/blob/main/docs/getting-started.md#running-firecracker

2. Update the slack channel invitation link, as https://github.com/firecracker-microvm/firecracker-go-sdk/issues/340 is pointing out. 

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ GitHub issues, but instead follow
 guidelines](https://github.com/firecracker-microvm/firecracker/blob/master/SECURITY-POLICY.md).
 
 Other discussion: For general discussion, please join us in the `#containerd`
-channel on the [Firecracker Slack](https://tinyurl.com/firecracker-microvm).
+channel on the [Firecracker Slack](https://join.slack.com/t/firecracker-microvm/shared_invite/zt-oxbm7tqt-GLlze9zZ7sdRSDY6OnXXHg).
 
 ## License
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -51,7 +51,7 @@ You need to have the following things in order to use firecracker-containerd:
 You can use the following kernel:
 
 ```bash
-curl -fsSL -o hello-vmlinux.bin https://s3.amazonaws.com/spec.ccfc.min/img/hello/kernel/hello-vmlinux.bin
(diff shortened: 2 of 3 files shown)


kzys:
Generally speaking, it would be better to mention "why" rather than "what" on commit messages.

"Updating README.md, Quickstart, .." is technically correct, but people can understand that by just looking your commit. The important part is why do you need to update these files.

See also: https://thoughtbot.com/blog/5-useful-tips-for-a-better-commit-message

RoyceDavison:
> Generally speaking, it would be better to mention "why" rather than "what" on commit messages.
> 
> "Updating README.md, Quickstart, .." is technically correct, but people can understand that by just looking your commit. The important part is why do you need to update these files.
> 
> See also: https://thoughtbot.com/blog/5-useful-tips-for-a-better-commit-message

Thanks, fixed it.

README.md
```diff
@@ -90,7 +90,7 @@ GitHub issues, but instead follow
 guidelines](https://github.com/firecracker-microvm/firecracker/blob/master/SECURITY-POLICY.md).
 
 Other discussion: For general discussion, please join us in the `#containerd`
-channel on the [Firecracker Slack](https://tinyurl.com/firecracker-microvm).
+channel on the [Firecracker Slack](https://join.slack.com/t/firecracker-microvm/shared_invite/zt-oxbm7tqt-GLlze9zZ7sdRSDY6OnXXHg).
```
alakesh, line 93: Any reason why we are not using tinyurl anymore?

FFMMM (approved):
🚀 !

README.md
RoyceDavison: Yeah, it can't re-generate the `https://tinyurl.com/firecracker-microvm` with the original url. If we decide to use tinyurl, we would have something like `https://tinyurl.com/[some_random_characters]`. Both Kazu and I agree to replace the tinyurl with the original url. Firecracker team use the original url as well. Besides, the firecracker-go-sdk also uses the original url.

Issue #488, opened 2021-03-08
test namespace directory removal


FFMMM:
Similar to https://github.com/firecracker-microvm/firecracker-containerd/issues/415, we want to make sure that we are removing namespace directories on `StopVM()` correctly.

This issue's scope involves:

* assessing current behavior (resolution could be a no-op)
* identifying any test gaps and bridging that thru an integ or unit test
* making code changes needed to address a potential issue here.

RoyceDavison:
@kzys A quick question regarding [this line of code](https://github.com/firecracker-microvm/firecracker-containerd/blob/435ade47c5462cbf2e7decdbf12a9570ae5b474a/runtime/jailer_integ_test.go#L154), it seems that fcClient.StopVM() has already removed the namespace, otherwise the assertion couldn't be passed. Why do we need extra testings?

kzys:
The problem is we can remove `[namespace]/[vm ID]`, but nobody remove `[namespace]/` directory. We may need to check the remaining VMs in the directory, or change the directory structure to `[namespace][some separator except /][vm ID]`.

RoyceDavison:
> The problem is we can remove `[namespace]/[vm ID]`, but nobody remove `[namespace]/` directory. We may need to check the remaining VMs in the directory, or change the directory structure to `[namespace][some separator except /][vm ID]` rather than `[namespace]/[vm ID]`.

Got it, thanks. But I still a little bit confused. As you mentioned we don't remove `[namespace]/` directory. Then how it is possible that [this line of code](https://github.com/firecracker-microvm/firecracker-containerd/blob/435ade47c5462cbf2e7decdbf12a9570ae5b474a/runtime/jailer_integ_test.go#L154) returns an IsNotExist error? Since this test is passed, so that mean `assert.True(t, os.IsNotExist(err))` doesn't return any error. But as you mentioned above that, we don't remove `[namespace]/ `directory.

kzys:
The code removes `[namespace]/[vm ID]`, not `[namespace]/` itself.

RoyceDavison:
> The code removes `[namespace]/[vm ID]`, not `[namespace]/` itself.

Yeah, thanks. I made a mistake, didn't see vmID following it.

RoyceDavison:
Close the issue as this pr is merged https://github.com/firecracker-microvm/firecracker-containerd/pull/504

PR #504, opened 2021-05-06
Remove namespace diretory

RoyceDavison:
Remove the namespace directory descried in #488 
Change the directory structure to `[namespace#vmID/]`

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/internal/vm/dir.go
+++ b/internal/vm/dir.go
@@ -49,7 +49,7 @@ func shimDir(varRunDir, namespace, vmID string) (Dir, error) {
 		return "", errors.Wrapf(err, "failed evaluating any symlinks in path %q", varRunDir)
 	}
 
-	return Dir(filepath.Join(resolvedVarRunDir, namespace, vmID)), nil
+	return Dir(filepath.Join(resolvedVarRunDir, namespace+"#"+vmID)), nil
 }
 
 // Dir represents the root of a firecracker-containerd VM directory, which
--- a/internal/vm/dir_test.go
+++ b/internal/vm/dir_test.go
@@ -48,9 +48,11 @@ func TestShimDir(t *testing.T) {
 		{name: "id with ?", ns: "test", id: "?", outErr: `invalid vm id: identifier "?" must match`},
 		{name: "id with *", ns: "test", id: "*", outErr: `invalid vm id: identifier "*" must match`},
 		{name: "id with ,", ns: "test", id: ",", outErr: `invalid vm id: identifier "," must match`},
-		{name: "valid", ns: "ns", id: "1", outDir: "ns/1"},
-		{name: "valid with dashes", ns: "test-123", id: "123-456", outDir: "test-123/123-456"},
-		{name: "valid with dots", ns: "test.123", id: "123.456", outDir: "test.123/123.456"},
+		{name: "valid", ns: "ns", id: "1", outDir: "ns#1"},
+		{name: "valid with dashes", ns: "test-123", id: "123-456", outDir: "test-123#123-456"},
+		{name: "valid with dots", ns: "test.123", id: "123.456", outDir: "test.123#123.456"},
+		{name: "ns with aaa", ns: "aaa", id: "bbb-ccc", outDir: "aaa#bbb-ccc"},
+		{name: "ns with aaa-bbb", ns: "aaa-bbb", id: "ccc", outDir: "aaa-bbb#ccc"},
 	}
 
 	for _, tc := range tests {
--- a/runtime/jailer_integ_test.go
+++ b/runtime/jailer_integ_test.go
@@ -138,7 +138,7 @@ func testJailer(t *testing.T, jailerConfig *proto.JailerConfig) {
 	stdout := startAndWaitTask(ctx, t, c)
 	require.Equal("hello\nadditional drive\n", stdout)
 
-	stat, err := os.Stat(filepath.Join(shimBaseDir(), "default", vmID))
+	stat, err := os.Stat(filepath.Join(shimBaseDir(), "default#"+vmID))
 	require.NoError(err)
 	assert.True(t, stat.IsDir())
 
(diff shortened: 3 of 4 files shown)


internal/vm/dir_test.go
```diff
@@ -48,9 +48,9 @@ func TestShimDir(t *testing.T) {
 		{name: "id with ?", ns: "test", id: "?", outErr: `invalid vm id: identifier "?" must match`},
 		{name: "id with *", ns: "test", id: "*", outErr: `invalid vm id: identifier "*" must match`},
 		{name: "id with ,", ns: "test", id: ",", outErr: `invalid vm id: identifier "," must match`},
-		{name: "valid", ns: "ns", id: "1", outDir: "ns/1"},
-		{name: "valid with dashes", ns: "test-123", id: "123-456", outDir: "test-123/123-456"},
-		{name: "valid with dots", ns: "test.123", id: "123.456", outDir: "test.123/123.456"},
+		{name: "valid", ns: "ns", id: "1", outDir: "ns-1/1"},
+		{name: "valid with dashes", ns: "test-123", id: "123-456", outDir: "test-123-123-456/123-456"},
+		{name: "valid with dots", ns: "test.123", id: "123.456", outDir: "test.123-123.456/123.456"},
```
kzys, line 53: It is a bit odd to see the VM ID part is repeated twice. Why do you want to do that?

internal/vm/dir.go
```diff
@@ -49,7 +49,8 @@ func shimDir(varRunDir, namespace, vmID string) (Dir, error) {
 		return "", errors.Wrapf(err, "failed evaluating any symlinks in path %q", varRunDir)
 	}
 
-	return Dir(filepath.Join(resolvedVarRunDir, namespace, vmID)), nil
+	n := namespace + "-" + vmID
+	return Dir(filepath.Join(resolvedVarRunDir, n, vmID)), nil
```
kzys, line 53: Let's say we have (namespace=`aaa`, vmID=`bbb-ccc`) and (namespace=`aaa-bbb`, vmID=`ccc`), does it work?

kzys (changes requested):

internal/vm/dir_test.go
RoyceDavison: To get a unique namespace dir for each VM.

internal/vm/dir.go
RoyceDavison: It pass all unit tests, I just add these two cases into `dir_test.go`. I am not sure if it still works if both of them try to remove the namespace dir, they share the same namespace dir.

internal/vm/dir_test.go
kzys: I meant, why don't you use `[namespace][separator][vmID]` instead of `[namespace][separator][vmID]/[vmID]`?

internal/vm/dir.go
kzys: Yeah. I think it wouldn't work.

internal/vm/dir_test.go
RoyceDavison: Yeah, I can change to `[namespace-vmID]/`. I also have a concern over this design. If we have 100 VMs, then there will be 100 directories for each vm. Does it look reasonable to you? I don't feel confident on this design choice. 
Do you prefer wait all VMs finish and then remove the namespace dir approach?

internal/vm/dir.go
```diff
@@ -49,7 +49,8 @@ func shimDir(varRunDir, namespace, vmID string) (Dir, error) {
 		return "", errors.Wrapf(err, "failed evaluating any symlinks in path %q", varRunDir)
 	}
 
-	return Dir(filepath.Join(resolvedVarRunDir, namespace, vmID)), nil
+	n := namespace + "-" + vmID
+	return Dir(filepath.Join(resolvedVarRunDir, n)), nil
```
FFMMM, line 53: let's add this as a 1liner please

runtime/jailer_integ_test.go
```diff
@@ -141,7 +141,8 @@ func testJailer(t *testing.T, jailerConfig *proto.JailerConfig) {
 	stdout := startAndWaitTask(ctx, t, c)
 	require.Equal("hello\nadditional drive\n", stdout)
 
-	stat, err := os.Stat(filepath.Join(shimBaseDir(), "default", vmID))
+	nsp := "default-" + vmID
```
FFMMM, line 144: here and everywhere else -- is there a `getter` that we could use?

internal/vm/dir_test.go
kzys: What if the namespace itself contains `nsp`? containerd limits the characters clients can use in namespace names. Please you take a look that first.

RoyceDavison: Another format I am considering is `[namespace.vmID]/`

kzys: Hmm, why? Seems a namespace itself can contain `.`.

internal/vm/dir.go
RoyceDavison: Fixed it.

runtime/jailer_integ_test.go
RoyceDavison: I don't think it is necessary to create a getter here. I move the namespace into oneline.

internal/vm/dir_test.go
RoyceDavison: I test it on https://regexr.com/
I think it is ok for namespace to contain `.` but not as its suffix. 
You can have `default`, `def.ault`, but not `default.`

kzys (commented):
Looks good to me. I was initially concerning to use `#` since it is shell's meta character, meaning the beginning of a comment.

However,

```
% cat a.sh
#! /bin/sh
echo hello#world
echo hello #world
% bash a.sh
hello#world
hello
%
```

As long as there is no space before `#`, it won't be treated as a comment.

kzys (approved):

Issue #456, opened 2020-11-02
The usage of hard links limits customers' options


kzys:
firecracker-containerd uses hard links in some places. However making a hard link doesn't work if its destination and its source are in different partitions.

We need to either 1) document all hard link usages or 2) use more less restricting options such as bind-mount.

RoyceDavison:
To clarify this issue, do you mean [this line of code] use hard links(https://github.com/firecracker-microvm/firecracker-containerd/blob/435ade47c5462cbf2e7decdbf12a9570ae5b474a/runtime/runc_jailer.go#L298)?

kzys:
Yes.

RoyceDavison:
I am not sure if it is a good idea to use bind mount, as the implementation shown in PR#508, I think documenting the restriction is the best for now simply cause bind mounting opens so many security issues. It requires at least chmod w/ 0706 for granting permission which basically means everyone can read and write. I tried many different file modes which are smaller than 0706 such as 0700, 0701, none of them work. Without 0706 file mode, PutLogger will complain: Permission Denied. 
```
msg="failed to create VM" error="failed to start the VM: [PUT /logger][400] putLoggerBadRequest  &{FaultMessage:Permission denied (os error 13)}" runtime=aws.firecracker vmID=vm-2
```

kzys:
How about changing the owner of these files, rather than opening the read and write permissions to others?

RoyceDavison:
> How about changing the owner of these files, rather than opening the read and write permissions to others?

[This line of code](https://github.com/firecracker-microvm/firecracker-containerd/blob/07661f68e69a8663c1b6b1accc569bb9b12280a3/runtime/runc_jailer.go#L435) already changed the whole directory ownership. Do you mean change the ownership of the file as well?

kzys:
Yes. In order to read a file, the file itself must be readable from the user.

In this below example, ec2-user/ is owned by ec2-user, but the file ec2-user/foobar is owned by root. So ec2-user cannot read the file.

```
[ec2-user@ip-172-31-25-68 tmp]$ ls -ld ec2-user
drwxrwxr-x 2 ec2-user ec2-user 20 May 17 17:34 ec2-user
[ec2-user@ip-172-31-25-68 tmp]$ ls -ld ec2-user/foobar
-rw--w---- 1 root ec2-user 0 May 17 17:34 ec2-user/foobar
[...]

PR #508, opened 2021-05-11
Remove hard links in runc jailer

RoyceDavison:
To remove hard links for FIFOs(LogFifo and MetricsFifo) in runc jailer, as described in #456 

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/runtime/runc_jailer.go
+++ b/runtime/runc_jailer.go
@@ -146,8 +146,7 @@ func (j *runcJailer) prepareBindMounts(mounts []*proto.FirecrackerDriveMount) er
 	return nil
 }
 
-// JailPath returns the base directory from where the jail binary will be ran
-// from
+// JailPath returns the base directory from where the jail binary will be ran from
 func (j *runcJailer) OCIBundlePath() string {
 	return j.Config.OCIBundlePath
 }
@@ -167,7 +166,7 @@ func (j *runcJailer) JailPath() vm.Dir {
 // jailed values, like SocketPath in the machineConfig.
 func (j *runcJailer) BuildJailedMachine(cfg *config.Config, machineConfig *firecracker.Config, vmID string) ([]firecracker.Opt, error) {
 	handler := j.BuildJailedRootHandler(cfg, machineConfig, vmID)
-	fifoHandler := j.BuildLinkFifoHandler()
+	fifoHandler := j.BuildBindMountFifoHandler()
 
 	var debugSDK bool
 	if level, set := cfg.DebugHelper.GetFirecrackerSDKLogLevel(); set {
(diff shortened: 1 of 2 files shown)


runtime/runc_jailer.go
```diff
@@ -200,6 +199,48 @@ func (j *runcJailer) BuildJailedMachine(cfg *config.Config, machineConfig *firec
...
+}
+
+// BuildBindMountFifoHandler will return a new firecracker.Handler with the function
+// that will allow bind mounting of the fifos making them visible to Firecracker.
+func (j *runcJailer) BuildBindMountFifoHandler() firecracker.Handler {
+	return firecracker.Handler{
+		Name: jailerFifoHandlerName,
+		Fn: func(ctx context.Context, m *firecracker.Machine) error {
+			logFifo, err := j.bindMountFIFO(m.Cfg.LogPath, internal.FirecrackerLogFifoName)
+			m.Cfg.LogFifo = logFifo
```
kzys, line 228: Can you check err before using logFifo? It may not change the behavior, but I'd like to make this code consistent with others.

runtime/runc_jailer.go
```diff
@@ -287,52 +326,6 @@ func (j *runcJailer) BuildJailedRootHandler(cfg *config.Config, machineConfig *f
 			}
 
 			j.logger.Info("Successfully ran jailer handler")
-			j.started = true
```
kzys, line 290: Why do you want to move `j.started = true` from here to the handler above? I'd like to keep this line here, since we may reorder the handlers in future.

runtime/runc_jailer.go
```diff
@@ -447,6 +440,13 @@ func (j *runcJailer) bindMountFileToJail(src, dst string) error {
 	}
 	defer f.Close()
 
+	if strings.Contains(dst, "fifo") {
```
kzys, line 443: I do think this is too hacky. I know we need to chown the fifo files. Are there files that we cannot chown?

kzys (changes requested):

runtime/runc_jailer.go
RoyceDavison: Good catch. Fixed it.

runtime/runc_jailer.go
RoyceDavison: `j.started = true` in the `BuildJailedRootHandler` will block `BuildBindMountFifoHandler`, since it is always the first handler. Once it set up `j.started=true` which means the runcJailer is started, no bind mount is allowed as [this line of code](https://github.com/firecracker-microvm/firecracker-containerd/blob/7e9026aae9099b8e8be7c0a1acd20ea04c5aab93/runtime/runc_jailer.go#L424) is shown. `BuildBindMountFifoHandler` is triggered after `firecracker.CreateLogFilesHandlerName`.

runtime/runc_jailer.go
RoyceDavison: I also don't like the hardcoding, fixed it.

runtime/runc_jailer.go
kzys: Hmm, how does that work? bindMountFileToJail is modifying runc's configuration. It has to be executed before starting runc.

RoyceDavison: If `j.started=true` is placed inside `BuildBindMountFifoHandler`, it will work. I thought the runc started after all the defaultHandlers. So I just moved the line inside `BuildBindMountFifoHandler`. Now I am not quite sure when the runc is started.

PR #512, opened 2021-05-18
Fix BuildKite's race condition on Amazon Linux

kzys:
On our Amazon Linux 2 BuildKite setup, we run all tests without Docker
to make the setup more closer to production. However runc uses `/run/runc`
directory that would cause a race condition if we have multiple builds
in parallel.

This change adds `--root` parameter to runc to avoid the race condition.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.buildkite/setup_al2.sh
+++ b/.buildkite/setup_al2.sh
@@ -53,4 +53,17 @@ cat << EOF > $runtime_config_path
 EOF
 
 cp ./runtime/firecracker-runc-config.json.example $dir/config.json
-cp ./_submodules/runc/runc $bin_path/runc
+
+# runc uses /run/runc directory by default. Since our Amazon Linux 2 tests on
+# BuildKite don't use Docker, sharing the directory across multiple test
+# runs causes a race condition.
+#
+# This wrapper script gives runc /run/runc-$unique_id instead to avoid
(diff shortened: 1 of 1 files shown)


.buildkite/setup_al2.sh
```diff
@@ -53,4 +53,17 @@ cat << EOF > $runtime_config_path
...
+# BuildKite don't use Docker, sharing the directory across multiple test
+# runs causes a race condition.
+#
+# This wrapper script gives runc /run/runc-$unique_id instead to avoid
+# the race condition.
+cp ./_submodules/runc/runc "$bin_path/runc.real"
+cat > "$bin_path/runc" <<EOF
+#! /bin/bash
+set -euo pipefail
+"$bin_path/runc.real" --root "/run/runc-$unique_id" "\$@"
```
alakesh, line 67: Where is the $unique_id coming from?

kzys: https://github.com/firecracker-microvm/firecracker-containerd/blob/main/.buildkite/al2env.sh which this script loads in the beginning.

alakesh: LGTM

alakesh (approved):

PR #513, opened 2021-05-19
Test fixes

kzys:
*Issue #, if available:*

This PR has multiple test-related fixes. See the commit messages for the detail.

*Description of changes:*

### DIFF
--- a/tools/thinpool.sh
+++ b/tools/thinpool.sh
@@ -83,12 +83,12 @@ else
 
     pool_create() {
         echo sudo lvcreate --type thin-pool \
-             --poolmetadatasize 16GiB \
-             --size 1G \
+             --poolmetadatasize 16MiB \
+             --size 1GiB \
              -n "$name" "$FICD_DM_VOLUME_GROUP"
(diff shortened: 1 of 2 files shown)


IRCody (approved):

tools/thinpool.sh
```diff
@@ -83,12 +83,12 @@ else
 
     pool_create() {
         echo sudo lvcreate --type thin-pool \
-             --poolmetadatasize 16GiB \
-             --size 1G \
+             --poolmetadatasize 16MiB \
```
RoyceDavison, line 86: Just curious, why do we want to shrink the amount of pool size? from 16GiB to 16MiB seems a huge change.

kzys: Not really sure the right size, but allocating 16GiB for metadata while allocating 1GB for data seems excessive.

ustiugov: @kzys do you have an intuition what these numbers should be set to in a close-to-production setting? any rules of thumb?

PR #522, opened 2021-06-25
Fix extra block device bind mounting

RoyceDavison:
# General Description  TLDR;
We tried to attach an extra block device containing filesystems but received **file exists** error. This error is only existed when we used runc_jailer (will call it jailer in the following context) + `DriveExposePolicy_BIND`. Not surprisingly, without the jailer, the extra block device was successfully mounted without any issue.

**Root cause is that the same block device being mounted twice. The duplicated mountings cause an error returned.**

Detailed error message is below
```
An error is returned "rpc error: code = Unknown desc = failed to create VM: failed to patch drive mount stub: failed to expose
patched drive contents to jail: file exists". This commit is for fixing it.
```

### Setup
The jailer config was
```
proto.JailerConfig{
    UID:               300001,
    GID:               300001,
    DriveExposePolicy: proto.DriveExposePolicy_BIND,
}
```
The FirecrackerDriveMount definition we used was
```
    driveMounts := []*proto.FirecrackerDriveMount{&proto.FirecrackerDriveMount{
        HostPath: "/dev/nvme5n1",
        VMPath: "/mnt",
        FilesystemType: "ext4",
    }}
```
/dev/nvme5n1 is a block device, aks *block special file*, and has a ext4 filesystem as shown below
```
admin@xxx:~$ sudo ls -al /dev/nvme5n1
brw-rw---- 1 root disk 259, 8 Jul  9 20:40 /dev/nvme5n1

admin@xxx:~$ sudo file -sL /dev/nvme5n1
/dev/nvme5n1: Linux rev 1.0 ext4 filesystem data, UUID=99fae2c1-aaae-4e40-8f90-620b696a0c65 (extents) (64bit) (large files) (huge files)
```

### Stacktrace
Before deep dive into this problem, we want to clarify that the above error is returned only when we used `DriveExposePolicy: proto.DriveExposePolicy_BIND` in the jailer configuration. If we don't use `DriveExposePolicy_BIND`, no error is returned.

In the current implementation, when a new runc_jailer or jailer is created, it will check if the DriveExposePolicy as shown below. or you can click the [LINK](https://github.com/firecracker-microvm/firecracker-containerd/blob/c2323bc71886b3abfefd9afa53244740f70db0b8/runtime/runc_jailer.go#L129) to find the code snippet.
```
	if j.Config.DriveExposePolicy == proto.DriveExposePolicy_BIND {
		err := j.prepareBindMounts(mounts)
		if err != nil {
			return nil, err
		}
	}
```
**prepareBindMounts()** will then call **bindMountFileToJail()** to bind mount all files, in this case, either block special files or regular files. If these files are regular files contain filesystem images, we can bind mount them into VM and ultimately into the container without any problem. **BUT** if they are block devices, or block special files, then the problem comes into play. 

Before you understand why it is problematic, you need to remeber runc_jailer creation is happened at very beginning, before **CreateVM()**, which means **bindMountFileToJail()** mentioned above is called even before **CreateVM()**. Remember, inside runc_jailer creation, we first check DriveExposePolicy, if it is *DriveExposePolicy_BIND*, we call **prepareBindMounts()** which then calls **bindMountFileToJail()**. All of these calls are happened in the jailer creation, at the very first stage, and all of them are happened before **CreateVM()**. The order is crucial because it explains why the **file exists** error is returned.

Inside **CreateVM()**, there is a **createVM()** which is a helper function, and it calls **mountDrives()** at very end. [LINK to the code](https://github.com/firecracker-microvm/firecracker-containerd/blob/c2323bc71886b3abfefd9afa53244740f70db0b8/runtime/service.go#L618). **mountDrives()** will call **PatchAndMount()** which is then call **jail.ExposeFileToJail(Your_HostPath)**.

**jail.ExposeFileToJail(Your_HostPath)** will inspect the given file at *Your_HostPath* and based on the file type, proper handling will occur to ensure that the file is visible in the jail.
- block device or block special file: call **exposeBlockDeviceToJail()** which then call **syscall.Mknod()** to create a block device, as you may notice, we first check DriveExposePoly at **bindMountFileToJail** in the jailer creation, and then called **mountDrives()** at very end again in createVM(). **syscall.Mknod()** will not create a block file if there is a file with the same name already existed. So **file exists** error is returned.
- regular file: call **exposeFileToJail()**

So the overall timestamp looks like:
CreateVM()-->createVM()-->newJailer() which then call newRuncJailer() to create a jailer--> mountDrives()--> PatchAndMount()

Meanwhile,
**newJailer() => Check DriveExposePolicy and then call prepareBindMounts() is possible**

**PatchAndMount => jail.ExposeFileToJail(Your_HostPath) => exposeBlockDeviceToJail() or exposeFileToJail()**

### Conclusion
With the above stack trace, we found out that when we create a VM using **createVM()** and want to attach an extra block device as well, there are two places gonna create a block special file. One place is when runc_jailer is created, **prepareBindMounts()** is called which means the block special file will be bind mounted into the VM. A second place is after the jailer is created, **createVM()** calls **mountDrives()** and eventually a syscall.Mknod will be made to create a block special file. However, if the file is a regular file, calling **prepareBindMounts()** and both **mountDrives()** will not be problematic, seems the regular file can be truncated.

### Description of changes
Given what we found out, a solution will focus on avoiding create a block special file twice.

More specifically, **prepareBindMounts()**  will be only called for regular files. If they are block devices, we don't need to bind mount them at the runc_jailer creation step.

## Q&A and some interesting findings
- If we ultimately call ExposeFileToJail(), what is the responsibility of prepareBindMounts()?
  - See the conversation below.

- Can we simply set permissions if block device already exists, otherwise call mknod?
  - No, the error message is shown below. This solution doesn't work due to the permission issue. It seems syscall.Mknod() has handled the permission while creating the device file.
```
2021/06/22 22:00:43 failed to create a VM: rpc error: code = Unknown desc = failed to create VM: failed to patch drive mount stub: failed to patch drive: [PATCH /drives/{drive_id}][400] patchGuestDriveByIdBadRequest  &{FaultMessage:Error during drive update (patch): device error: Permission denied (os error 13)}
exit status 1
```
- Can we remove **s.mountDrives()** instead? 
  - No. Because **mountDrives()** is needed to mount stubDrives. And each stubDrive contains a block device information that are needed by **fc-ctrd Agent** for future mounting. [More implementation details about stubDrive](https://github.com/firecracker-microvm/firecracker-containerd/blob/7cf94848ae2e587f7469d42037e54607a461ff34/docs/drive-mounts-proposal.md).
- We also tried to use `bb` to create a disk image from a block device, then the VM is created just fine. 
  - This is because `bb` creates a regular file, not block special file.

- What is the difference bw. mount a root drive and attach extra block device?
  - root drive is handled differently. It is exposed to jailer at [BuildJailedRootHandler](https://github.com/firecracker-microvm/firecracker-containerd/blob/f320d3636aee41661eb525b284ce6213f6c7a3d5/runtime/runc_jailer.go#L256). It doesn't rely on **prepareBindMounts()**, neither on **s.mountDrive()**
  
### DIFF
--- a/internal/fsutil.go
+++ b/internal/fsutil.go
@@ -21,6 +21,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"testing"
 
 	"github.com/pkg/errors"
@@ -33,7 +34,7 @@ func CreateFSImg(ctx context.Context, t *testing.T, fsType string, testFiles ...
 	t.Helper()
 
 	switch fsType {
-	case "ext3", "ext4":
+	case "ext4":
 		return createTestExtImg(ctx, t, fsType, testFiles...)
 	default:
 		require.FailNowf(t, "unsupported fs type %q", fsType)
@@ -70,6 +71,48 @@ func createTestExtImg(ctx context.Context, t *testing.T, extName string, testFil
 	output, err := exec.CommandContext(ctx, "mkfs."+extName, "-d", tempdir, imgFile.Name(), "65536").CombinedOutput()
 	require.NoErrorf(t, err, "failed to create ext img, command output:\n %s", string(output))
 	return imgFile.Name()
+}
+
+// BlockDeviceFile represents a block device
+type BlockDeviceFile struct {
+	Subpath string
+	UID     int
+	GID     int
+	Dev     int // major number
+}
+
+// CreateBlockDevice creates a block device, or block special file for testing
+func CreateBlockDevice(ctx context.Context, t *testing.T, fsType string, testFile BlockDeviceFile) string {
+	t.Helper()
+
+	switch fsType {
+	case "ext4":
+		return createTestBlockDevice(ctx, t, fsType, testFile)
+	default:
+		require.FailNowf(t, "unsupported fs type %q", fsType)
+		return ""
+	}
+}
+
+func createTestBlockDevice(ctx context.Context, t *testing.T, extName string, testFile BlockDeviceFile) string {
+	t.Helper()
+
+	deviceFile := filepath.Join("/tmp/blockExample", testFile.Subpath)
+	err := os.MkdirAll(filepath.Dir(deviceFile), 0750)
+	require.NoError(t, err, "failed to mkdir for the block device")
+
+	err = syscall.Mknod(deviceFile, syscall.S_IFBLK, testFile.Dev)
+	require.NoError(t, err, "failed to create a new block device")
+
+	err = os.Chmod(deviceFile, 0600)
+	require.NoError(t, err, "failed to change file mode for the new created block device")
+
+	err = os.Chown(deviceFile, testFile.UID, testFile.GID)
+	require.NoError(t, err, "failed to grant permission to the new created block device")
+
+	output, err := exec.CommandContext(ctx, "mkfs."+extName, "-v", deviceFile).CombinedOutput()
+	require.NoErrorf(t, err, "failed to create ext img, command output:%s \n", string(output))
+	return deviceFile
 }
 
 // MountInfo holds data parsed from a line of /proc/mounts
--- a/runtime/jailer_integ_test.go
+++ b/runtime/jailer_integ_test.go
@@ -35,6 +35,11 @@ import (
 	"github.com/firecracker-microvm/firecracker-containerd/runtime/firecrackeroci"
 )
 
+const (
+	jailerUID = 300001
+	jailerGID = 300001
+)
+
 func TestJailer_Isolated(t *testing.T) {
 	prepareIntegTest(t)
 	t.Run("Without Jailer", func(t *testing.T) {
@@ -44,15 +49,35 @@ func TestJailer_Isolated(t *testing.T) {
 	t.Run("With Jailer", func(t *testing.T) {
 		t.Parallel()
 		testJailer(t, &proto.JailerConfig{
-			UID: 300001,
-			GID: 300001,
+			UID: jailerUID,
+			GID: jailerGID,
 		})
 	})
 	t.Run("With Jailer and bind-mount", func(t *testing.T) {
 		t.Parallel()
 		testJailer(t, &proto.JailerConfig{
-			UID:               300001,
-			GID:               300001,
+			UID:               jailerUID,
+			GID:               jailerGID,
+			DriveExposePolicy: proto.DriveExposePolicy_BIND,
+		})
+	})
+}
+
+func TestAttachBlockDevice_Isolated(t *testing.T) {
+	prepareIntegTest(t)
+	t.Run("Without Jailer", func(t *testing.T) {
+		testAttachBlockDevice(t, nil)
+	})
+	t.Run("With Jailer", func(t *testing.T) {
+		testAttachBlockDevice(t, &proto.JailerConfig{
+			UID: jailerUID,
+			GID: jailerGID,
+		})
+	})
+	t.Run("With Jailer and bind-mount", func(t *testing.T) {
+		testAttachBlockDevice(t, &proto.JailerConfig{
+			UID:               jailerUID,
+			GID:               jailerGID,
 			DriveExposePolicy: proto.DriveExposePolicy_BIND,
 		})
 	})
@@ -169,4 +194,111 @@ func TestJailerCPUSet_Isolated(t *testing.T) {
 		GID:  300000,
 	}
 	testJailer(t, config)
+}
+
+func testAttachBlockDevice(t *testing.T, jailerConfig *proto.JailerConfig) {
+	require := require.New(t)
+	client, err := containerd.New(containerdSockPath, containerd.WithDefaultRuntime(firecrackerRuntime))
+	require.NoError(err, "unable to create client to containerd service at %s, is containerd running?", containerdSockPath)
+	defer client.Close()
+
+	ctx := namespaces.WithNamespace(context.Background(), "default")
+
+	image, err := alpineImage(ctx, client, defaultSnapshotterName)
+	require.NoError(err, "failed to get alpine image")
+
+	fcClient, err := newFCControlClient(containerdSockPath)
+	require.NoError(err)
+
+	vmID := testNameToVMID(t.Name())
+
+	// default BlockDeviceFile setup, will change UID & GID if a jailerConfig is passed
+	deviceFile := internal.BlockDeviceFile{
+		Subpath: "dir/block1",
+		UID:     0,
+		GID:     0,
+		Dev:     259, // 259 is the major number for blkext which is one type of block devices
+	}
+
+	if jailerConfig != nil {
+		// cover "With Jailer" & "With Jailer and bind-mount" test cases
+		deviceFile.UID = int(jailerConfig.UID)
+		deviceFile.GID = int(jailerConfig.GID)
+	}
+	additionalBlockDevice := internal.CreateBlockDevice(ctx, t, "ext4", deviceFile)
+	blockExampleDir := filepath.Dir(additionalBlockDevice)
+	defer os.RemoveAll(filepath.Dir(blockExampleDir)) // clean up
+
+	request := proto.CreateVMRequest{
+		VMID:         vmID,
+		JailerConfig: jailerConfig,
+		DriveMounts: []*proto.FirecrackerDriveMount{
+			{HostPath: additionalBlockDevice, VMPath: "/home/driveMount", FilesystemType: "ext4"},
+		},
+	}
+
+	// If the drive files are bind-mounted, the files must be readable from the jailer's user.
+	if jailerConfig != nil && jailerConfig.DriveExposePolicy == proto.DriveExposePolicy_BIND {
+		f, err := ioutil.TempFile("", fsSafeTestName(t)+"_rootfs")
+		require.NoError(err)
+		defer f.Close()
+
+		dst := f.Name()
+
+		// Copy the root drive before chown, since the file is used by other tests.
+		err = copyFile(defaultRuntimeConfig.RootDrive, dst, 0400)
+		require.NoErrorf(err, "failed to copy a rootfs as %q", dst)
+
+		err = os.Chown(dst, int(jailerConfig.UID), int(jailerConfig.GID))
+		require.NoError(err, "failed to chown %q", dst)
+
+		request.RootDrive = &proto.FirecrackerRootDrive{HostPath: dst}
+
+		// The additional drive file is only used by this test.
+		err = os.Chown(additionalBlockDevice, int(jailerConfig.UID), int(jailerConfig.GID))
+		require.NoError(err, "failed to chown %q", additionalBlockDevice)
+	}
+
+	_, err = fcClient.CreateVM(ctx, &request)
+	require.NoError(err)
+
+	// create a container to test bind mount block device into the container
+	c, err := client.NewContainer(ctx,
+		vmID+"-container",
+		containerd.WithSnapshotter(defaultSnapshotterName),
+		containerd.WithNewSnapshot(vmID+"-snapshot", image),
+		containerd.WithNewSpec(
+			oci.WithProcessArgs(
+				"/bin/sh", "-c", "echo heyhey && cd /mnt/blockDeviceTest",
+			),
+			firecrackeroci.WithVMID(vmID),
+			oci.WithMounts([]specs.Mount{{
+				Source:      "/home/driveMount",
+				Destination: "/mnt/blockDeviceTest",
+				Options:     []string{"bind"},
+			}}),
+		),
+	)
+	require.NoError(err)
+
+	stdout := startAndWaitTask(ctx, t, c)
+	require.Equal("heyhey\n", stdout)
+
+	stat, err := os.Stat(filepath.Join(shimBaseDir(), "default#"+vmID))
+	require.NoError(err)
+	assert.True(t, stat.IsDir())
+
+	err = c.Delete(ctx, containerd.WithSnapshotCleanup)
+	require.NoError(err, "failed to delete a container-block-device")
+
+	_, err = fcClient.StopVM(ctx, &proto.StopVMRequest{VMID: vmID})
+	require.NoError(err)
+
+	_, err = os.Stat(filepath.Join(shimBaseDir(), "default#"+vmID))
+	assert.Error(t, err)
+	assert.True(t, os.IsNotExist(err))
+
+	shimContents, err := ioutil.ReadDir(shimBaseDir())
+	require.NoError(err)
+	assert.Len(t, shimContents, 0)
 }
--- a/runtime/runc_jailer.go
+++ b/runtime/runc_jailer.go
@@ -138,9 +138,17 @@ func newRuncJailer(
 
 func (j *runcJailer) prepareBindMounts(mounts []*proto.FirecrackerDriveMount) error {
 	for _, m := range mounts {
-		err := j.bindMountFileToJail(m.HostPath, filepath.Join(j.RootPath(), m.HostPath))
-		if err != nil {
+		stat := syscall.Stat_t{}
+		if err := syscall.Stat(m.HostPath, &stat); err != nil {
 			return err
+		}
+		// Only bindMount regular file.
+		// To avoid duplicate files, for block device, we will use system call to create device file for it.
+		if stat.Mode&syscall.S_IFMT == syscall.S_IFREG {
+			err := j.bindMountFileToJail(m.HostPath, filepath.Join(j.RootPath(), m.HostPath))
+			if err != nil {
+				return err
+			}
 		}
 	}
 	return nil
--- a/runtime/service_integ_test.go
+++ b/runtime/service_integ_test.go
@@ -1007,7 +1007,7 @@ func TestDriveMount_Isolated(t *testing.T) {
 		},
 		{
 			VMPath:         "/mnt",
-			FilesystemType: "ext3",
+			FilesystemType: "ext4",
 			// don't specify "ro" to validate it's automatically set via "IsWritable: false"
 			VMMountOptions: []string{"relatime"},
 			ContainerPath:  "/bar",


RoyceDavison:
It seems syscall.Mknod() has handled the permission while creating the device file. Obviously, the device file should be created by syscall.Mknod() when firecracker-containerd exposes the block device to jail. A piece of code somewhere takes the Mknod() jobs.

runtime/runc_jailer.go
```diff
@@ -465,6 +465,12 @@ func (j *runcJailer) bindMountFileToJail(src, dst string) error {
 // exposeBlockDeviceToJail will call mknod on the block device to ensure
 // visibility of the device
 func exposeBlockDeviceToJail(dst string, rdev, uid, gid int) error {
+	if _, err := os.Stat(dst); err == nil {
```
kzys, line 468: Why do we have the conflicting file?

kzys (changes requested):
Please add some tests regarding the fix.

runtime/runc_jailer.go
RoyceDavison: I don't know where the file is being created. I plan to work on this later today, to track it down. If you have some ideas what are some places I should look into would be very helpful.

RoyceDavison:
Once the block file is removed, the error can't be reproduced even with a new EBS volume. Even after you remove the committed code as shown below, you still can not reproduce the error. 
```
if _, err := os.Stat(dst); err == nil {
		if err := os.Remove(dst); err != nil {
			return errors.Wrapf(err, "Can not remove %s", dst)
		}
	}
```

runtime/runc_jailer.go
akshat-kmr: I may not be understanding the issue so I have a question:

Is this function creating a file different from the one created by EBS/EC2 when attaching the volume? For example, assume `/dev/nvme5n1` as the device file for an EBS Volume attached to EC2. If we are passing that same `/dev/nvme5n1` to this function, then it will definitely cause a conflict because it has already been created on our behalf when EBS volume was attached.

Am I understanding the workflow correctly?

RoyceDavison: This function will create a device file in a different directory. The dst passed into the function is diffrent from /dev/nvm5n1, more precisely, it will look like 
```
/var/lib/firecracker-containerd/shim-base/default#test-drive-mount/rootfs/dev/nvme4n1
```
default is namespace, test-drive-mount is VMID.

RoyceDavison:
Two possible solutions to fix the issue by removing the duplicate device file:
1. before syscall.Mknod(), if the file is already existed, remove it
2. Remove prepareBindMounts, since this is only used to bind mount for the block devices, imo, we can safely remove it

RoyceDavison:
Based off my test runs, I think option 1 is much better than option 2. Option 2 will break the jailer integration test in many different ways: doesn't have permission or could not find the root drive.

runtime/runc_jailer.go
```diff
@@ -138,10 +138,18 @@ func newRuncJailer(
...
 	for _, m := range mounts {
-		err := j.bindMountFileToJail(m.HostPath, filepath.Join(j.RootPath(), m.HostPath))
-		if err != nil {
+		stat := syscall.Stat_t{}
+		if err := syscall.Stat(m.HostPath, &stat); err != nil {
 			return err
 		}
+		// Only bindMount regular file.
+		// To avoid duplicate files, for block device, we will use system call to create device file for it.
+		if stat.Mode&syscall.S_IFMT == syscall.S_IFREG {
```
kzys, line 147: I'm a bit confused about what we are doing here.

- For regular files, calling bindMoutFileToJail() is fine since ExposeFileToJail() is ultimately doing the same
- For block devices, calling bindMoutFileToJail() is problematic since it would break ExposeFileToJail().

Is it correct understanding? If we ultimately call ExposeFileToJail(), what is the responsibility of prepareBindMounts()? Sorry I might be the one who wrote this function though.

Also, please write some tests around your change.

kzys (changes requested):

runtime/runc_jailer.go
RoyceDavison: I will add more details on this.

RoyceDavison: I was actually thinking to remove prepareBindMount() before. I was not quite confident to remove it given the below commit. For better or worse, I decided to work around. If prepareBindMount() could be possibly used for a feature developing in the future, we'd better to keep it, I just don't have the broad picture where it should go.
https://github.com/firecracker-microvm/firecracker-containerd/commit/666cb5e5a3ecc2a6bf286c56ab791c0a2d3d6201

internal/fsutil.go
```diff
@@ -72,6 +73,44 @@ func createTestExtImg(ctx context.Context, t *testing.T, extName string, testFil
 	return imgFile.Name()
 }
 
+// Devicefile represents a block device
+type Devicefile struct {
```
kzys, line 77: Devicefile -> DeviceFile?

internal/fsutil.go
```diff
@@ -72,6 +73,44 @@ func createTestExtImg(ctx context.Context, t *testing.T, extName string, testFil
...
+	require.NoError(t, err, "failed to create a new block device")
+
+	os.Chmod(deviceFile, 0600)
+	require.NoError(t, err, "failed to change file mode for the new created block device")
+
+	os.Chown(deviceFile, testFile.UID, testFile.GID)
+	require.NoError(t, err, "failed to grant permission to the new created block device")
+
+	_, err = exec.CommandContext(ctx, "mkfs."+extName, deviceFile).CombinedOutput()
+	require.NoError(t, err, "Can not mkfs.ext4 for the new block device")
```
kzys, line 109: You cannot call out mkfs.ext4 since extName can be something else. That being said, I think you can remove extName parameter. We only use ext4 right now.

internal/fsutil.go
```diff
@@ -72,6 +73,44 @@ func createTestExtImg(ctx context.Context, t *testing.T, extName string, testFil
...
+	Subpath string
+	UID     int
+	GID     int
+	Dev     int // major number
+}
+
+// CreateBlockDevice creates a block device, or block special file for testing
+func CreateBlockDevice(ctx context.Context, t *testing.T, extName string, testFile Devicefile) string {
+	t.Helper()
+	tempdir, err := ioutil.TempDir("", "")
```
kzys, line 87: The tempdir wouldn't be automatically deleted.

runtime/jailer_integ_test.go
```diff
@@ -74,18 +73,43 @@ func testJailer(t *testing.T, jailerConfig *proto.JailerConfig) {
...
+			Dev:     259, // 259 is the major number for blkext which is a block device
+		})
+	} else {
+		// cover "Without Jailer" test case
+		additionalBlockDevice = internal.CreateBlockDevice(ctx, t, "ext4", internal.Devicefile{
+			Subpath: "dev/sda",
+			UID:     0, // Without jailer, set the block device user & group owners to root
+			GID:     0,
+			Dev:     259, // 259 is the major number for blkext which is one type of block devices
+		})
```
kzys, line 104: How about declarating internalDeviceFile first and setting UID/GID later in `if`?

kzys (changes requested):
Looks almost good to me. Some comments on the tests.

internal/fsutil.go
RoyceDavison: Done.

internal/fsutil.go
RoyceDavison: Fixed.

runtime/jailer_integ_test.go
RoyceDavison: Done.

internal/fsutil.go
RoyceDavison: It contains the device file which will be used to bind mount later, we can't remove it at this point. A tempdir created by [createTestExtImg](https://github.com/firecracker-microvm/firecracker-containerd/blob/c2323bc71886b3abfefd9afa53244740f70db0b8/internal/fsutil.go#L51) didn't remove its tempdir as well.

RoyceDavison: But with the update code, an error was returned, is that why we need to remove tempdie in order to resolve this?
```
/tmp/With_Jailer_and_bind-mount509265757/dev/sda is apparently in use by the system; will not make a filesystem here!
```

kzys: I just don't want to ask developers to remove /tmp directory regularly just because we don't remove files.

Not really sure the error you got. Did you mount the file in somewhere?

internal/fsutil.go
```diff
@@ -72,6 +73,55 @@ func createTestExtImg(ctx context.Context, t *testing.T, extName string, testFil
...
+	require.NoError(t, err, "failed to create a new block device")
+
+	os.Chmod(deviceFile, 0600)
+	require.NoError(t, err, "failed to change file mode for the new created block device")
+
+	os.Chown(deviceFile, testFile.UID, testFile.GID)
+	require.NoError(t, err, "failed to grant permission to the new created block device")
+
+	_, err = exec.CommandContext(ctx, "mkfs."+extName, deviceFile).CombinedOutput()
+	require.NoError(t, err, "Can not mkfs.ext4 for the new block device, error: \n %s", err)
```
kzys, line 121: Regarding the error you currently have, you should take the first result from CommandContext() instead of using `_` to capture its stderr output.

RoyceDavison: Fixed.

internal/fsutil.go
```diff
@@ -72,6 +73,55 @@ func createTestExtImg(ctx context.Context, t *testing.T, extName string, testFil
 	return imgFile.Name()
 }
 
+// DeviceFile represents a block device
```
alakesh, line 76: DeviceFile should be renamed as BlockDeviceFile if it only used for block device.

RoyceDavison: Will fix that. Do you have any idea regarding to the error returned by the integration test?
```
Messages:   	Can not mkfs.ext4 for the new block device: mke2fs 1.43.4 (31-Jan-2017)
 /tmp/208162464/example/hey is apparently in use by the system; will not make a filesystem here!
```

internal/fsutil.go
```diff
@@ -72,6 +73,55 @@ func createTestExtImg(ctx context.Context, t *testing.T, extName string, testFil
...
+	GID     int
+	Dev     int // major number
+}
+
+// CreateBlockDevice creates a block device, or block special file for testing
+func CreateBlockDevice(ctx context.Context, t *testing.T, fsType string, testFile DeviceFile) string {
+	t.Helper()
+
+	switch fsType {
+	case "ext3", "ext4":
```
alakesh, line 89: do we still use/support ext3?

RoyceDavison: @kzys What do you think? Should we change both **CreateFSImg** and **CreateBlockDevice** to only include **ext4**?
I keep `ext3` to be consistent with [CreateFSImg](https://github.com/firecracker-microvm/firecracker-containerd/blob/c2323bc71886b3abfefd9afa53244740f70db0b8/internal/fsutil.go#L32).

runtime/runc_jailer.go
```diff
@@ -138,10 +138,18 @@ func newRuncJailer(
 
 func (j *runcJailer) prepareBindMounts(mounts []*proto.FirecrackerDriveMount) error {
 	for _, m := range mounts {
-		err := j.bindMountFileToJail(m.HostPath, filepath.Join(j.RootPath(), m.HostPath))
-		if err != nil {
+		stat := syscall.Stat_t{}
+		if err := syscall.Stat(m.HostPath, &stat); err != nil {
 			return err
 		}
+		// Only bindMount regular file.
```
alakesh, line 145: we need to be absolutely be sure if this assumption won't break anything.

RoyceDavison: I am pretty sure it won't.

internal/fsutil.go
RoyceDavison: Nope. I didn't mount the file. It is a new file generated every time CreateBlockDevice() is called, given the fact that we are creating the block device file under tempDir.

internal/fsutil.go
kzys: I think getting rid of ext3 is fine.

RoyceDavison: Fixed.

RoyceDavison:
1. Remove "ext3" for both **CreateFSImg** and **CreateBlockDevice**, since it's no longer supported.
2. Remove **mkfs** for **CreateBlockDevice**. In the previous integration test draft, we use 259 as the major number to generate a new block device, which is a blkext and extra block device such as EBS volume using this type. However, we can't **mkfs** for this new block device, given the fact that we don't actually have an extra block device, such as EBS volume attach to our ci/cd instance. In summary, we previously tried to use mkfs for new created blkext type block device without actually having an attached drive/volume, so an error was returned
```
Messages:   	Can not mkfs.ext4 for the new block device: mke2fs 1.43.4 (31-Jan-2017)
 /tmp/208162464/example/hey is apparently in use by the system; will not make a filesystem here!
```

internal/fsutil.go
RoyceDavison: Found the root cause, see comment below.

runtime/runc_jailer.go
alakesh, line 147: Instead of filtering out all non-regular file from bind mount, what about gracefully handling the failure if we encounter the "File exist" error, and move on with the rest of the mount points? The reason being, we are not absolutely sure if we can skip all other devices. I think its safer to just skip the duplicate one and move on to the next one rather than not at all looking at all non-regular files.

Looks like out of these, we are only bind mounting S_IFREG and skipping BLK, CHR, DIR, LNK, FIFO etc. Seems like a problem to me. Correct me if i am wrong. 
Ref: https://www.gnu.org/software/libc/manual/html_node/Testing-File-Type.html

runtime/jailer_integ_test.go
```diff
@@ -74,18 +74,44 @@ func testJailer(t *testing.T, jailerConfig *proto.JailerConfig) {
...
+	if jailerConfig != nil {
+		// cover "With Jailer" & "With Jailer and bind-mount" test cases
+		deviceFile.UID = int(jailerConfig.UID)
+		deviceFile.GID = int(jailerConfig.GID)
+		additionalBlockDevice = internal.CreateBlockDevice(ctx, t, "ext4", deviceFile)
+	} else {
+		// cover "Without Jailer" test case
+		additionalBlockDevice = internal.CreateBlockDevice(ctx, t, "ext4", deviceFile)
+	}
+	defer os.RemoveAll(path.Dir(additionalBlockDevice)) // clean up
```
kzys, line 107: You should use filepath.Dir here.

runtime/jailer_integ_test.go
```diff
@@ -74,18 +74,44 @@ func testJailer(t *testing.T, jailerConfig *proto.JailerConfig) {
...
+
+	var additionalBlockDevice string
+	if jailerConfig != nil {
+		// cover "With Jailer" & "With Jailer and bind-mount" test cases
+		deviceFile.UID = int(jailerConfig.UID)
+		deviceFile.GID = int(jailerConfig.GID)
+		additionalBlockDevice = internal.CreateBlockDevice(ctx, t, "ext4", deviceFile)
+	} else {
+		// cover "Without Jailer" test case
+		additionalBlockDevice = internal.CreateBlockDevice(ctx, t, "ext4", deviceFile)
```
kzys, line 105: You can have internal.CreateBlockDevice outside of the if block.

kzys (changes requested):

alakesh (commented):
The solution looks good to me overall. But there are still not enough data to show that this wont break anything else. One particular thing that worries me is the big hammer approach of rejecting all non-regular files/devices.

runtime/runc_jailer.go
RoyceDavison: FIFO and CHR LNK are all deal in different places. There are special handlers and functions for them as well. The reason why we don't handle the "File exist" error where we encounter it due to (you can read more details on my above discussion)
1. No possible way to workaround without introducing complexity or unnecessary
2. We should remove the blockDevice file at the first place where it being mounted, and here it is.

runtime/jailer_integ_test.go
RoyceDavison: Fixed.

runtime/jailer_integ_test.go
RoyceDavison: Fixed.

aarongoulet (approved):
Looks good to me.

internal/fsutil.go
```diff
@@ -72,6 +73,50 @@ func createTestExtImg(ctx context.Context, t *testing.T, extName string, testFil
...
+	default:
+		require.FailNowf(t, "unsupported fs type %q", fsType)
+		return ""
+	}
+}
+
+func createTestBlockDevice(ctx context.Context, t *testing.T, extName string, testFile BlockDeviceFile) string {
+	t.Helper()
+
+	deviceFile := filepath.Join("/tmp/blockExample", testFile.Subpath)
```
kzys, line 100: Can't you use os.TempDir or ioutil.TempDir?

internal/fsutil.go
```diff
@@ -72,6 +73,50 @@ func createTestExtImg(ctx context.Context, t *testing.T, extName string, testFil
...
+	err := os.MkdirAll(filepath.Dir(deviceFile), 0750)
+	require.NoError(t, err, "failed to mkdir for the block device")
+
+	if _, err := os.Stat(deviceFile); os.IsNotExist(err) {
```
kzys, line 104: You can remove this `if` if you use a different file.

kzys (changes requested):

internal/fsutil.go
RoyceDavison: No. This will introduce problem due to parallel testing.

internal/fsutil.go
RoyceDavison: See above.

kzys (approved):

PR #531, opened 2021-08-03
Make integration tests work in aarch64

kzys:
Makefile and CPU templates handling is not quite ready for aarch64.

*Issue #, if available:*

*Description of changes:*


Makefile
```diff
@@ -32,8 +32,15 @@ FIRECRACKER_CONTAINERD_BUILDER_IMAGE?=golang:1.13-buster
 export FIRECRACKER_CONTAINERD_TEST_IMAGE?=localhost/firecracker-containerd-test
 export GO_CACHE_VOLUME_NAME?=gocache
 
+host_arch=$(shell arch)
+ifeq ($(host_arch),aarch64)
+	kernel_sha256sum="e2d7c3d6cc123de9e6052d1f434ca7b47635a1f630d63f7fcc1b7164958375e4"
+else
+	kernel_sha256sum="bc7e2dbf12cf7038937abf42056f6bcd405d3eef4d27aaa3016f0ba15069ae4b"
```
Kern--, line 39: If we're no longer assuming the architecture, does it make sense to explicitly highlight that this is for the x86_64 kernel?

kzys: It does. How about adding some comments on this section?

Kern--: I was thinking that it might be easier to just define all the sha256sums and select the one needed, e.g.:
```
kernel_sha256sum_aarch64="e2d7c3d6cc123de9e6052d1f434ca7b47635a1f630d63f7fcc1b7164958375e4"
kernel_sha256sum_x86_64="bc7e2dbf12cf7038937abf42056f6bcd405d3eef4d27aaa3016f0ba15069ae4b"
...
echo "$(kernel_sha256sum_$(host_arch)) $@" | sha256sum -c -
```

Then the list of hashes would be the list of supported architectures, although that might still be a bit too implicit as the recipe would probably fail with a download error or something.

Kern-- (approved):

PR #552, opened 2022-01-14
Adjust TestMultipleVMs_Isolated

kzys:
We are constantly breaching 60 seconds timeout. While we should investigate
whether we actually have a performance regression or not, the investigation
needs some time and it shouldn't block other development activities.

*Issue #, if available:*

*Description of changes:*


runtime/service_integ_test.go
```diff
@@ -344,9 +344,11 @@ func TestMultipleVMs_Isolated(t *testing.T) {
 				// which makes the agent resource-hoggy than its production build
 				// So the default VM size (128MB) is too small.
 				MachineCfg: &proto.FirecrackerMachineConfiguration{MemSizeMib: 1024},
-				// Because this test starts multiple VMs in parallel, some of them may not start within
-				// the default timeout (20 seconds).
-				TimeoutSeconds: 60,
+				// CreateVM's default timeout is 20 seconds and it cannot be
+				// disabled.
+				// However because this test starts multiuple VMs in parallel,
```
ginglis13, line 349: nit: multiple

ginglis13 (approved):

aarongoulet (approved):

runtime/service_integ_test.go
kzys: 🤦 fixed in the last revision.

PR #562, opened 2022-02-16
Fix race conditions in TestMultipleVMs_Isolated

kzys:
*Issue #, if available:*

*Description of changes:*


runtime/service_integ_test.go
```diff
@@ -351,71 +355,104 @@ func TestMultipleVMs_Isolated(t *testing.T) {
...
+			}
 
-			var containerWg sync.WaitGroup
+			eg := &errgroup.Group{}
 			for containerID := 0; containerID < int(containerCount); containerID++ {
-				containerWg.Add(1)
-				go func(containerID int) {
-					defer containerWg.Done()
-					testMultipleExecs(
+				containerID := containerID
```
Kern--, line 373: was this leftover from some previous code?

runtime/service_integ_test.go
```diff
@@ -351,71 +355,104 @@ func TestMultipleVMs_Isolated(t *testing.T) {
...
+				}
+				return fmt.Errorf(
+					"failed to create a VM while there are %d Firecracker processes: %w",
+					len(matches),
+					createVMErr,
+				)
+			}
 
-			var containerWg sync.WaitGroup
+			eg := &errgroup.Group{}
```
Kern--, line 371: We have `vmEg` above, does it make sense to call this `containerEg`?

runtime/service_integ_test.go
```diff
@@ -351,71 +355,104 @@ func TestMultipleVMs_Isolated(t *testing.T) {
...
 					)
-				}(containerID)
+				})
 			}
 
 			// verify duplicate CreateVM call fails with right error
 			_, err = fcClient.CreateVM(ctx, &proto.CreateVMRequest{VMID: strconv.Itoa(vmID)})
-			require.Error(t, err, "did not receive expected error for duplicate CreateVM call")
+			if err == nil {
+				return fmt.Errorf("VM ID must be unique (err=%s)", err)
```
Kern--, line 389: If we return here and there are still goroutines running `testMultipleExecs`, those will keep running right? Would it be better to create the error group with a derived, cancellable context and `defer cancel()`?

runtime/service_integ_test.go
kzys: Yeah. Will rename the variable.

runtime/service_integ_test.go
kzys: It is for https://eli.thegreenplace.net/2019/go-internals-capturing-loop-variables-in-closures/, but I think it is unnecessary for this loop. Will remove.

runtime/service_integ_test.go
kzys: Right. Let me add context to the groups.

runtime/service_integ_test.go
kzys: Hmm, seems golangci-lint (scopelint) is not happy about that. It could be a false positive though.

runtime/service_integ_test.go
```diff
@@ -351,71 +355,104 @@ func TestMultipleVMs_Isolated(t *testing.T) {
...
+				}
+				return fmt.Errorf(
+					"failed to create a VM while there are %d Firecracker processes: %w",
+					len(matches),
+					createVMErr,
+				)
+			}
 
-			var containerWg sync.WaitGroup
+			containerEg, _ := errgroup.WithContext(vmEgCtx)
```
Kern--, line 371: Shouldn't we bind the context that's returned so we can use it it `testMultipleExecs` so that it gets cancelled if any fail?

kzys: Right. Let me fix.

runtime/service_integ_test.go
```diff
@@ -307,20 +309,22 @@ func TestMultipleVMs_Isolated(t *testing.T) {
...
 		c := cases[caseTypeNumber]
-		vmWg.Add(1)
-		go func(vmID int, containerCount int32, jailerConfig *proto.JailerConfig) {
-			defer vmWg.Done()
+		f := func() error {
+			containerCount := c.MaxContainers
+			jailerConfig := c.JailerConfig
 
 			tapName := fmt.Sprintf("tap%d", vmID)
 			err := createTapDevice(ctx, tapName)
```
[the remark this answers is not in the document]
kzys: Fixed in the last revision.

Kern-- (approved):

PR #563, opened 2022-02-16
[WIP] Add critest to Makefile and Docker test image

ginglis13:
*Description of changes:*
[WIP] 

Add critest tool and dependencies to Docker test image.
Add rule to Makefile for running critest.

Preliminary step before adding critest to BuildKite pipeline.

*Issue #, if available:*

Related: #88


Makefile
```diff
@@ -256,8 +256,12 @@ TEST_BRIDGED_TAP_BIN?=$(BINPATH)/test-bridged-tap
 $(TEST_BRIDGED_TAP_BIN): $(shell find internal/cmd/test-bridged-tap -name *.go) $(GOMOD) $(GOSUM)
 	go build -o $@ $(CURDIR)/internal/cmd/test-bridged-tap
 
+LOOPBACK_BIN?=$(BINPATH)/loopback
+$(LOOPBACK_BIN): 
+	GOBIN=$(dir $@) GO111MODULE=off go get -u github.com/containernetworking/plugins/plugins/main/loopback
```
kzys, line 261: Is it possible to pin the version of loopback? I believe go get is going to fetch the latest version.

runtime/Makefile
```diff
@@ -141,6 +141,24 @@ perf-test:
 		$(FIRECRACKER_CONTAINERD_TEST_IMAGE):$(DOCKER_IMAGE_TAG) \
 		"go test -timeout 0 -v -count=1 -run \"^$(PERF_TESTNAME)$$\""
 
+.PHONY: critest
+critest:
+	$(CURDIR)/../tools/thinpool.sh reset "$(FICD_DM_POOL)"
+	docker run --rm -it \
+		--privileged \
```
kzys, line 148: Does it work with --network=none?

tools/docker/Dockerfile.integ-test
```diff
@@ -40,6 +45,18 @@ RUN containerd 2>/dev/null & \
...
+ENV VERSION="v1.23.0"
+RUN wget https://github.com/kubernetes-sigs/cri-tools/releases/download/$VERSION/critest-$VERSION-linux-amd64.tar.gz
+RUN tar zxvf critest-$VERSION-linux-amd64.tar.gz -C /tmp/
+RUN install -D -o root -g root -m755 -t /usr/local/bin /tmp/critest
+RUN rm -f critest-$VERSION-linux-amd64.tar.gz
+
+RUN wget https://github.com/kubernetes-sigs/cri-tools/releases/download/$VERSION/crictl-$VERSION-linux-amd64.tar.gz
+RUN tar zxvf crictl-$VERSION-linux-amd64.tar.gz -C /tmp/
+RUN install -D -o root -g root -m755 -t /usr/local/bin /tmp/crictl
+RUN rm -f crictl-$VERSION-linux-amd64.tar.gz
```
kzys, line 58: Let's put them in single RUN. https://hackernoon.com/tips-to-reduce-docker-image-sizes-876095da3b34

runtime/Makefile
ginglis13: as I have right now, it does not work with --network=none because critest need to pull an image for creating Pod sandboxes. Would it be best to pull this sandbox image into FIRECRACKER_CONTAINERD_TEST_IMAGE , then ?

Makefile
ginglis13: Yea, go get will get version @ latest . I was following how we currently get/install other container networking plugins: https://github.com/firecracker-microvm/firecracker-containerd/blob/3c257503cc1fa13b3dc941b823a0627e5c77e88f/Makefile#L235-L249

kzys: Oh okay. I'm fine doing that in a separate PR in this case.

runtime/Makefile
kzys: I see. It would be better to keep that as is. Technically we could pull some images beforehand, but that may make critest less useful, assuming we might break the pulling part someday.

tools/docker/Dockerfile.integ-test
```diff
@@ -40,6 +45,13 @@ RUN containerd 2>/dev/null & \
   ctr --address /run/firecracker-containerd/containerd.sock content fetch docker.io/library/alpine:3.10.1 >/dev/null && \
   ctr --address /run/firecracker-containerd/containerd.sock content fetch docker.io/mlabbe/iperf3:3.6-r0 >/dev/null
 
+# Install critest
+ENV VERSION="v1.23.0"
+RUN wget https://github.com/kubernetes-sigs/cri-tools/releases/download/$VERSION/critest-$VERSION-linux-amd64.tar.gz && \
+  tar zxvf critest-$VERSION-linux-amd64.tar.gz -C /tmp/ && \
+  install -D -o root -g root -m755 -t /usr/local/bin /tmp/critest && \
+  rm -f critest-$VERSION-linux-amd64.tar.gz && \
```
kzys, line 53: Please remove the last `&& \`

runtime/Makefile
ginglis13: Ok yes and just more on that, the sandbox image could be predownloading but networking would still have to be enable as there are tests for pulling images too (e.g.  [It] public image with tag should be pulled and removed)

So for now I think we'll have to enable networking for this test.

kzys (approved):

PR #561, opened 2022-02-15
Upgrade github.com/containernetworking/cni to 1.0.1

kzys:
Signed-off-by: Kazuyoshi Kato <katokazu@amazon.com>

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #564, opened 2022-02-17
Update to firecracker to 1.0 release

henry118:
Update to firecracker 1.0 release


internal/cmd/test-bridged-tap/main.go
```diff
@@ -28,7 +28,7 @@ import (
 
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
-	"github.com/containernetworking/cni/pkg/types/current"
+	current "github.com/containernetworking/cni/pkg/types/040"
```
Kern--, line 31: I think 100 is current.

go.mod
```diff
@@ -3,27 +3,29 @@ module github.com/firecracker-microvm/firecracker-containerd
...
 	github.com/containerd/continuity v0.2.0 // indirect
 	github.com/containerd/fifo v1.0.0
+	github.com/containerd/go-cni v1.1.1 // indirect
 	github.com/containerd/go-runc v1.0.0
 	github.com/containerd/ttrpc v1.1.0
 	github.com/containerd/typeurl v1.0.2
-	github.com/containernetworking/cni v0.8.1
-	github.com/containernetworking/plugins v0.9.1
-	github.com/firecracker-microvm/firecracker-go-sdk v0.22.1-0.20210520223842-abd0815b8bf9
+	github.com/containernetworking/cni v1.0.1
```
Kern--, line 15: @kzys closed a PR to update this in #561, Does that mean this is also blocked on updating containerd?

kzys: It may be. I'm not 100% sure though.

internal/cmd/test-bridged-tap/main.go
henry118: i chose `040` because it won't break the build. `100` requires more code changes. i take a look into it.

henry118: fixed.

go.mod
henry118: Regarding upgrading `containerd` to `1.6.0` release which is only compatible with `golang 1.16` upwards, do we want to do the same? What's the consensus here?

henry118: BTW I've created #569. This will be addressed by a standalone PR.

Kern--:
FYI, I cancelled the buildkite build because it seems to have gotten deadlocked. I'll re-run it when the other succeeds.

runtime/service_integ_test.go
```diff
@@ -1945,8 +1945,8 @@ func TestOOM_Isolated(t *testing.T) {
 		containerd.WithSnapshotter(defaultSnapshotterName),
 		containerd.WithNewSnapshot("snapshot-"+vmID, image),
 		containerd.WithNewSpec(
-			// The container is having 2MB of memory.
-			oci.WithMemoryLimit(2*1024*1024),
+			// The container is having 3MB of memory.
```
kzys, line 1948: Can you cut a GitHub issue? A bit concerning to see we have less memory somehow.

.github/workflows/build.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-18.04', 'ubuntu-20.04']
-        go: ['1.13', '1.14', '1.15', '1.16', '1.17']
+        go: ['1.15', '1.16', '1.17']
```
kzys, line 17: Because of Firecracker Go SDK?

henry118: * [go-openapi/runtime@0.23.0](https://github.com/firecracker-microvm/firecracker-go-sdk/commit/d6a80900f9228dc7a5d12f2cb4f6816d4347e369) does not work with go:1.13
* [golang.org/x/net](https://github.com/firecracker-microvm/firecracker-go-sdk/commit/bfa9fa8caac72864e08618ad2e5a7761826a5550) does not work with go:1.13 & 1.14

runtime/service_integ_test.go
henry118: Sure. so the issue should be cut to firecracker or firecracker-containerd?

kzys (approved):

runtime/service_integ_test.go
henry118: Cut issue #572

PR #576, opened 2022-02-24
Add critest as step to buildkite pipeline.

ginglis13:
*Issue #, if available:*

*Description of changes:*

Add in step to buildkite pipeline for critest. Use fcnet configuration list rather than the example CNI plugin configurations in tools/docker/critest

This step will diff the current known status of failing tests with output in the pipeline such that we can maintain a state of known failures while not failing the pipeline on every new build. The version of ginkgo used by critest does not support output to a format file, so these changes include some scripting to capture output and compare.

Current passing tests related to image management.

Current tests failing with network namespacing issues. Host and VM are not sharing the same network namespace. critest (crictl) creates CNI namespace to be used for pod sandbox. FC-CD will create this namespace as well as a network namespace for the VM itself, but critest cannot find the CNI namespace as it is contained within the VM's network namespace.


runtime/Makefile
```diff
@@ -150,14 +150,18 @@ critest:
...
 		--volume $(GO_CACHE_VOLUME_NAME):/go \
 		--env FICD_DM_VOLUME_GROUP=$(FICD_DM_VOLUME_GROUP) \
 		--env FICD_DM_POOL=$(FICD_DM_POOL) \
 		--env GOPROXY=direct \
 		--env GOSUMDB=off \
+      	--env ACK_GINKGO_DEPRECATIONS=1.16.5 \
 		--workdir="/src/runtime" \
 		$(FIRECRACKER_CONTAINERD_TEST_IMAGE):$(DOCKER_IMAGE_TAG) \
-		"critest -runtime-endpoint unix:///run/firecracker-containerd/containerd.sock"
+		"sleep 1 && critest -runtime-endpoint unix:///run/firecracker-containerd/containerd.sock"
```
kzys, line 164: Why does it need to sleep?

ginglis13: This is to allow fc-cd to be fully up and running before running CRI tests. Before, the first test or two would fail with connection issues due to fc-cd not having started listening for incoming requests.

tools/critest/critest_diff.sh
```diff
@@ -0,0 +1,31 @@
...
+#
+# 	http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+
+FICD_DM_POOL="$FICD_DM_POOL"
```
kzys, line 17: How about taking the input from the stdin or command line parameters? It is a bit odd to see the name of the device mapper pool in this script.

kzys (changes requested):

tools/critest/critest_diff.sh
ginglis13: yes agreed, I think BUILDKITE_BUILD_NUMBER may be more appropriate.

kzys: I'd like to make it an explicit parameter rather than referencing environment variables. This script doesn't have to depend on BuildKite and environmental variables are hard to track since they are essentially global variables.

kzys: In addition to, your PR description is great! Please put that in the commit message as well.

ginglis13: sounds good, I can update the script to take an explicit parameter. In the pipeline, however, should we reference the BuildKite number to avoid concurrent builds overwriting the critest output file?

kzys (approved):

ginglis13:
Hmm looks like build against ubuntu 20.04 Go 1.15 is failing. This is the first I've seen any of these builds in our CI fail - any flakiness there or further investigation needed before merging ?

kzys:
> Hmm looks like build against ubuntu 20.04 Go 1.15 is failing. This is the first I've seen any of these builds in our CI fail - any flakiness there or further investigation needed before merging ?

Maybe cut a ticket for future reference?

kzys (approved):

Issue #567, opened 2022-02-18
Upgrade golangci-lint


kzys:
We've been using v1.21.0, while the latest version is v1.44.2.

PR #590, opened 2022-03-05
chore: upgrade golangci-lint to v1.44.2

austinvazquez:
*Issue #, if available:*
#567 

*Description of changes:*
Bump golangci-lint from v1.21.0 to v1.44.2

Removes deprecated linters and adds recommended replacement linters
'revive' and 'exportloopref'.

Marks G306 and G307 as excluded until those errors can be audited
seperately.

Removes an instance of unused variable flagged by 'unused' linter.

Updates few instances of local variable names to be inline with 'revive'
naming standards.

Updates few instances of exported function names to be inline with
'revive' naming standards.

Fixes a few instances where 'append' result was not stored in original
slice.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -13,18 +13,16 @@ linters:
     - ineffassign
     - deadcode
     - typecheck
-    - golint
     - gosec
     - unconvert
     - goconst
     - goimports
     - misspell
     - prealloc
-    - scopelint
     - gocritic
-    - interfacer # Suggests narrower interface types
-    - scopelint # Checks for unpinned variables
     - gofmt
+    - revive
+    - exportloopref
 
 issues:
   exclude-use-default: false
@@ -33,3 +31,5 @@ issues:
     - G103 # Use of unsafe calls should be audited
     - G204 # Subprocess launched with variable
     - G304 # Potential file inclusion via variable
+    - G306 # WriteFile permissions 0600 or less to be audited
+    - G307 # Deferring unsafe method "Close" on type "*os.File" to be audited
--- a/internal/common_test.go
+++ b/internal/common_test.go
@@ -70,7 +70,8 @@ func TestGenerateStubContent(t *testing.T) {
 	stubContent, err := GenerateStubContent(driveID)
 	assert.NoError(t, err)
 
-	expected := append(MagicStubBytes, byte(len(driveID)))
+	expected := append([]byte{}, MagicStubBytes...)
+	expected = append(expected, byte(len(driveID)))
 	expected = append(expected, []byte(driveID)...)
 	assert.Equal(t, string(expected), stubContent)
 }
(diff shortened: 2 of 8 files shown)


.golangci.yml
```diff
@@ -33,3 +31,5 @@ issues:
     - G103 # Use of unsafe calls should be audited
     - G204 # Subprocess launched with variable
     - G304 # Potential file inclusion via variable
+    - G306 # WriteFile permissions 0600 or less to be audited
+    - G307 # Deferring unsafe method "Close" on type "*os.File" to be audited
```
austinvazquez, line 35: I think each these two groups of errors should be handled separately outside this change. Each uncovered numerous issues which were not trivial to resolve or was not surgical like the other fixes in this PR.

runtime/firecrackeroci/vm.go
```diff
@@ -47,11 +47,11 @@ var (
 	}
 )
 
-// WithVmLocalImageConfig configures a spec with the content of the image's config.
+// WithVMLocalImageConfig configures a spec with the content of the image's config.
 // It is similar to containerd's oci.WithImageConfig except that it does not access
 // the image's rootfs on the host. Instead, it configures what it can on the host and
 // passes information to the agent running in the VM to inspect the image's rootfs.
-func WithVmLocalImageConfig(image containerd.Image) oci.SpecOpts {
+func WithVMLocalImageConfig(image containerd.Image) oci.SpecOpts {
```
austinvazquez, line 54: Need eyes here. Exported function name change is not ideal for a chore commit. Should we exclude the 'revive' linter which is flagging this function name? I was unable to gleam from their documentation how to exclude this instance with an inline comment or similar.

Kern--: I think this is fine. This is a new function with a narrow use-case. I don't think there will be any impact to renaming it.

internal/common_test.go
```diff
@@ -70,7 +70,9 @@ func TestGenerateStubContent(t *testing.T) {
 	stubContent, err := GenerateStubContent(driveID)
 	assert.NoError(t, err)
 
-	expected := append(MagicStubBytes, byte(len(driveID)))
+	expected := make([]byte, len(MagicStubBytes))
+	copy(expected, MagicStubBytes)
+	expected = append(expected, byte(len(driveID)))
```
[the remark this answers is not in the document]
austinvazquez: containerd appears to be compliant, although their solution to scenarios like this was much more concise and I will follow their example.
```
expected := append([]byte{}, MagicStubBytes...)
```

runtime/firecrackeroci/vm.go
```diff
@@ -85,25 +85,26 @@ func WithVmLocalImageConfig(image containerd.Image) oci.SpecOpts {
 		}
 		spec.Process.Env = replaceOrAppendEnvValues(defaults, spec.Process.Env)
 		cmd := config.Cmd
-		spec.Process.Args = append(config.Entrypoint, cmd...)
+		config.Entrypoint = append(config.Entrypoint, cmd...)
+		copy(spec.Process.Args, config.Entrypoint)
```
kzys, line 89: It doesn't look "better" than the original.

austinvazquez: Good catch. Not as concise, but probably more correct would be the following:
```go
spec.Process.Args = append([]string{}, config.Entrypoint...)
spec.Process.Args = append(spec.Process.Args, cmd...)
```
@kzys, thoughts?

kzys: That looks better. We shouldn't modify `config.Entrypoint` to just make linters happy. 👍

austinvazquez: Agreed.

kzys (approved):

PR #593, opened 2022-03-08
Allow each test to run for an hour at maximum instead of 10 minutes

kzys:
As like 69d69b1, this commit relaxes the timeout of slow tests.

*Issue #, if available:*

*Description of changes:*


austinvazquez (approved):
I think this is fine, but it might be worth investigating if we can track performance metrics for integration test runtime.

kzys:
> I think this is fine, but it might be worth investigating if we can track performance metrics for integration test runtime.

Yes, but our BuildKite is currently optimized for running multiple tests in parallel. We should bake an AMI or have a dedicated machine for latency-related guarantees.

PR #594, opened 2022-03-16
Add VSockPath to GetVmInfoResponse

Kern--:
Firecracker's virtio-vsock implementation bypasses the the host kernel
by mediating between AF_UNIX on the host and AF_VSOCK in the guest.
Firecracker-containerd places the host AF_UNIX socket path in the
internal shim storage directory when launching a VM. Vsock is currently
only used for communication between the shim and the in-VM agent.

For some use-cases (e.g. remote snapshotters) it would be useful to
have additional services listening on vsock inside the VM with corresponding
host services to communicate with them. This change introduces an API
to allow a host service to get the host AF_UNIX socket path to open
additional connections to vsock services in the guest.

*Issue #, if available:*

*Description of changes:*


austinvazquez (approved):

kzys (commented):
The change looks good to me. Can you add some tests? We should have some tests for GetVMInfo already. So just adding a check there should be enough.

PR #596, opened 2022-03-18
Disable cgroups v2 in our getting started guides

Kern--:
Our getting started guides recommend a pre-built kernel at version 4.14.
Cgroups v2 require kernel 4.15 and our image builder scripts now use
Debian 11 which enabled cgroups v2 by default.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -247,7 +247,7 @@ configuration file has the following fields:
 {
   "firecracker_binary_path": "/usr/local/bin/firecracker",
   "kernel_image_path": "/var/lib/firecracker-containerd/runtime/hello-vmlinux.bin",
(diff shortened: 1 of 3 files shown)


austinvazquez (approved):

kzys:
While I made a similar change in somewhere, shouldn't we upgrade our kernel?

Kern--:
Yes, but also our getting started guides should work.

Issue to track kernel update (or simultaneously testing both 4.14 and 5.10) https://github.com/firecracker-microvm/firecracker-containerd/issues/597

kzys:
Yeah I do agree that the docs should work.

kzys (approved):

PR #598, opened 2022-03-22
refactor: internal HTTP address resolver

austinvazquez:
Requires #594 

*Issue #, if available:*

*Description of changes:*
Change the internal mechanics to query firecracker-containerd
instead of using the requests namespace as the VM ID.

This internal mechanism will be used for integration testing.

### DIFF
--- a/snapshotter/internal/http_address_resolver.go
+++ b/snapshotter/internal/http_address_resolver.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"os/signal"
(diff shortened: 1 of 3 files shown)


snapshotter/internal/http_address_resolver.go
```diff
@@ -132,17 +108,24 @@ func queryAddress(writ http.ResponseWriter, req *http.Request) {
...
-	networkAddress, ok := store[namespace]
-	if !ok {
-		http.Error(writ, "Socket path not found", http.StatusNotFound)
+
+	fcClient, err := client.New(containerdSockPath + ".ttrpc")
+	if err != nil {
+		http.Error(writ, "Internal server error", http.StatusInternalServerError)
+		return
+	}
+
```
Kern--, line 117: nit: `defer fcClient.Close()`

snapshotter/internal/http_address_resolver.go
```diff
@@ -132,17 +108,24 @@ func queryAddress(writ http.ResponseWriter, req *http.Request) {
...
+	vmInfo, err := fcClient.GetVMInfo(context.Background(), &proto.GetVMInfoRequest{VMID: namespace})
+	if err != nil {
+		http.Error(writ, "Internal server error", http.StatusInternalServerError)
```
Kern--, line 120: should we log here for debugging?

snapshotter/internal/http_address_resolver.go
```diff
@@ -56,7 +51,7 @@ func init() {
 //
 // Response:
 // {
-//     "network": "tcp",
+//     "network": "unix",
 //     "address": "192.168.0.1:80"
```
Kern--, line 55: nit: should this address be a path?

austinvazquez: Good call out. I was working to spin up a VM to get an example Vsock path and got sidetracked with the kernel upgrade issue. I'll work on this again and get an update here.

snapshotter/internal/http_address_resolver.go
```diff
@@ -132,17 +108,27 @@ func queryAddress(writ http.ResponseWriter, req *http.Request) {
...
+
+	fcClient, err := client.New(containerdSockPath + ".ttrpc")
+	if err != nil {
+		logger.WithError(err).Error("could not create firecracker client")
+		http.Error(writ, "Internal server error", http.StatusInternalServerError)
+		return
+	}
+	defer fcClient.Close()
+
+	vmInfo, err := fcClient.GetVMInfo(context.Background(), &proto.GetVMInfoRequest{VMID: namespace})
```
ginglis13, line 120: Go HTTP request objects have their own context associated, which defaults to `context.Background()`

https://pkg.go.dev/net/http#Request.Context

should we use that instead?

austinvazquez: Yeah thanks for the catch. I'll add that.

Kern-- (approved):

PR #601, opened 2022-04-04
Add List interface to snapshotter cache

ginglis13:
*Description of changes:*

Add a List interface to Cache and implement it for SnapshotterCache. The
List interface returns keys existing in a cache.

These changes also modify the SnapshotterCache to use a RWMutex rather
than a vanilla Mutex https://pkg.go.dev/sync#RWMutex . This way any
number of readers can read from cache, but only one writer may acquire
the lock at a time.

*Issue #, if available:*

### DIFF
--- a/snapshotter/demux/cache/snapshotter_cache.go
+++ b/snapshotter/demux/cache/snapshotter_cache.go
@@ -25,13 +25,13 @@ import (
 // SnapshotterCache implements a read, write protected cache mechanism
 // for keyed snapshotters.
 type SnapshotterCache struct {
-	mutex        *sync.Mutex
+	mutex        *sync.RWMutex
 	snapshotters map[string]snapshots.Snapshotter
 }
 
 // NewSnapshotterCache creates a new instance with an empty cache.
 func NewSnapshotterCache() *SnapshotterCache {
-	return &SnapshotterCache{&sync.Mutex{}, make(map[string]snapshots.Snapshotter)}
+	return &SnapshotterCache{&sync.RWMutex{}, make(map[string]snapshots.Snapshotter)}
 }
 
 // Get fetches and caches the snapshotter for a given key.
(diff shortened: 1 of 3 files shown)


snapshotter/demux/cache/snapshotter_cache.go
```diff
@@ -25,40 +25,41 @@ import (
...
-
-		snapshotter, ok = cache.snapshotters[key]
-		if !ok {
-			newSnapshotter, err := fetch(ctx, key)
-			if err != nil {
-				return nil, err
-			}
-
-			cache.snapshotters[key] = newSnapshotter
-			snapshotter = newSnapshotter
```
austinvazquez, line 53: I think we want to keep the double check. There is still the chance we leak resources if two threads have a cache miss and end up populating the cache.

austinvazquez (changes requested):

snapshotter/demux/cache/snapshotter_cache.go
ginglis13: Ahh I remember now: https://github.com/firecracker-microvm/firecracker-containerd/pull/588#discussion_r822192190 . I'll revert to include the double check 🙂

austinvazquez (approved):
LGTM

snapshotter/demux/cache/snapshotter_cache.go
```diff
@@ -25,18 +25,20 @@ import (
...
 
 // NewSnapshotterCache creates a new instance with an empty cache.
 func NewSnapshotterCache() *SnapshotterCache {
-	return &SnapshotterCache{&sync.Mutex{}, make(map[string]snapshots.Snapshotter)}
+	return &SnapshotterCache{&sync.RWMutex{}, make(map[string]snapshots.Snapshotter)}
 }
 
 // Get fetches and caches the snapshotter for a given key.
 func (cache *SnapshotterCache) Get(ctx context.Context, key string, fetch SnapshotterProvider) (snapshots.Snapshotter, error) {
+	cache.mutex.RLock()
```
Kern--, line 39: This was specifically setup this way to not need a read lock here. Did something change with this PR that makes it necessary?

ginglis13: Hmm I suppose not. My thinking was the read lock in `Get` would be necessary for read consistency... but given that on a cache miss, a write lock is acquired and used for both reading and potentially writing to the cache, this read lock is not necessary.

Kern-- (approved):

kzys (approved):
Can you re-run the tests to see whether the breakage is transient or not?

Issue #602, opened 2022-04-08
Add functionality for discovering and pulling remote snapshotter metrics


ginglis13:
Remote snapshotters typically expose metrics. We should support pulling these metrics and discovering these metrics endpoints. We can split implementation into these parts:

- [x] Update HTTP resolver to return port for metrics as well as snapshotter API port.
- [x] HTTP proxy for remote snapshotter metrics
    + when a remote snapshotter is initialized, create HTTP proxy server that vsock Dial's, pulls metrics from remote snapshotter's metrics endpoint, and serves them

- [x] Serve list of remote snapshotter proxies for metrics aggregation
    + create endpoint that serves list of remote snapshotter HTTP proxy endpoints for aggregation of metrics from multiple remote snapshotters. 

[`snapshotter/config`](https://github.com/firecracker-microvm/firecracker-containerd/tree/main/snapshotter/config) will be updated to add configuration for both tasks as well as options to enable/disable these metrics collection components.

ginglis13:
closing in favor of #607 's follow up tasks.

PR #605, opened 2022-04-11
Update HTTP resolver to return metrics port

ginglis13:

*Issue #, if available:*

#602 task 1

*Description of changes:*

These changes update the Response of the HTTP resolver to include the
port for pulling metrics from a remote snapshotter. They also include initial update to configuration.

### DIFF
--- a/snapshotter/app/service.go
+++ b/snapshotter/app/service.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"net/url"
 	"os"
 	"os/signal"
 	"strconv"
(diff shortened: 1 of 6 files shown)


austinvazquez (approved):

snapshotter/demux/proxy/address/resolver.go
```diff
@@ -25,6 +25,9 @@ type Response struct {
 	//
 	// Reference: https://pkg.go.dev/net#SplitHostPort
 	Address string `json:"address"`
```
Kern--, line 27: Can we split `SnapshotterPort` out of `Address` now that we have a use-case for multiple guest services using the same host addresss?

ginglis13: Sure thing, that'll make a Response look like:

```
Response{
    Address: /path/to/uds.sock,
    SnapshotterPort: 1234,
    MetricsPort: 5678,
}
```

snapshotter/app/service.go
```diff
@@ -134,17 +135,33 @@ func initSnapshotter(ctx context.Context, config config.Config) (snapshots.Snaps
 		if err != nil {
 			return nil, err
 		}
-		host, portstr, err := net.SplitHostPort(response.Address)
+		u, err := url.Parse(response.Address)
 		if err != nil {
 			return nil, err
 		}
-		port, err := strconv.ParseUint(portstr, base10, bits32)
+		host := u.Hostname()
```
ginglis13, line 142: `net.SplitHostPort` will error if a port is not present in `response.Address` (which is the case per changes to `Resposne`), so opted to use net/url [`Parse`](https://pkg.go.dev/net/url#Parse) for validating address and [`Hostname()`](https://pkg.go.dev/net/url#URL.Hostname) for retrieving.

Kern-- (approved):

PR #606, opened 2022-04-13
feat: add snapshotter metrics proxy and service discovery mechanism

ginglis13:

*Issue #, if available:*

Addresses tasks 2 and 3 (metrics proxy / service discovery) of #602 

*Description of changes:*

Snapshotters handled by the the demux snapshotter may expose metrics
endpoints. These changes modify the newProxySnapshotterFunc to
optionally start a metrics proxy server when a new proxy snapshotter is
initialized. The demux snapshotter's cache has been modified to contain
a map of `RemoteSnapshotter`s, which are simply wrappers around the
containerd snapshotter construct and a new type for metrics proxies.

These changes also introduce a mechanism for discovering these proxied
snapshotter metrics for the purposes of aggregating these metrics.

### DIFF
--- a/snapshotter/app/service.go
+++ b/snapshotter/app/service.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/signal"
 	"strconv"
+	"strings"
 	"syscall"
 
 	snapshotsapi "github.com/containerd/containerd/api/services/snapshots/v1"
@@ -35,6 +36,8 @@ import (
 	"github.com/firecracker-microvm/firecracker-containerd/snapshotter/config"
 	"github.com/firecracker-microvm/firecracker-containerd/snapshotter/demux"
 	"github.com/firecracker-microvm/firecracker-containerd/snapshotter/demux/cache"
+	"github.com/firecracker-microvm/firecracker-containerd/snapshotter/demux/metrics"
+	"github.com/firecracker-microvm/firecracker-containerd/snapshotter/demux/metrics/discovery"
 	"github.com/firecracker-microvm/firecracker-containerd/snapshotter/demux/proxy"
 	proxyaddress "github.com/firecracker-microvm/firecracker-containerd/snapshotter/demux/proxy/address"
 )
@@ -52,7 +55,8 @@ func Run(config config.Config) error {
 
 	group, ctx := errgroup.WithContext(ctx)
 
-	snapshotter, err := initSnapshotter(ctx, config)
+	cache := cache.NewSnapshotterCache()
+	snapshotter, err := initSnapshotter(ctx, config, cache)
 	if err != nil {
 		log.G(ctx).WithFields(
 			logrus.Fields{"resolver": config.Snapshotter.Proxy.Address.Resolver.Type},
@@ -76,6 +80,15 @@ func Run(config config.Config) error {
 		return err
 	}
 
+	var serviceDiscovery *discovery.ServiceDiscovery
+	if config.Snapshotter.Metrics.Enable {
+		sdPort := config.Snapshotter.Metrics.ServiceDiscoveryPort
+		serviceDiscovery = discovery.NewServiceDiscovery(config.Snapshotter.Metrics.Host, sdPort, cache)
+		group.Go(func() error {
+			return serviceDiscovery.Serve()
+		})
+	}
+
 	group.Go(func() error {
 		return grpcServer.Serve(listener)
 	})
@@ -87,6 +100,11 @@ func Run(config config.Config) error {
 
 			if err := snapshotter.Close(); err != nil {
 				log.G(ctx).WithError(err).Error("failed to close snapshotter")
+			}
+			if serviceDiscovery != nil {
+				if err := serviceDiscovery.Shutdown(ctx); err != nil {
+					log.G(ctx).WithError(err).Error("failed to shutdown service discovery server")
+				}
 			}
 		}()
 
@@ -123,13 +141,13 @@ func initResolver(config config.Config) (proxyaddress.Resolver, error) {
 const base10 = 10
 const bits32 = 32
 
-func initSnapshotter(ctx context.Context, config config.Config) (snapshots.Snapshotter, error) {
+func initSnapshotter(ctx context.Context, config config.Config, cache cache.Cache) (snapshots.Snapshotter, error) {
 	resolver, err := initResolver(config)
 	if err != nil {
 		return nil, err
 	}
 
-	newProxySnapshotterFunc := func(ctx context.Context, namespace string) (snapshots.Snapshotter, error) {
+	newProxySnapshotterFunc := func(ctx context.Context, namespace string) (*proxy.RemoteSnapshotter, error) {
 		r := resolver
 		response, err := r.Get(namespace)
 		if err != nil {
@@ -148,22 +166,58 @@ func initSnapshotter(ctx context.Context, config config.Config) (snapshots.Snaps
 			return vsock.DialContext(ctx, host, uint32(port), vsock.WithLogger(log.G(ctx)))
 		}
 
+		var metricsProxy *metrics.Proxy
+		// TODO (ginglis13): port management and lifecycle ties in to overall metrics proxy
+		// server lifecycle. tracked here: https://github.com/firecracker-microvm/firecracker-containerd/issues/607
+		portMap := make(map[int]bool)
 		if config.Snapshotter.Metrics.Enable {
 			metricsPort, err := strconv.ParseUint(response.MetricsPort, base10, bits32)
 			if err != nil {
 				return nil, err
 			}
 
-			// TODO (ginglis13) metricsDialer func to be defined here using metricsPort. It will dial
-			// the same host but connect via its own port. The metrics proxy will be configured in NewProxySnapshotter
-			// task 2 of https://github.com/firecracker-microvm/firecracker-containerd/issues/602
-			_ = func(ctx context.Context, _ string) (net.Conn, error) {
+			metricsDialer := func(ctx context.Context, _, _ string) (net.Conn, error) {
 				return vsock.DialContext(ctx, host, uint32(metricsPort), vsock.WithLogger(log.G(ctx)))
 			}
+
+			portRange := config.Snapshotter.Metrics.PortRange
+			metricsHost := config.Snapshotter.Metrics.Host
+
+			// Assign a port for metrics proxy server.
+			ports := strings.Split(portRange, "-")
+			portRangeError := fmt.Errorf("invalid port range %s", portRange)
+			if len(ports) < 2 {
+				return nil, portRangeError
+			}
+			lower, err := strconv.Atoi(ports[0])
+			if err != nil {
+				return nil, portRangeError
+			}
+			upper, err := strconv.Atoi(ports[1])
+			if err != nil {
+				return nil, portRangeError
+			}
+			port := -1
+			for p := lower; p <= upper; p++ {
+				if _, ok := portMap[p]; !ok {
+					port = p
+					portMap[p] = true
+					break
+				}
+			}
+			if port < 0 {
+				return nil, fmt.Errorf("invalid port: %d", port)
+			}
+
+			metricsProxy, err = metrics.NewProxy(metricsHost, port, response.Labels, metricsDialer)
+			if err != nil {
+				return nil, err
+			}
+
 		}
 
-		return proxy.NewProxySnapshotter(ctx, host, snapshotterDialer)
+		return proxy.NewProxySnapshotter(ctx, host, snapshotterDialer, metricsProxy)
 	}
 
-	return demux.NewSnapshotter(cache.NewSnapshotterCache(), newProxySnapshotterFunc), nil
+	return demux.NewSnapshotter(cache, newProxySnapshotterFunc), nil
 }
--- a/snapshotter/config/config.go
+++ b/snapshotter/config/config.go
@@ -59,7 +59,10 @@ type debug struct {
 }
 
 type metrics struct {
-	Enable bool `toml:"enable" default:"false"`
+	Enable               bool   `toml:"enable" default:"false"`
+	Host                 string `toml:"host"`
+	PortRange            string `toml:"port_range"`
+	ServiceDiscoveryPort int    `toml:"service_discovery_port"`
 }
 
 // Load parses application configuration from a specified file path.
--- a/snapshotter/demux/metrics/discovery/service_discovery.go
+++ b/snapshotter/demux/metrics/discovery/service_discovery.go
@@ -1,0 +1,105 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package discovery
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strconv"
+
+	"github.com/containerd/containerd/log"
+	"github.com/firecracker-microvm/firecracker-containerd/snapshotter/demux/cache"
+)
+
+// ServiceDiscovery discovers metrics services in the snapshotter cache
+// and serves a []MetricsTarget on an HTTP server.
+type ServiceDiscovery struct {
+	// cache is the shared host-level cache of snapshotters also used by the demux snapshotter.
+	cache cache.Cache
+	// server is the HTTP server that returns a list of targets to pull metrics from and apply labels to those metrics.
+	server *http.Server
+}
+
+// NewServiceDiscovery returns a ServiceDiscovery with configured HTTP server and provided cache.
+func NewServiceDiscovery(host string, port int, c cache.Cache) *ServiceDiscovery {
+	return &ServiceDiscovery{
+		server: &http.Server{
+			Addr: host + ":" + strconv.Itoa(port),
+		},
+		cache: c,
+	}
+}
+
+// Serve starts the HTTP server for receiving service discovery requests.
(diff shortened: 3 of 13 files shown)


snapshotter/demux/metrics/proxy.go
```diff
@@ -0,0 +1,144 @@
...
+
+// pullVMSnapshotterMetrics gets metrics from a remote snapshotter.
+func (mp *Proxy) pullVMSnapshotterMetrics(ctx context.Context) (string, error) {
+	res, err := mp.Client.Get(RequestPath)
+	if err != nil {
+		return "", err
+	}
+
+	defer res.Body.Close()
+	body, err := ioutil.ReadAll(res.Body)
```
Kern--, line 138: We probably want to wrap `res.Body` in `MaxBytesReader` or something.

snapshotter/demux/metrics/proxy.go
```diff
@@ -0,0 +1,144 @@
...
+// metrics is an HTTP handler receiving requests on /metrics for a given VM/snapshotter.
+func (mp *Proxy) metrics(w http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+
+	rawMetrics, err := mp.pullVMSnapshotterMetrics(ctx)
+	if err != nil {
+		log.G(ctx).Errorf("error reading response body: %v\n", err)
+	}
+	// Simply write the metrics string to ResponseWriter
+	fmt.Fprintf(w, "%s", rawMetrics)
```
Kern--, line 127: Hmm. Maybe we should use `io.Copy` to stream the response body to the `ResponseWriter` rather than buffering the full response in memory.

snapshotter/app/service.go
```diff
@@ -129,7 +130,7 @@ func initSnapshotter(ctx context.Context, config config.Config) (snapshots.Snaps
 		return nil, err
 	}
 
-	newProxySnapshotterFunc := func(ctx context.Context, namespace string) (snapshots.Snapshotter, error) {
+	newProxySnapshotterFunc := func(ctx context.Context, namespace string) (*proxy.RemoteSnapshotter, error) {
```
Kern--, line 133: Can you explain your reasoning for this? It *feels* more correct to return the interface to me, but I don't have a strong reason for it.

snapshotter/demux/metrics/proxy.go
ginglis13: hm yes, will have to determine some reasonable limit for size of body then. The alternative would be that some collector of the metrics can be configured with a limit, although I see the benefit in imposing that limit here rather than leaving that to the caller.

snapshotter/demux/metrics/proxy.go
ginglis13: Ah yes good callout, can update.

snapshotter/app/service.go
ginglis13: Sure. My reasoning for this is to couple the proxy snapshotter with its metrics proxy so that their lifetimes can be tracked together, i.e. when a the demux snapshotter invokes a cache miss and needs to use this `newProxySnapshotterFunc`, fetching that new proxy snapshotter will both fetch the snapshotter itself as well as initialize its metrics proxy server. When the demux snapshotter goes to evict a proxy snapshotter, this coupling allows for shutting down the metrisc proxy server as well.

snapshotter/demux/metrics/proxy.go
ginglis13: hmm `MaxBytesReader` is intended to be for incoming request bodies. Since this reader is a response body, I think we can use `io.LimitReader` directly.

snapshotter/app/service.go
Kern--: Maybe I missed it but I didn't see a place outside of the snapshotter where we need to expose the metrics proxy (e.g. shutting down the metrics proxy could be done inside the `RemoteSnapshotter`s close rather than requiring the cache to know about it).

However, on deeper reflection, I think the place we will need it is the eventual http discovery API so I think this is probably a change we need.

ginglis13: Updated to implement the `Close` method of the `snapshots.Snapshotter` interface for `RemoteSnapshotter`. I'll be updating this PR with service discovery implementation so that we can consider those changes along with whether or not the cache should be caching the `RemoteSnapshotter` type or a `snapshots.Snapshotter` type - the latter of which might require some type assertion down the line.

snapshotter/demux/metrics/proxy.go
```diff
@@ -0,0 +1,137 @@
...
+		ln, err := net.Listen("tcp", ":"+strconv.Itoa(port))
+		if err != nil {
+			continue
+		}
+		ln.Close()
+		if err != nil {
+			continue
+		}
+		return port, nil
+	}
```
kzys, line 95: Can we assume that the TCP port in the range is "owned" by ourselves and manage them, instead of actually listening to see whether the port is available or not?

snapshotter/demux/proxy/snapshotter.go
```diff
@@ -28,8 +30,31 @@ import (
 // SnapshotterDialer defines an interface for establishing a network connection.
 type SnapshotterDialer = func(context.Context, *logrus.Entry, string, uint32)
 
+// RemoteSnapshotter embeds a snapshots.Snapshotter and its metrics proxy.
+type RemoteSnapshotter struct {
+	snapshots.Snapshotter
+	MetricsProxy *metrics.Proxy
+}
```
kzys, line 37: Do we need to expose these types? Could they be remoteSnapshotter and metricsProxy?

ginglis13: I can refactor to allow metricsProxy to not be exposed. 

As for exposing RemoteSnapshotter as a type, I think the argument for exposing would be that if this were not exposed, then there would have to be some type assertions in other areas. E.g. In `snapshotter/app/service.go` , `snapshotter/demux/cache/snapshotter_cache.go`, etc. where we would change the signatures of many of these functions to be of type `snapshots.Snapshotter` and assert that the returned valued is a `proxy.remoteSnapshotter` before performing any remote specific operations.

snapshotter/demux/metrics/proxy.go
ginglis13: Sure - so if we own this range, we are basically making the assumption that as the "owner" of the range, no other application will be listening on a port in this owned range outside of remote snapshotters metrics? this would then entail something tracking the state of used/open ports in that range, I imagine.

snapshotter/demux/proxy/snapshotter.go
aarongoulet: Regarding type assertions, you could potentially do something like this:

```
type RemoteSnapshotter interface {
    snapshots.Snapshotter    // Extend the snapshots.Snapshotter interface.
    NewProxySnapshotter( ... )
}

// Private struct.  Implements RemoteSnapshotter and Snapshotter interfaces.
type remoteSnapshotter struct {
	snapshots.Snapshotter
	metricsProxy *metrics.Proxy
}

// NewProxySnapshotter creates a proxy snapshotter using gRPC over vsock connection.
func NewProxySnapshotter( ... ) (RemoteSnapshotter, error) {    // Now returns your implementation of the interface.
     ...
}
```

You'll still need some refactoring to expose whichever aspects of `MetricsProxy` that you're currently using elsewhere, but if you _did_ want to make this private, this'd probably be the best approach.

snapshotter/demux/metrics/discovery/service_discovery.go
```diff
@@ -0,0 +1,103 @@
...
+	cache cache.Cache
+	// server is the HTTP server that returns a list of targets to pull metrics from and apply labels to those metrics.
+	server *http.Server
+}
+
+// NewServiceDiscovery returns a ServiceDiscovery with configured HTTP server and provided cache.
+func NewServiceDiscovery(port int, c cache.Cache) *ServiceDiscovery {
+	return &ServiceDiscovery{
+		server: &http.Server{
+			Addr: ":" + strconv.Itoa(port),
```
Kern--, line 40: We should probably specify a host so it doesn't listen on externally routable addresses.

snapshotter/demux/metrics/discovery/service_discovery.go
```diff
@@ -0,0 +1,103 @@
...
+	Labels  map[string]string `json:"labels"`
+}
+
+// serviceDiscoveryHandler scans the cache for snapshotters and their proxy server information,
+// and builds and returns a JSON response in the format of a Prometheus service discovery endpoint.
+func (sd *ServiceDiscovery) serviceDiscoveryHandler(w http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+
+	namespaces := sd.cache.List()
+	services := make([]MetricsTarget, 0)
```
Kern--, line 78: nit: I usually see the `[]MetricsTarget{}` syntax for empty slices.

snapshotter/demux/metrics/discovery/service_discovery.go
```diff
@@ -0,0 +1,103 @@
...
+	for _, ns := range namespaces {
+		cachedSnapshotter, err := sd.cache.Get(ctx, ns, nil)
+		if err != nil {
+			log.G(ctx).Error("unable to retrieve from snapshotter cache: ", err)
+		}
+
+		// make target "localhost:{PORT}" -> the metrics proxy for given snapshotter
+		target := make([]string, 1)
+		targetString := fmt.Sprintf("localhost:%v", cachedSnapshotter.MetricsProxy.Port)
+		target[0] = targetString
```
Kern--, line 88: nit: `target := []string{targetString}`

snapshotter/demux/metrics/discovery/service_discovery.go
```diff
@@ -0,0 +1,103 @@
...
+		// build list of discovered services
+		mt := MetricsTarget{Targets: target, Labels: cachedSnapshotter.MetricsProxy.Labels}
+		services = append(services, mt)
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+
+	response, err := json.Marshal(services)
+	if err != nil {
+		log.G(ctx).Error("unable to marshal service discovery response: ", err)
```
Kern--, line 99: We should probably write an http error here and early exit, right?

snapshotter/demux/metrics/proxy.go
```diff
@@ -0,0 +1,137 @@
...
+	return 0, fmt.Errorf("could not find open port in range %s", portRange)
+}
+
+// Serve starts the metrics proxy server for a single in-VM snapshotter.
+func (mp *Proxy) Serve(ctx context.Context) error {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/metrics", mp.metrics)
+
+	mp.Server = &http.Server{
+		Addr:    ":" + strconv.Itoa(mp.Port),
```
Kern--, line 106: Again, we should have a host so we don't listen externally.

snapshotter/demux/proxy/snapshotter.go
```diff
@@ -39,5 +64,10 @@ func NewProxySnapshotter(ctx context.Context, address string, dialer func(contex
 	if err != nil {
 		return nil, err
 	}
-	return proxy.NewSnapshotter(snapshotsapi.NewSnapshotsClient(gRPCConn), address), nil
+
+	if metricsProxy != nil {
+		go metricsProxy.Serve(ctx)
```
Kern--, line 69: Possibly for followup PR:

Should we have something smarter than just thowing a goroutine into the wild? (e.g. somewhere that selects over an error channel so we can be notified if the metrics proxy unexpectedly exits?)

snapshotter/demux/proxy/snapshotter.go
kzys: What are the remote snapshotter-specific operations?

Generally speaking, it is better to have fewer, larger packages.

https://dave.cheney.net/practical-go/presentations/qcon-china.html#_consider_fewer_larger_packages

snapshotter/demux/metrics/discovery/service_discovery.go
aarongoulet: Not to further muddy the waters, but according to [this doc](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices) from the Golang wiki about code reviews, the `var slicename []type` can be preferable in cases like this because the slice is an actual `nil` value instead of a non-nil, zero-length slice.  Since you are only `append`ing to the slice and not performing a direct assignment by index it works, because `append()` knows exactly what to do (the case is different with direct index `[n]` assignments).  You can test this behavior of `append()` like so:

```go
func main() {
	var services []string

	target := "test"

	services = append(services, target)

	for _, word := range(services) {
		fmt.Println(word)
	}
}
```

You'll get the expected result of `test`:
```
% ./main
test
```

snapshotter/demux/proxy/snapshotter.go
ginglis13: The remote snapshotter doesn't really have specific operations that aren't already defined by the `snapshots.Snapshotter` type... the main caveat being that on close, the remote snapshotter will also shut down the metrics proxy server.

snapshotter/demux/metrics/discovery/service_discovery.go
aarongoulet: (I should add, this is also a nit and the code is functionally fine as-is, but if you do change it, consider the `var x []y` syntax as well.)

snapshotter/demux/metrics/discovery/service_discovery.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		// make target "localhost:{PORT}" -> the metrics proxy for given snapshotter
+		target := make([]string, 1)
+		targetString := fmt.Sprintf("localhost:%v", cachedSnapshotter.MetricsProxy.Port)
+		target[0] = targetString
+
+		// build list of discovered services
+		mt := MetricsTarget{Targets: target, Labels: cachedSnapshotter.MetricsProxy.Labels}
+		services = append(services, mt)
```
kzys, line 92: How about having MetricsProxyPort() and MetricsProxyLabels() on RemoteSnapshotter instead of exposing the fields directly? Right now anyone can modify these fields freely.

snapshotter/config/config.go
```diff
@@ -59,7 +59,14 @@ type debug struct {
 }
 
 type metrics struct {
+	Enable           bool             `toml:"enable" default:"false"`
+	PortRange        string           `toml:"port_range"`
+	ServiceDiscovery serviceDiscovery `toml:"service_discovery"`
+}
+
+type serviceDiscovery struct {
 	Enable bool `toml:"enable" default:"false"`
```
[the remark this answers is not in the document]
ginglis13: I was thinking of having this be optional in the case that a consumer of fc-cd would wish to write their own service discovery but revisiting that thought, yea it doesn't make much sense to enable one and not the other :) I'll take out the option

snapshotter/demux/proxy/snapshotter.go
ginglis13: yea definitely should manage this goroutine explicitly over a channel... I've created issue #607 to track it and added a TODO comment for follow up PR to this one.

snapshotter/demux/metrics/discovery/service_discovery.go
ginglis13: looks like the linter doesn't like `var x[]y` syntax: 
https://github.com/firecracker-microvm/firecracker-containerd/runs/6103232280?check_suite_focus=true
> Error: snapshotter/demux/metrics/discovery/service_discovery.go:78:2: Consider preallocating `services` (prealloc)
var services []MetricsTarget

`services := []MetricsTarget{}` lints fine. marshalling a nil slice rather than an empty, non-nil slice such as `var slicename []type` to JSON would result in a null value rather than an empty JSON array. See https://apoorvam.github.io/blog/2017/golang-json-marshal-slice-as-empty-array-not-null/

snapshotter/config/config.go
```diff
@@ -59,7 +59,14 @@ type debug struct {
...
 type metrics struct {
-	Enable bool `toml:"enable" default:"false"`
+	Enable           bool             `toml:"enable" default:"false"`
+	Host             string           `toml:"host"`
+	PortRange        string           `toml:"port_range"`
+	ServiceDiscovery serviceDiscovery `toml:"service_discovery"`
+}
+
+type serviceDiscovery struct {
+	Port int `toml:"port"`
```
kzys, line 69: How about having ServiceDiscoveryPort on metrics? Would serviceDiscovery have more fields later?

snapshotter/demux/metrics/discovery/service_discovery.go
```diff
@@ -0,0 +1,106 @@
...
+
+// MetricsTarget represents the underlying JSON structure
+// that a Prometheus HTTP Service Discovery server is
+// expected to return.
+//
+// https://prometheus.io/docs/prometheus/latest/http_sd/
+type MetricsTarget struct {
+	Targets []string          `json:"targets"`
+	Labels  map[string]string `json:"labels"`
+}
```
kzys, line 70: You don't have to expose MetricsTarget.

snapshotter/config/config.go
ginglis13: sure can do, serviceDiscovery may have more fields later but at this time I don't anticipate any.

snapshotter/demux/metrics/proxy.go
ginglis13: updated in recent commit. follow up work required for where this owner should live and "freeing" of port once a proxy is shutdown https://github.com/firecracker-microvm/firecracker-containerd/issues/607

Kern-- (approved):

kzys (approved):

snapshotter/demux/metrics/discovery/service_discovery.go
```diff
@@ -0,0 +1,106 @@
...
+	ctx := req.Context()
+
+	namespaces := sd.cache.List()
+	services := []metricsTarget{}
+	for _, ns := range namespaces {
+		cachedSnapshotter, err := sd.cache.Get(ctx, ns, nil)
+		if err != nil {
+			log.G(ctx).Error("unable to retrieve from snapshotter cache: ", err)
+			w.WriteHeader(500)
+			return
```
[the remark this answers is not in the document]
ginglis13: ah good point, this can leave the response header unmodified and `continue` rather than return. Should consider that there are other snapshotters in the cache. Small enough adjustment that I've included in most recent commit.

PR #610, opened 2022-04-22
Support volumes

kzys:
Since firecracker-containerd puts containers in VMs, mounting the host's
directories is not straightforward.

This commit adds "volume" package that makes an empty ext4 image to pack
and create helper functions around to make the image usable from
containers.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/volume/set.go
+++ b/volume/set.go
@@ -1,0 +1,200 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package volume
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/containerd/containerd/containers"
+	"github.com/containerd/containerd/mount"
+	"github.com/containerd/containerd/oci"
+	"github.com/containerd/continuity/fs"
+	"github.com/firecracker-microvm/firecracker-containerd/proto"
+	"github.com/hashicorp/go-multierror"
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+const (
+	vmVolumePath = "/vmv"
+	fsType       = "ext4"
+)
+
+// Set is a set of volumes.
+type Set struct {
+	volumes map[string]*Volume
+	tempDir string
+}
+
+// NewSet returns a new volume set.
+func NewSet() *Set {
+	return NewSetWithTempDir(os.TempDir())
+}
+
+// NewSetWithTempDir returns a new volume set and creates all temporary files under the tempDir.
+func NewSetWithTempDir(tempDir string) *Set {
+	return &Set{volumes: make(map[string]*Volume), tempDir: tempDir}
+}
+
+// Add a volume to the set.
+func (vs *Set) Add(v *Volume) {
+	vs.volumes[v.name] = v
+}
+
+func (vs *Set) createDiskImage(ctx context.Context, size int64) (path string, retErr error) {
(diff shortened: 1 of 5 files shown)


volume/set.go
```diff
@@ -0,0 +1,151 @@
...
+
+func createDiskImage(ctx context.Context, size int64) (string, error) {
+	f, err := os.CreateTemp("", "")
+	if err != nil {
+		return "", err
+	}
+
+	err = f.Truncate(size)
+	if err != nil {
+		return "", err
```
Kern--, line 59: Should we close the file before returning? Also, should we remove the file in cases where either `Truncate` or `mkfs` fails?

volume/volume.go
```diff
@@ -0,0 +1,34 @@
...
+// Package volume provides volumes like Docker and Amazon ECS.
+// Volumes are specicial directories that could be shared by multiple containers.
+//
+// https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#volumes
+package volume
+
+// Volume is a special directory that could be shared between containers.
+type Volume struct {
+	name string
+	host string
```
Kern--, line 23: Can we rename this `hostPath` to make it more clear that it's not a hostname?

volume/set.go
```diff
@@ -0,0 +1,151 @@
...
+	return &Set{volumes: make(map[string]*Volume)}
+}
+
+// Add a volume to the set.
+func (vs *Set) Add(v *Volume) {
+	vs.volumes[v.name] = v
+}
+
+func createDiskImage(ctx context.Context, size int64) (string, error) {
+	f, err := os.CreateTemp("", "")
```
Kern--, line 52: We should probably allow the client to control where this file is created and allow them to delete it during cleanup steps.

volume/volume.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// New returns an empty volume.
+func New(name string) *Volume {
+	return &Volume{name: name}
+}
+
+// FromHost returns a volume which has the files from the given path.
+func FromHost(name, path string) *Volume {
```
Kern--, line 32: Is there any intention to extend this to support dockerfile-style VOLUMEs where the contents of the volume are populated from the contents inside the container?

kzys: Yeah. I'd like to have that one we have a snapshotter in microVMs :)

volume/set.go
kzys: Will do. I may have that in Set and tie this function to *Set.

volume/volume.go
kzys: Sure.

volume/set.go
kzys: Thanks. Will do.

kzys: Done

volume/volume.go
kzys: But in a separate PR.

volume/set.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// PrepareDriveMount returns a FirecrackerDriveMount that could be used with CreateVM.
+func (vs *Set) PrepareDriveMount(ctx context.Context, size int64) (*proto.FirecrackerDriveMount, error) {
+	path, err := vs.createDiskImage(ctx, size)
+	if err != nil {
+		return nil, err
+	}
+
+	dir, err := os.MkdirTemp("", "")
```
Kern--, line 100: `dir` never gets removed. Also, if `mount.Unmount` fails, the error is lost and the directory can't be unmounted since there's no reference to `dir` anymore. I'd like to see some more robust error handling and some place that the client can either get information to do manual cleanup, or trigger a retry, or connect this to GC.

Kern--:
Overall, I think this looks good. 

Some implications of this design that I wanted to make sure I have right:
1) There should be exactly one `volume.Set` per VM
2) All `volume.Volume`s should be added to the set before calling `Set.PrepareDriveMount`

Kern-- (approved):

volume/set.go
kzys: You don't have to approve a PR if you find critical issues :)

Let me fix...

aarongoulet (approved):
Looks great.  Kudos for extra-friendly, readable code!

volume/set.go
kzys: Okay. The last revision shouldn't have any resource leaks.

It is probably difficult to connect to containerd's GC currently, since these Go code is basically living outside of containerd.

PR #615, opened 2022-04-28
Allow configuration of block_device cache strategy in firecracker

swagatbora90:
Signed-off-by: Swagat Bora <sbora@amazon.com>

*Issue #, if available:*

*Description of changes:*
Firecracker offers the possiblity of choosing the block device caching strategy. Caching strategy affects the path data written from inside the microVM takes to the host persistent storage. More details: https://github.com/swagatbora90/firecracker/blob/main/docs/api_requests/block-caching.md

This change allows clients of firecracker-container to pass `cacheType` attribute while creating a VM and additional drive mounts. Also updated `go.mod` to include latest `firecracker-go-sdk` changes.

*Testing*
1. `make test`
2. `make test-in-docker`
3. `make integ-test`


runtime/drive_handler.go
```diff
@@ -185,15 +184,15 @@ func CreateDriveMountStubs(
...
 		stubFileName := fmt.Sprintf("drivemntstub%d", i)
 		options, err := setReadWriteOptions(driveMount.Options, isWritable)
 		if err != nil {
 			return nil, err
 		}
-
 		stubDrive, err := newStubDrive(
 			filepath.Join(jail.JailPath().RootPath(), stubFileName),
-			jail, isWritable, rateLimiter, logger)
+			jail, isWritable, rateLimiter, &cacheType, logger)
```
aarongoulet, line 195: `cacheType` should be passed as a `string`, not a `*string` pointer.

runtime/drive_handler.go
```diff
@@ -259,6 +259,7 @@ func newStubDrive(
 	jail jailer,
 	isWritable bool,
 	rateLimiter *proto.FirecrackerRateLimiter,
+	cacheType *string,
```
aarongoulet, line 262: `cacheType` should be a `string` type, not a `*string` pointer.

runtime/drive_handler.go
```diff
@@ -316,14 +317,20 @@ func newStubDrive(
 		}
 	}
 
+	driveMount := &proto.FirecrackerDriveMount{
+		IsWritable:  isWritable,
+		RateLimiter: rateLimiter,
+	}
+
+	if cacheType != nil {
+		driveMount.CacheType = *cacheType
```
aarongoulet, line 326: Ditto here.

proto/types.proto
```diff
@@ -153,6 +157,10 @@ message FirecrackerDriveMount {
   // writes to the image from within the guest. Defaults to false, in which case
   // the block device in the VM will be read-only.
   bool IsWritable = 6;
+
+  // (Optional) CacheType specifies the caching strategy for the block device.
+  // The supported caching strategies are: "Unsafe"(default) and "WriteBack".
```
aarongoulet, line 162: The canonical value of `DriveCacheTypeWriteback` is `"Writeback"`, not `"WriteBack"`.  The API may not be case-sensitive, but I haven't tested that theory yet.

aarongoulet (changes requested):

ginglis13:
Looks like buildkite integ test is failing a test due to cache type option missing:

> rpc error: code = Unknown desc = failed to create VM: failed to start the VM: [PUT /drives/{drive_id}][400] putGuestDriveByIdBadRequest  &{FaultMessage:An error occurred when deserializing the json body of a request: unknown variant ``, expected `Unsafe` or `Writeback` at line 1 column 16.}

runtime/drive_handler.go
swagatbora90: The intention is to pass whatever values client sends to Firecracker, and let firecracker handle/validate the actual string. Also, that means FC-CD should not make any assumption on the default cache strategy that Firecracker has.

I am thinking of keeping the existing `newStubDrive()` as is and just adding a new receiver method to enable changing the `driveMount.cacheType` of a stubDrive? Something like this:
```
func (sd *stubDrive) WithCacheType(cacheType string) {
	if cacheType != "" {
		sd.driveMount.CacheType = cacheType
	}
}
```

proto/types.proto
swagatbora90: Good catch. I will change the comment to  have the correct string.

runtime/drive_handler.go
aarongoulet: That's perfect.  If `firecracker-containerd` doesn't even need to be aware of the cache type values and just pass them through, it will be less fragile to upstream changes.

proto/types.proto
swagatbora90: Yeah, the API does not appear to be case-sensitive.
```
--- FAIL: TestDriveMountFails_Isolated (0.06s)
    service_integ_test.go:1242: 
                Error Trace:    service_integ_test.go:1242
                Error:          Received unexpected error:
                                rpc error: code = Unknown desc = failed to create VM: failed to start the VM: [PUT /drives/{drive_id}][400] putGuestDriveByIdBadRequest  &{FaultMessage:An error occurred when deserializing the json body of a request: unknown variant `WriteBack`, expected `Unsafe` or `Writeback` at line 1 column 25.}
                Test:           TestDriveMountFails_Isolated
FAIL
```

runtime/drive_handler.go
```diff
@@ -378,3 +383,10 @@ func (sd stubDrive) PatchAndMount(
 
 	return nil
 }
+
+// CacheType sets the stub drive's cacheType value from the provided DriveMount configs.
+func (sd *stubDrive) WithCacheType(cacheType string) {
```
kzys, line 388: setCacheType?

- WithXxx is mainly used for functional options.
- Seems we don't have to expose this method.

kzys, line 388: setCacheType?

- WithXxx is mainly used for functional options.
- Seems we don't have to expose this method.

swagatbora90: Updated to `setCacheType()`

kzys (approved):

aarongoulet (approved):
Looks good to me! Ship it! :shipit:

Issue #619, opened 2022-05-03
git-validation doesn't work due to CR + LF


kzys:

swagatbora90:
For context, `firecracker-containerd` currently uses [git-validation](https://github.com/vbatts/git-validation) to validate git commit message.

`git-validation` is failing on this particular commit:
```
[ec2-user@ip-172-31-9-211 firecracker-containerd]$ /home/ec2-user/firecracker-microvm/firecracker-containerd/bin/git-validation -run DCO,short-subject -range HEAD~2..HEAD
 * 78381c3 "Allow configuration of block_device cache strategy in firecracker (#615)" ... FAIL
  - FAIL - does not have a valid DCO
 * b88e16e "Merge pull request #604 from ginglis13/upgrade-imgcrypt" ... PASS
 * 601bf1f "Upgrade containerd/1.6.3 (CVE-2022-24778)" ... PASS
1 commits to fix
```

`git-validation` is currently using `git --pretty-format: %b`
```
[ec2-user@ip-172-31-9-211 firecracker-containerd]$ git --no-pager log -1 --pretty=format:%b 78381c382b17ac8c8ef6ffbfae27f65db8b186a7 > /tmp/crlf_check
[ec2-user@ip-172-31-9-211 firecracker-containerd]$ cat -A /tmp/crlf_check
Signed-off-by: Swagat Bora <sbora@amazon.com>^M$
^M$
```

PR #620, opened 2022-05-03
Fix lint

Kern--:
78381c382b17ac8c8ef6ffbfae27f65db8b186a7 has a CRLF in the DCO which
git-validation doesn't like. Update lint target to start after that
commit.

*Issue #, if available:*

*Description of changes:*


swagatbora90 (approved):
Looks good to me. Should we create a separate Issue to track revert this change or adding a proper fix to `git-validation`.

Kern--:
I think we can use #619 to track the follow up.

kzys (approved):
LGTM. Waiting the build...

PR #622, opened 2022-05-05
add monitoring of metrics proxy servers

### LINKED ISSUE #607: metrics proxy: monitor proxy server lifecycle
#606 introduces metrics proxy and service discovery mechanism for remote snapshotter metrics. Per this review https://github.com/firecracker-microvm/firecracker-containerd/pull/606#discussion_r853535348 , we should be monitoring the goroutine responsible for serving the metrics proxy deliberately. 

This issue serves to track for a follow up PR adding that functionality. As part of this monitoring, we need to consider that we own the port range for metrics proxy servers. When a server shuts down, we should "free" up this port (i.e. delete it from some set or other data structure) that tracks open ports within the range.

- [ ] select over error channel for proxy server errors
- [ ] on error or graceful termination/shutdown, indicate the port for that server is no longer in use.

ginglis13:
This change introduces a Monitor type which is made aware of a metrics
proxy's closure via a channel. It contains the logic for port selection
for a new metrics proxy and port freeing for a metrics proxy that has
been shut down.

For maintainability, open to discussions around moving Monitor implementation to its own file in the metrics package.

refactoring done to move metrics proxy server initialization to its own function, as well as removal of `enable` configuration option for metrics. Included some unit tests for the `metrics` HTTPResponseHandler

resolves #607

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/snapshotter/app/service.go
+++ b/snapshotter/app/service.go
@@ -17,7 +17,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"net/url"
 	"os"
 	"os/signal"
 	"strconv"
@@ -56,7 +55,29 @@ func Run(config config.Config) error {
 	group, ctx := errgroup.WithContext(ctx)
 
 	cache := cache.NewSnapshotterCache()
-	snapshotter, err := initSnapshotter(ctx, config, cache)
+
+	var (
+		monitor          *metrics.Monitor
+		serviceDiscovery *discovery.ServiceDiscovery
+	)
+	if config.Snapshotter.Metrics.Enable {
+		sdHost := config.Snapshotter.Metrics.Host
+		sdPort := config.Snapshotter.Metrics.ServiceDiscoveryPort
+		serviceDiscovery := discovery.NewServiceDiscovery(sdHost, sdPort, cache)
+		monitor, err := initMetricsProxyMonitor(config.Snapshotter.Metrics.PortRange)
+		if err != nil {
+			log.G(ctx).WithError(err).Fatal("failed creating metrics proxy monitor")
+			return err
+		}
+		group.Go(func() error {
+			return serviceDiscovery.Serve()
+		})
+		group.Go(func() error {
+			return monitor.Start()
+		})
+	}
+
+	snapshotter, err := initSnapshotter(ctx, config, cache, monitor)
 	if err != nil {
 		log.G(ctx).WithFields(
 			logrus.Fields{"resolver": config.Snapshotter.Proxy.Address.Resolver.Type},
@@ -80,15 +101,6 @@ func Run(config config.Config) error {
 		return err
 	}
 
-	var serviceDiscovery *discovery.ServiceDiscovery
-	if config.Snapshotter.Metrics.Enable {
-		sdPort := config.Snapshotter.Metrics.ServiceDiscoveryPort
-		serviceDiscovery = discovery.NewServiceDiscovery(config.Snapshotter.Metrics.Host, sdPort, cache)
-		group.Go(func() error {
-			return serviceDiscovery.Serve()
-		})
-	}
-
 	group.Go(func() error {
 		return grpcServer.Serve(listener)
 	})
@@ -101,10 +113,13 @@ func Run(config config.Config) error {
 			if err := snapshotter.Close(); err != nil {
 				log.G(ctx).WithError(err).Error("failed to close snapshotter")
 			}
-			if serviceDiscovery != nil {
+			if config.Snapshotter.Metrics.Enable {
 				if err := serviceDiscovery.Shutdown(ctx); err != nil {
 					log.G(ctx).WithError(err).Error("failed to shutdown service discovery server")
 				}
+				// Senders to this channel would panic if it is closed. However snapshotter.Close() will
+				// shutdown all metrics proxies and ensure there are no more senders over the channel.
+				monitor.Stop()
 			}
 		}()
 
@@ -141,23 +156,19 @@ func initResolver(config config.Config) (proxyaddress.Resolver, error) {
 const base10 = 10
 const bits32 = 32
 
-func initSnapshotter(ctx context.Context, config config.Config, cache cache.Cache) (snapshots.Snapshotter, error) {
+func initSnapshotter(ctx context.Context, config config.Config, cache cache.Cache, monitor *metrics.Monitor) (snapshots.Snapshotter, error) {
 	resolver, err := initResolver(config)
 	if err != nil {
 		return nil, err
 	}
 
-	newProxySnapshotterFunc := func(ctx context.Context, namespace string) (*proxy.RemoteSnapshotter, error) {
+	newRemoteSnapshotterFunc := func(ctx context.Context, namespace string) (*proxy.RemoteSnapshotter, error) {
 		r := resolver
 		response, err := r.Get(namespace)
 		if err != nil {
 			return nil, err
 		}
-		u, err := url.Parse(response.Address)
-		if err != nil {
-			return nil, err
-		}
-		host := u.Hostname()
+		host := response.Address
 		port, err := strconv.ParseUint(response.SnapshotterPort, base10, bits32)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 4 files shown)


snapshotter/app/service.go
```diff
@@ -56,7 +54,21 @@ func Run(config config.Config) error {
 	group, ctx := errgroup.WithContext(ctx)
 
 	cache := cache.NewSnapshotterCache()
-	snapshotter, err := initSnapshotter(ctx, config, cache)
+	// Start service discovery and metrics proxy monitoring if enabled.
+	var serviceDiscovery *discovery.ServiceDiscovery
+	var monitor *metrics.Monitor
+	if config.Snapshotter.Metrics.Enable {
```
kzys, line 60: - You could declare these variables inside `if` below.
- Do we need to keep Enable=flase case? My worry is that we wouldn't test the path much and could break silently in future.

snapshotter/demux/metrics/proxy.go
```diff
@@ -30,6 +33,75 @@ const (
...
 )
 
+// Monitor is used for port and lifecycle management of metrics proxies.
+type Monitor struct {
+	// monitorChan indicates a port is free for metrics proxy server.
+	ch chan int
+	// portMap tracks used and unused metrics proxy server ports.
+	portMap map[int]bool
+	// portRange is the configurable range of ports to serve metrics proxies from.
+	portRange string
```
kzys, line 43: How about keeping the begin/end of the range, instead of its string representation. This struct doesn't have to know the string representation we have chosen in the TOML file.

kzys (changes requested):

snapshotter/app/service.go
ginglis13: if I declare inside `if` below wouldn't they be out of scope in paths below? e.g. `snapshotter, err := initSnapshotter(ctx, config, cache, monitor)`

Are you suggesting enabling by default/always enabling? I suppose we don't really *need* `Enabled` config, I just am unsure if consumers of this would always want metrics enabled.

snapshotter/demux/metrics/proxy.go
ginglis13: addressed in recent commit 👍

snapshotter/app/service.go
kzys: Oh sorry. Maybe just moving serviceDiscovery inside?

Yes. Always enabling. There might be customers who'd like to disable, but we don't know that yet.

ginglis13: Hm ok along with these changes, I can update config to remove disable / enable option for metrics, easy to add back in if that is every requested / required.

In doing so, should address both points

snapshotter/app/service.go
```diff
@@ -166,58 +168,51 @@ func initSnapshotter(ctx context.Context, config config.Config, cache cache.Cach
...
 
-		return proxy.NewProxySnapshotter(ctx, host, snapshotterDialer, metricsProxy)
+	metricsHost := config.Snapshotter.Metrics.Host
+	metricsProxy, err = metrics.NewProxy(metricsHost, monitor, labels, metricsDialer)
+	if err != nil {
+		return nil, err
 	}
 
-	return demux.NewSnapshotter(cache, newProxySnapshotterFunc), nil
+	return metricsProxy, nil
```
austinvazquez, line 217: No special error handling here right now. Could be more concise.

```suggestion
return metrics.NewProxy(metricsHost, monitor, labels, metricsDialer)
```

snapshotter/demux/metrics/proxy.go
```diff
@@ -30,10 +32,71 @@ const (
 	MaxMetricsResponseSize = 32768 // 32 KB
 )
 
+// Monitor is used for port and lifecycle management of metrics proxies.
+type Monitor struct {
+	// monitorChan indicates a port is free for metrics proxy server.
```
austinvazquez, line 37: Slightly confusing for me. Perhaps wording to indicate the channel is used for port de-allocation communication or to free monitor resources.

snapshotter/demux/metrics/proxy.go
```diff
@@ -30,10 +32,71 @@ const (
...
+func NewMonitor(startPort, endPort int) *Monitor {
+	return &Monitor{make(chan int), make(map[int]bool), startPort, endPort, &sync.Mutex{}}
+}
+
+// Start receives messages from the proxy server channel indiciating a server has closed or terminated.
+// It marks the metrics proxy server's port as available.
+func (m *Monitor) Start() error {
+	for freePort := range m.ch {
+		m.mu.Lock()
+		delete(m.portMap, freePort)
```
austinvazquez, line 59: So once a port is free'd it is no longer available for metrics proxy allocation?

snapshotter/demux/metrics/proxy.go
```diff
@@ -100,5 +172,10 @@ func (mp *Proxy) metrics(w http.ResponseWriter, req *http.Request) {
 	defer res.Body.Close()
 	response := io.LimitReader(res.Body, MaxMetricsResponseSize)
 
-	io.Copy(w, response)
+	_, err = io.Copy(w, response)
+	if err != nil {
+		log.G(ctx).Errorf("error writing response body: %v\n", err)
+		w.WriteHeader(500)
+		return
+	}
```
austinvazquez, line 180: Will the newline look strange in logs?

```suggestion
    if _, err := io.Copy(w, response); err != nil {
        log.G(ctx).Errorf("error writing response body: %v", err)
        w.WriteHeader(500)
    }
```

snapshotter/demux/metrics/proxy.go
```diff
@@ -30,10 +32,71 @@ const (
...
+			m.portMap[p] = true
+			break
+		}
+	}
+
+	if metricsProxyPort < 0 {
+		return 0, fmt.Errorf("invalid port: %d", metricsProxyPort)
+	}
+
+	return metricsProxyPort, nil
```
austinvazquez, line 88:
```suggestion
    for port, ok := range m.portMap {
        if !ok {
            m.portMap[port] = true
            return port, nil
        }
    }
    
    return 0, fmt.Errorf("unable to allocate a port for metrics proxy server")
```

snapshotter/demux/metrics/proxy.go
```diff
@@ -30,10 +32,71 @@ const (
...
+// Monitor is used for port and lifecycle management of metrics proxies.
+type Monitor struct {
+	// monitorChan indicates a port is free for metrics proxy server.
+	ch chan int
+	// portMap tracks used and unused metrics proxy server ports.
+	portMap map[int]bool
+	// startPort is first port in the configurable range of ports to serve metrics proxies from.
+	startPort int
+	// endPort is last port in the configurable range of ports to serve metrics proxies from.
+	endPort int
```
austinvazquez, line 44: Are these really needed after `portMap` allocation?

snapshotter/demux/metrics/proxy.go
```diff
@@ -30,10 +32,71 @@ const (
...
+	// mu is the lock used for threadsafe reading and writing of the metrics proxy port map.
+	mu *sync.Mutex
+}
+
+// NewMonitor returns a new metrics proxy monitor.
+func NewMonitor(startPort, endPort int) *Monitor {
+	return &Monitor{make(chan int), make(map[int]bool), startPort, endPort, &sync.Mutex{}}
+}
```
austinvazquez, line 52: Might be best to add some port range validation. start < end and start > 0

snapshotter/demux/metrics/proxy.go
ginglis13: No, the port map is really treated as a set. so the "set" is checked in `findOpenPort` to see if the element exists in it. If not, then that means the element (port) is free.

snapshotter/demux/metrics/proxy.go
ginglis13: I elected to iterate through the port range rather than the map elements because we are treating this map as a set. The portMap itself has no concept of the port range. So if we were to take this change and iterate through the map itself, we wouldn't actually be considering the port range.
I'll call that we treat it as a set in a comment

snapshotter/demux/metrics/proxy.go
ginglis13: Gp, I can remove the newline 👍

snapshotter/demux/metrics/proxy.go
ginglis13: I'll add, thanks.

snapshotter/demux/metrics/proxy.go
ginglis13: They are needed for finding an open port within the port range with the current implementation of portMap as a set. I suppose we could truly treat portMap as a map, and initialize it on monitor creation to have all keys from `startPort` to `endPort` in the map populated with false values. With that approach, your suggested change of `range`ing over the map would fit. I'm not necessarily in favor of one over the over, do you have a preference or argument for one approach vs the other?

snapshotter/demux/metrics/proxy.go
ginglis13: I will update the wording here to make it clearer 👍

snapshotter/app/service.go
ginglis13: gp, I'll make this change 👍

snapshotter/demux/metrics/proxy.go
austinvazquez: Yes, but maybe better to fail fast choosing to fail at monitor start vs failing on proxy object initialization. Thoughts?

snapshotter/demux/metrics/proxy.go
austinvazquez: Okay, I think understand your usage better now. That resolves my concern.

snapshotter/app/service.go
ginglis13: Revisiting this, there is a possibility that the in-VM remote snapshotter itself is not configured to emit metrics. Given that's the case, would it make more sense to keep `enable` configuration option so that we aren't attempting to pull metrics that aren't being emitted? @kzys @austinvazquez

Kern-- (approved):

austinvazquez (approved):

snapshotter/demux/metrics/proxy.go
```diff
@@ -30,10 +32,76 @@ const (
 	MaxMetricsResponseSize = 32768 // 32 KB
 )
 
+// Monitor is used for port and lifecycle management of metrics proxies.
+type Monitor struct {
+	// monitorChan is used for communicating when a metrics proxy and its port are no longer in use
+	// so that the Monitor can mark the port as free.
+	ch chan int
```
kzys, line 39: ch or monitorChan?

kzys (changes requested):

snapshotter/demux/metrics/proxy.go
ginglis13: ah thanks 🤦 `monitor.monitorChan` is a bit verbose, updated godoc to `ch`

kzys (approved):

Issue #595, opened 2022-03-16
Add integration test for launching a container using a remote snapshotter inside the microVM.


austinvazquez:
I would like to add an integration test that exercises launching a container using a remote snapshotter from inside the Firecracker microVM.

PR #627, opened 2022-05-12
refactor: integration test common code into internal module

austinvazquez:
*Issue #, if available:*
Prep for #595 

*Description of changes:*
Refactor some of the common integration test setup code into a reusable module for integration tests.

### DIFF
--- a/internal/integtest/config.go
+++ b/internal/integtest/config.go
@@ -1,0 +1,62 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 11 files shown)


internal/testutils/integ.go
```diff
@@ -0,0 +1,90 @@
...
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package testutils
```
Kern--, line 14: Nit:
I know we don't already do this, but `utils` is discouraged: https://github.com/golang/go/wiki/CodeReviewComments#package-names

Maybe it would be better to call this package `integtest` and rename the method to just `Prepare` so in tests, you get `integtest.Prepare()` Thoughts?

austinvazquez: I think that is much better than the direction I was heading. Let me push those changes and we can have a second look. :)

kzys (approved):

PR #629, opened 2022-05-18
add: integration test for remote snapshotter

austinvazquez:
*Issue #, if available:*
#595 

*Description of changes:*
Adds integration test to launch a task in using a remote snapshotter running inside the microVM.

### DIFF
--- a/snapshotter/Makefile
+++ b/snapshotter/Makefile
@@ -12,14 +12,28 @@
 # permissions and limitations under the License.
 
 EXTRAGOARGS?=
+NUMBER_OF_VMS?=
 
 SOURCES := $(shell find . -name '*.go')
 GOMOD := $(shell go env GOMOD)
 GOSUM := $(GOMOD:.mod=.sum)
+DOCKER_IMAGE_TAG?=latest
+GO_CACHE_VOLUME_NAME?=gocache
+FIRECRACKER_CONTAINERD_TEST_IMAGE?=localhost/firecracker-containerd-test
 
 REVISION := $(shell git rev-parse HEAD)
 
-all: demux-snapshotter
+INTEG_TEST_SUFFIX := _Isolated
+INTEG_TESTNAMES=$(shell docker run --rm \
+	--network=none \
+	--volume $(CURDIR)/..:/src \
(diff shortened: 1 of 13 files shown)


runtime/drive_handler.go
```diff
@@ -172,6 +172,11 @@ func (h *StubDriveHandler) Release(
 	return nil
 }
 
+func (h *StubDriveHandler) IsReserved(requestCtx context.Context, id string) bool {
+	_, ok := h.usedDrives[id]
+	return ok
+}
```
austinvazquez, line 178: Possible solution for #630. I would like to split this change and review separately from this PR.

austinvazquez:
Relies on #641 being merged.

snapshotter/Makefile
```diff
@@ -12,33 +12,72 @@
...
+		--init \
+		$(FIRECRACKER_CONTAINERD_TEST_IMAGE):$(DOCKER_IMAGE_TAG) \
+		"go test $(EXTRAGOARGS) -run '^$(subst integ-test-,,$@)$$'"
+
+logs:
+	mkdir logs
 
 clean:
 	- rm -f demux-snapshotter
 	- rm -f http-resolver
```
ginglis13, line 78: `rm -f http-address-resolver`

snapshotter/Makefile
```diff
@@ -12,33 +12,72 @@
...
+	- rm -f http-address-resolver
 
 distclean: clean
+	- rm -rf logs
 
-# Install is a noop here, for now.
-install:
-
-.PHONY: all install test clean distclean
+.PHONY: all demux-snapshotter http-address-resolver clean distclean install test integ-test
```
kzys, line 83: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

> A phony target is one that is not really the name of a file; rather it is just a name for a recipe to be executed when you make an explicit request.

So demux-snapshotter and http-address-resolver shouldn't be listed here.

kzys (changes requested):

snapshotter/Makefile
austinvazquez: Oh no my `make`-foo is slipping.

kzys (approved):

PR #631, opened 2022-05-18
Make volume package usable from non-Firecracker runtimes

kzys:
It isn't fully transparent though.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/runtime/volume_integ_test.go
+++ b/runtime/volume_integ_test.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


austinvazquez (approved):

Kern-- (approved):
I think this is fine for now, but I'd like to see a follow up (maybe just an issue for now) with a deeper dive on if we could make this more transparent. Could we do it with the abstractions we have today? Is there a way to extend the containerd abstractions to allow this?

kzys:
@Kern-- Yeah. containerd 1.7's sandbox might help, but it couldn't take something like fc-cd's DriveMount. Filed #634.

PR #636, opened 2022-05-20
Add volume.FromImage

kzys:
OCI images can have multiple volumes. Orchestrators such as ECS import
the volumes from the images in addition to.

volume.FromImage() will be used to implement such a feature.

https://github.com/opencontainers/image-spec/blob/v1.0.2/config.md#properties
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/bind-mounts.html#bind-mount-examples

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/volume/image.go
+++ b/volume/image.go
@@ -1,0 +1,188 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package volume
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/content"
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/mount"
+	"github.com/containerd/containerd/snapshots"
+	"github.com/containerd/continuity/fs"
+	"github.com/hashicorp/go-multierror"
+	"github.com/opencontainers/image-spec/identity"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+type imageVolumeProvider struct {
+	client      *containerd.Client
+	image       string
+	target      string
+	snapshotter string
+	snapshot    string
+}
+
+// ImageOpt allows setting optional properties of the provider.
+type ImageOpt func(*imageVolumeProvider)
+
+// WithSnapshotter sets the snapshotter to pull images.
+func WithSnapshotter(ss string) ImageOpt {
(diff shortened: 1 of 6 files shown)


volume/image.go
```diff
@@ -0,0 +1,184 @@
...
+
+	root, err := p.mount(ctx, tempDir)
+	if err != nil {
+		return nil, err
+	}
+	p.target = root
+
+	result := make([]*Volume, 0, len(config.Config.Volumes))
+	for path := range config.Config.Volumes {
+		result = append(result, &Volume{hostPath: filepath.Join(root, path), containerPath: path})
```
Kern--, line 147: I'm pretty sure this `filepath.Join` is [CVE-2022-23648](https://github.com/containerd/containerd/security/advisories/GHSA-crp2-qrr5-8pq7) :)

We should make the same change as the CRI plugin and probably also add a test to make sure that image volumes can't read from the host.

kzys: Right. The fix was https://github.com/containerd/containerd/commit/8acbb276477c8353ddb7551b58cb3fdeb0b13caa. Let me check where the fs is coming.

kzys: Fixed in the last revision.

volume/set.go
```diff
@@ -37,24 +38,67 @@ const (
...
+	vs.volumes[fmt.Sprintf("named_%s", v.name)] = v
+}
+
+// AddFrom adds volumes from the given provider.
+func (vs *Set) AddFrom(ctx context.Context, vp Provider) error {
+	if _, exists := vs.providers[vp.Name()]; exists {
+		return fmt.Errorf("failed to add %q: %w", vp.Name(), errdefs.ErrAlreadyExists)
+	}
+
+	volumes, err := vp.CreateVolumesUnder(ctx, vs.tempDir)
```
Kern--, line 85: IIUC, when adding volumes from an image, we mount the image at `$VSTempDir/mount`, right? Can you add volumes from 2 containers to the same VS?

Kern--:
Just trying to understand the API:
```
vs := volume.NewSet(runtime)

// Add volumes from host
vs.Add(volume.FromHost("someVol", "/tmp/somehostpath"))

// Add volumes from an image
provider := volume.FromImage(client, ref, "vfc-snapshot", volume.WithSnapshotter(defaultSnapshotterName))
defer func() {
	err := provider.Delete(ctx)
	require.NoError(t, err)
}()
vs.AddFrom(ctx, provider)

// Actually do the copy
vs.PrepareDirecotry(ctx)
// or vs.PrepareDriveMount(ctx, 10*mib) for firecracker

// When creating a container:
// Get OCI Spec Opts
hostVolMounts := vs.WithMounts([]volume.Mount{{Source: volName, Destination: dir, ReadOnly: false}})
imageVolMounts := vs.FromProvider(ref)

// Create a spec
...
containerd.WithNewSpec(
				...,
				hostVolMounts,
				imageVolMounts,
			),
			...


```

Most of this makes sense to me, `FromProvider`  seems more opaque than the others. Could we rename it to `WithMountsFromProvider` or something? Then there's some straight forwards steps: first `Add*` then `Prepare*` then `With*`.

volume/image.go
Kern--: Don't we need `fs.RootPath` here as well?

volume/set.go
kzys: Ah right. Added one more MkdirTemp

Kern-- (approved):

Issue #630, opened 2022-05-18
Release of unused stub drive results in error.


austinvazquez:
Ran into an issue when starting a container with a remote snapshotter. When the task completed and was tearing down, the stub drive handler errors on release of an unused drive.

Two possible solutions:
1. Implement `IsReserved` to verify a drive is in use before calling release.
2. `Release` on unused stub drive is a no-op.

https://github.com/firecracker-microvm/firecracker-containerd/blob/main/runtime/drive_handler.go#L155

PR #641, opened 2022-05-20
fix: only cleanup allocated stub drives

austinvazquez:
*Issue #, if available:*
#630

*Description of changes:*
Only cleanup stub drives for tasks that have allocated them.

### DIFF
--- a/runtime/service.go
+++ b/runtime/service.go
@@ -143,6 +143,8 @@ type service struct {
 	driveMountStubs          []MountableStubDrive
 	exitAfterAllTasksDeleted bool // exit the VM and shim when all tasks are deleted
 
+	blockDeviceTasks map[string]struct{}
+
 	cleanupErr  error
 	cleanupOnce sync.Once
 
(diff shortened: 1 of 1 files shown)


runtime/service.go
```diff
@@ -1281,7 +1281,7 @@ func (s *service) Delete(requestCtx context.Context, req *taskAPI.DeleteRequest)
 	var result *multierror.Error
 	// Trying to release stub drive for further reuse
 	err = s.containerStubHandler.Release(requestCtx, req.ID, s.driveMountClient, s.machine)
-	if err != nil {
+	if err != nil && !errors.Is(err, ErrDriveNotFound) {
 		result = multierror.Append(result, errors.Wrapf(err, "failed to release stub drive for container: %s", req.ID))
```
kzys, line 1285: Hmm, why does it release drives that are not there?

kzys: And Delete() releases all of them regardless of they are managed or not?

Is it possible to handle that inside the implementation of Release()? (cc @Kern--)

austinvazquez: The problem is on task creation we know the container is mounted inside the VM by inspecting the `rootfs` on the request. On deletion however, all we have is the request ID. Another solution is to implement a stub handler function to check if the drive is in use before attempting to release, but that feels more weak than a no-op.

Kern--: I think we probably want to do the book keeping in the service to know whether the container is drive-backed or inside the VM. It's pretty easy for implicit fallback to either mask bugs later, or make debugging more difficult, or just make it harder to understand the code from an outside perspective.

austinvazquez: Okay, I'll try my hand at that solution and get another revision out.

runtime/service.go
```diff
@@ -143,6 +143,8 @@ type service struct {
 	driveMountStubs          []MountableStubDrive
 	exitAfterAllTasksDeleted bool // exit the VM and shim when all tasks are deleted
 
+	taskCleanupStack map[string][]func(context.Context) error
```
kzys, line 146: Let's keep it simple. We right now don't need an arbitrary function and one task wouldn't need multiple cleanup functions.

kzys (approved):

PR #643, opened 2022-05-24
Fix some inefficiencies in our build process

kzys:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,6 +121,7 @@ distclean: clean
 	docker volume rm -f $(GO_CACHE_VOLUME_NAME)
 	$(call rmi-if-exists,$(FIRECRACKER_CONTAINERD_TEST_IMAGE):$(DOCKER_IMAGE_TAG))
 	$(call rmi-if-exists,localhost/$(PROTO_BUILDER_NAME):$(DOCKER_IMAGE_TAG))
+	$(FIRECRACKER_DIR)/tools/devtool distclean
 
 lint:
 	$(BINPATH)/ltag -t ./.headers -excludes "tools $(SUBMODULES)" -check -v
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -340,7 +340,7 @@ $(FIRECRACKER_BIN): $(FIRECRACKER_DIR)/Cargo.toml
 
 .PHONY: firecracker-clean
 firecracker-clean:
-	- $(FIRECRACKER_DIR)/tools/devtool distclean
+	- $(FIRECRACKER_DIR)/tools/devtool clean
```
swagatbora90, line 343: Surprisingly I don't see any method named `cmd_clean` in firecracker `main` branch https://github.com/firecracker-microvm/firecracker/blob/main/tools/devtool

It only has a `cmd_distclean` that it supports right now.  Do you have a PR for Firecracker as well?

kzys: You are right.

```
_submodules/firecracker/tools/devtool clean

[31m[Firecracker devtool] Unknown command: clean. Please use `_submodules/firecracker/tools/devtool help` for help.(B[m

make[1]: [Makefile:343: firecracker-clean] Error 1 (ignored)
```

Makefile
```diff
@@ -121,6 +121,7 @@ distclean: clean
 	docker volume rm -f $(GO_CACHE_VOLUME_NAME)
 	$(call rmi-if-exists,$(FIRECRACKER_CONTAINERD_TEST_IMAGE):$(DOCKER_IMAGE_TAG))
 	$(call rmi-if-exists,localhost/$(PROTO_BUILDER_NAME):$(DOCKER_IMAGE_TAG))
+	$(FIRECRACKER_DIR)/tools/devtool distclean
```
swagatbora90, line 124: Curious, if we should just call `firecracker-clean` target when someone does `make distclean`?

kzys: `devtool distclean` and `make firecracker-clean`? Sounds good to me.

PR #653, opened 2022-05-26
Fix unsafe map access in snapshotter cache

Kern--:
I have been giving bad advice and suggesting that we don't use read locks in the snapshotter cache. Turns out this is unsafe and can cause runtime panics if we concurrently read and write the map.

https://go.dev/doc/faq#atomic_maps

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/snapshotter/demux/cache/snapshotter_cache.go
+++ b/snapshotter/demux/cache/snapshotter_cache.go
@@ -37,7 +37,9 @@ func NewSnapshotterCache() *SnapshotterCache {
 
 // Get fetches and caches the snapshotter for a given key.
 func (cache *SnapshotterCache) Get(ctx context.Context, key string, fetch SnapshotterProvider) (*proxy.RemoteSnapshotter, error) {
+	cache.mutex.RLock()
 	snapshotter, ok := cache.snapshotters[key]
+	cache.mutex.RUnlock()
 
 	if !ok {
 		cache.mutex.Lock()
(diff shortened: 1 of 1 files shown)


kzys (approved):

ginglis13 (approved):

austinvazquez (approved):

snapshotter/demux/cache/snapshotter_cache.go
```diff
@@ -36,7 +36,9 @@ func NewSnapshotterCache() *SnapshotterCache {
 
 // Get fetches and caches the snapshotter for a given key.
 func (cache *SnapshotterCache) Get(ctx context.Context, key string, fetch SnapshotterProvider) (*proxy.RemoteSnapshotter, error) {
+	cache.mutex.RLock()
 	snapshotter, ok := cache.snapshotters[key]
+	cache.mutex.RUnlock()
 
 	if !ok {
 		cache.mutex.Lock()
```
swagatbora90, line 44: Previously you took the lock only when the first map access failed, so second map access was done with mutex. Now, that you have decided to take the lock regardless, can we can get rid of the second map access in L47? 

So if `!ok` we fetch and add to catch. Also we only need to take and release the lock once.

Kern--: We could take a write lock at the beginning of the method and defer the unlock, but then we'd be requiring serialization of snapshotter lookup across all remote snapshotters which is unnecessary. We only expect a cache miss once when we set up the initial connection which is the only point where we need to block all snapshotter actions with this.

swagatbora90: Aah I see your point now. I didn't realize the you used `RLock` and not `Lock`. Makes sense to me.

swagatbora90 (approved):
LGTM

kzys:
@Kern-- Can you rebase this PR? I'd like to have this in main.

PR #654, opened 2022-05-27
minor demux snapshotter metrics bug fixes

ginglis13:
fix a scoping issues in demux snapshotter issuer. initialize monitor
properly in metrics proxy server. add metrics info to example HTTP
address resolver

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/snapshotter/app/service.go
+++ b/snapshotter/app/service.go
@@ -63,8 +63,9 @@ func Run(config config.Config) error {
 	if config.Snapshotter.Metrics.Enable {
 		sdHost := config.Snapshotter.Metrics.Host
 		sdPort := config.Snapshotter.Metrics.ServiceDiscoveryPort
-		serviceDiscovery := discovery.NewServiceDiscovery(sdHost, sdPort, cache)
-		monitor, err := initMetricsProxyMonitor(config.Snapshotter.Metrics.PortRange)
+		serviceDiscovery = discovery.NewServiceDiscovery(sdHost, sdPort, cache)
+		var err error
+		monitor, err = initMetricsProxyMonitor(config.Snapshotter.Metrics.PortRange)
 		if err != nil {
 			log.G(ctx).WithError(err).Fatal("failed creating metrics proxy monitor")
 			return err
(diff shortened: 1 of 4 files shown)


snapshotter/app/service.go
```diff
@@ -63,17 +63,18 @@ func Run(config config.Config) error {
 	if config.Snapshotter.Metrics.Enable {
 		sdHost := config.Snapshotter.Metrics.Host
 		sdPort := config.Snapshotter.Metrics.ServiceDiscoveryPort
-		serviceDiscovery := discovery.NewServiceDiscovery(sdHost, sdPort, cache)
-		monitor, err := initMetricsProxyMonitor(config.Snapshotter.Metrics.PortRange)
+		serviceDiscovery = discovery.NewServiceDiscovery(sdHost, sdPort, cache)
+		mtr, err := initMetricsProxyMonitor(config.Snapshotter.Metrics.PortRange)
```
Kern--, line 67: What's going on here? Is there some issue with aliasing?

ginglis13: yes, it's issue with aliasing/scoping. `monitor, err := ...`, `serviceDiscovery := ...` would create new objects, but the variables initalized outside this scope:
```
var (
		monitor          *metrics.Monitor
		serviceDiscovery *discovery.ServiceDiscovery
)
```

would not actually be set properly; `monitor` and `serviceDiscovery` would be `nil`

ginglis13: I updated to remove `mtr` to be more consistent / readable.

austinvazquez (approved):

kzys (approved):

snapshotter/app/service.go
```diff
@@ -63,8 +63,9 @@ func Run(config config.Config) error {
 	if config.Snapshotter.Metrics.Enable {
 		sdHost := config.Snapshotter.Metrics.Host
 		sdPort := config.Snapshotter.Metrics.ServiceDiscoveryPort
-		serviceDiscovery := discovery.NewServiceDiscovery(sdHost, sdPort, cache)
-		monitor, err := initMetricsProxyMonitor(config.Snapshotter.Metrics.PortRange)
+		serviceDiscovery = discovery.NewServiceDiscovery(sdHost, sdPort, cache)
```
swagatbora90, line 66: how is this being tested today?

snapshotter/internal/http_address_resolver.go
```diff
@@ -44,6 +45,7 @@ func init() {
 	flag.IntVar(&port, "port", 10001, "service port for address resolver")
 	flag.StringVar(&containerdSockPath, "containerdSocket", "/run/firecracker-containerd/containerd.sock", "filepath to the containerd socket")
 	flag.IntVar(&remotePort, "remotePort", 10000, "the remote port on which the remote snapshotter is listening")
+	flag.IntVar(&metricsRemotePort, "metricsRemotePort", 10001, "the remote port on which the remote snapshotter metrics server is listening")
```
swagatbora90, line 48: I see that `10001` is being used as default for both address resolver port and for metrics remote port? intentional?

snapshotter/app/service.go
ginglis13: This will be tested as part of demux snapshotter/metrics integration tests that @austinvazquez and I are currently working on.

snapshotter/internal/http_address_resolver.go
ginglis13: The address resolver is running on the host while remote ports are used for services listening via vsock in VM, so there will not be a conflict. This internal resolver is used as an example and for integration testing. I can choose a different remote port for metrics since you're right - it does look fishy.

swagatbora90 (approved):

Kern-- (approved):

Issue #652, opened 2022-05-26
Walk Failing on Demux Snapshotter after Garbage Collection


ginglis13:
Walk fails on task exit after garbage collection when using the demux snapshotter with error:

```
time="2022-05-26T17:37:20Z" level=error msg="Function called without namespaced context" error="namespace is required: failed precondition" function=Walk
```

This shows up in demux snapshotter logs; not seeing a related error in firecracker-containerd logs but we should root cause this.

austinvazquez:
@ginglis13, I have opened #678 which addresses garbage collection using in-VM snapshotter. Please have a look and provide feedback.

PR #656, opened 2022-05-31
add: no namespace walk support to demux snapshotter

austinvazquez:

*Issue #, if available:*
#652 

*Description of changes:*
This PR adds the ability for no namespaced walks on the demux snapshotter to be broadcasted to all cached snapshotters.

### DIFF
--- a/snapshotter/demux/snapshotter.go
+++ b/snapshotter/demux/snapshotter.go
@@ -243,7 +243,8 @@ func (s *Snapshotter) Walk(ctx context.Context, fn snapshots.WalkFunc, filters .
 	contextLogger := log.G(ctx).WithField("function", "Walk")
 	namespace, err := getNamespaceFromContext(ctx, contextLogger)
 	if err != nil {
-		return err
+		contextLogger.Debug("no namespace found, proxying walk function to all cached snapshotters")
+		return s.cache.WalkAll(ctx, fn, filters...)
 	}
 	logger := contextLogger.WithField("namespace", namespace)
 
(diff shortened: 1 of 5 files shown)


snapshotter/demux/snapshotter.go
```diff
@@ -243,7 +243,8 @@ func (s *Snapshotter) Walk(ctx context.Context, fn snapshots.WalkFunc, filters .
 	contextLogger := log.G(ctx).WithField("function", "Walk")
 	namespace, err := getNamespaceFromContext(ctx, contextLogger)
 	if err != nil {
-		return err
+		contextLogger.Debug("no namespace found, proxying walk function to all cached snapshotters")
+		return s.cache.WalkAll(ctx, fn, filters...)
```
ginglis13, line 247: What kind of errors can we expect from `Walk`? I think they would propagate from the provided `snapshots.WalkFunc`. Will these errors affect runtime or will they just be logged as transient?

aarongoulet: It seems like the errors are just being bubbled up to the parent caller: the return type of `WalkAll` in the interface signature is `error`.  I think this should be fine as-is.

snapshotter/demux/snapshotter_test.go
```diff
@@ -73,15 +73,36 @@ func TestReturnErrorWhenCalledWithoutNamespacedContext(t *testing.T) {
...
 		{"Commit", func() error { return uut.Commit(ctx, "layer1", "layerKey") }},
 		{"Remove", func() error { return uut.Remove(ctx, "layerKey") }},
+	}
+
+	for _, test := range tests {
+		if err := test.run(); err == nil {
+			t.Fatalf("%s call did not return error", test.name)
+		}
+	}
+}
```
aarongoulet, line 83: Do any of these tests validate return values, or do they simply confirm that there were no errors?  Is output validation handled elsewhere?

I guess that's a moot point for this PR though, since `WalkAll` only returns the `error` type...

austinvazquez: In their current form, they do not. Maybe room for improvement. The idea was to test expected behaviors under specific conditions. The errors themselves are mocks.

Issue #658, opened 2022-05-31
FUSE enabled kernel needed by some integration tests.


austinvazquez:
Currently the kernel used for integration testing does not appear to have FUSE enabled.

The remote (in-VM) integration tests need this kernel configuration enabled to successfully mount and start a container instead of relying on the thinpool.

It would be nice to be able to install multiple kernels during integration test container image build time and specify which kernel is used during runtime.

austinvazquez:
Integration test error with default kernel installed in container image.

```
make -C snapshotter integ-test
make[1]: Entering directory `/home/ec2-user/workspace/firecracker-containerd/snapshotter'
make integ-test-TestLaunchContainerWithRemoteSnapshotter_Isolated integ-test-TestLaunchMultipleContainersWithRemoteSnapshotter_Isolated
make[2]: Entering directory `/home/ec2-user/workspace/firecracker-containerd/snapshotter'
docker run --rm -it \
        --privileged \
        --ipc=host \
        --network=host \
        --volume /dev:/dev \
        --volume /run/udev/control:/run/udev/control \
        --volume /home/ec2-user/workspace/firecracker-containerd/snapshotter/logs:/var/log/firecracker-containerd-test \
        --volume /home/ec2-user/workspace/firecracker-containerd/snapshotter/..:/src \
        --volume gocache:/go \
        --env ENABLE_ISOLATED_TESTS=1 \
        --env GOPROXY=direct \
        --env GOSUMDB=off \
        --env GO111MODULE=on \
        --env NUMBER_OF_VMS= \
        --workdir="/src/snapshotter" \
        --init \
        localhost/firecracker-containerd-test:latest \
        "go test  -run '^TestLaunchContainerWithRemoteSnapshotter_Isolated$'"
--- FAIL: TestLaunchContainerWithRemoteSnapshotter_Isolated (12.34s)
    service_integ_test.go:130: 
                Error Trace:    service_integ_test.go:130
                                                        service_integ_test.go:61
                Error:          Received unexpected error:
                                failed to extract layer sha256:d2742e2df4a4530fada3b8dbea2b65c4b6fc03fc89f250e8077ecd0425e1ab6c: failed to mount /var/lib/containerd/tmpmounts/containerd-mount1463665100: no such file or directory: unknown
                Test:           TestLaunchContainerWithRemoteSnapshotter_Isolated
                Messages:       Failed to pull alpine image for VM: 0
FAIL
exit status 1
[...]

PR #659, opened 2022-06-01
update kernel configs, make kernel in integ test

ginglis13:

*Issue #, if available:*

#658 

*Description of changes:*

in upgrading to firecracker v1.1, we noticed some configuration options
prompted for using when `make kernel`. These options are specific to
kernel 4.14. Additionally, this upgrade changed the build output from
firecracker's devtool to exclude the full kernel version, and installing would fail as it expects the path to include the patch number.

Some relevant links from the linux kernel driver db:

* https://cateee.net/lkddb/web-lkddb/KMEMCHECK.html
* https://cateee.net/lkddb/web-lkddb/CRYPTO_SALSA20_X86_64.html
* https://cateee.net/lkddb/web-lkddb/FRAME_POINTER_UNWINDER.html
* https://cateee.net/lkddb/web-lkddb/ORC_UNWINDER.html
* https://cateee.net/lkddb/web-lkddb/R3964.html

This PR also updates the integ test Docker image to use a freshly built
kernel such that FUSE is enabled.


tools/kernel-configs/microvm-kernel-x86_64-4.14.config
```diff
@@ -1208,6 +1208,7 @@ CONFIG_VIRTIO_BLK=y
 # CONFIG_ENCLOSURE_SERVICES is not set
 # CONFIG_SRAM is not set
 # CONFIG_C2PORT is not set
+# CONFIG_R3964 is not set
```
Kern--, line 1211: Are these x86 specific or do we also need these options in the ARM config?

tools/docker/Dockerfile.integ-test
```diff
@@ -61,7 +61,8 @@ RUN --mount=type=bind,target=/src make -C /src \
   install \
   install-firecracker \
   install-runc \
-  install-default-vmlinux \
+  kernel \
+  install-kernel \
```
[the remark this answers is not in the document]
ginglis13: I see... write permission may have to do with not building the image as privileged. We can investigate having a pre-built kernel for this use case.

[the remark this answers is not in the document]
ginglis13: +1 to pre-built kernel image for avoiding increased build time. Here are some sample times running locally.

Without the firecracker image public.ecr.aws/firecracker/fcuvm:v35 not yet built: 

```
$ time make kernel
...
real    1m23.834s
```

With firecracker image already built:

```
$ time make kernel
...
real    0m47.606s
```

Makefile
```diff
@@ -60,7 +60,7 @@ KERNEL_CONFIG=tools/kernel-configs/microvm-kernel-$(host_arch)-$(KERNEL_VERSION)
 # Copied from https://github.com/firecracker-microvm/firecracker/blob/v1.0.0/tools/devtool#L2015
 # This allows us to specify a kernel without the patch version, but still get the correct build path to reference the kernel binary
 KERNEL_FULL_VERSION=$(shell cat "$(KERNEL_CONFIG)" | grep -Po "^\# Linux\/$(kernel_config_pattern) (([0-9]+.){2}[0-9]+)" | cut -d ' ' -f 3)
```
swagatbora90, line 62: I normally prefer to clean up unused defined variables. In this case looks like `KERNEL_FULL_VERSION` is no longer being used. I will let you decide if we need to keep this around for future use.

austinvazquez (approved):

Makefile
ginglis13: found the real reason - the regex for kernel version slightly changed in firecracker v1.1.0: https://github.com/firecracker-microvm/firecracker/blob/v1.1.0/tools/devtool#L2082 . updated the regex in our Makefile and reverted back to using `KERNEL_FULL_VERSION`

Issue #651, opened 2022-05-26
Demux Snapshotter Not Closing Remote Snapshotters on VM Shutdown


ginglis13:
The demux snapshotter implements a Close method. We create a snapshotsapi from the demux snapshotter:

```
service := snapshotservice.FromSnapshotter(snapshotter)
```

`snapshotservice` does not implement a Close method, and it does not call `snapshotter.Close`, meaning we will not hit this code path. https://github.com/containerd/containerd/blob/3dbd6a2498f5ec5ffedb26bc377858abb5c9f32c/contrib/snapshotservice/service.go

Even if we did hit this code path, the in-VM snapshotter as a `proxy.RemoteSnapshotter`, which treats `Close` as a noop:

https://github.com/containerd/containerd/blob/c4e29027d42809417f237e4f1662501384dbb702/snapshots/proxy/proxy.go#L184-L186

ginglis13:
I've also noticed that there are no paths to Evict a snapshotter from demux snapshotter's cache.

austinvazquez:
Resolved with #662. Closing.

PR #662, opened 2022-06-06
fix: evict remote snapshotter from cache on connection failure

austinvazquez:

*Issue #, if available:*
#651 

*Description of changes:*
Refactor of the snapshotter caching mechanism to enable eviction policies to be optionally added to the cache during instantiation.

With this change a single caching eviction can be applied to a remote snapshotter cache. Future policies could include TTL based on time or operations.

This change includes an eviction on connection failure policy which will resolve the cache growth issue identified in #651.


snapshotter/app/service.go
```diff
@@ -54,12 +54,19 @@ func Run(config config.Config) error {
...
 		monitor          *metrics.Monitor
 		serviceDiscovery *discovery.ServiceDiscovery
 	)
+
+	cache, err := initCache(config, monitor)
+	if err != nil {
+		log.G(ctx).WithFields(
+			logrus.Fields{"resolver": config.Snapshotter.Proxy.Address.Resolver.Type},
+		).WithError(err).Fatal("failed creating socket resolver")
+		return err
```
kzys, line 67: Does it make sense to include this information into the `err` and pass that to the caller?

snapshotter/config/config.go
```diff
@@ -54,6 +55,11 @@ type resolver struct {
 	Address string `toml:"address"`
 }
 
+type cache struct {
+	EvictOnConnectionFailure bool `toml:"evict_on_connection_failure" default:"true"`
+	PollConnectionFrequency  int  `toml:"poll_connection_frequency"`
```
kzys, line 60: Does it need a default value?

austinvazquez: @kzys for this type of poll where we don't need realtime accuracy I might suggest a frequency of 5 minutes arbitrarily. Would you agree or push more frequent heartbeat?

snapshotter/config/config.go
```diff
@@ -54,6 +55,11 @@ type resolver struct {
 	Address string `toml:"address"`
 }
 
+type cache struct {
+	EvictOnConnectionFailure bool `toml:"evict_on_connection_failure" default:"true"`
+	PollConnectionFrequency  int  `toml:"poll_connection_frequency" default:"300"`
```
ginglis13, line 60: could we rename to `PollConnectionFrequencySeconds` or similar to indicate unit?

snapshotter/demux/cache/evict.go
```diff
@@ -0,0 +1,68 @@
...
+		loop := true
+		for loop {
+			select {
+			case <-ticker.C:
+				if _, err := dial(context.Background(), key); err != nil {
+					p.evict <- key
+					loop = false
+				}
+			case <-p.stop:
+				loop = false
```
ginglis13, line 64: rather than using `loop` you should be able to use plain `for` loop and just `return` in `case <-p.stop`

austinvazquez: Good catch. I generally prefer a single exit for my functions, but in this case the change results in smaller assembly so in theory more performant so I can get onboard with this. I will push the changes.

snapshotter/app/service.go
```diff
@@ -54,21 +53,24 @@ func Run(config config.Config) error {
...
-	cache := cache.NewSnapshotterCache()
-
 	var (
 		monitor          *metrics.Monitor
 		serviceDiscovery *discovery.ServiceDiscovery
 	)
+
+	cache, err := initCache(config, monitor)
+	if err != nil {
+		return fmt.Errorf("failed initializing cache: %v", err)
```
kzys, line 63: (This one and others below) Use %w instead of %v: https://pkg.go.dev/fmt#Errorf

snapshotter/demux/cache/cache.go
```diff
@@ -15,29 +15,166 @@ package cache
...
+func (c *RemoteSnapshotterCache) Get(ctx context.Context, key string) (*proxy.RemoteSnapshotter, error) {
+	c.mutex.RLock()
+	s, ok := c.snapshotters[key]
+	c.mutex.RUnlock()
+
+	if !ok {
+		c.mutex.Lock()
+		defer c.mutex.Unlock()
+
+		s, ok = c.snapshotters[key]
```
kzys, line 115: We did `s, ok := c.snapshotters[key]` a few lines above. Does it need to do that again?

austinvazquez: This is our double check lock to prevent a race condition where two requests populate a snapshotter in the cache potentially leaking resources.

snapshotter/demux/cache/evict.go
```diff
@@ -0,0 +1,67 @@
...
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ticker.C:
+				if _, err := dial(context.Background(), key); err != nil {
+					p.evict <- key
+					return
+				}
+			case <-p.stop:
```
ginglis13, line 62: On stop, should we also be attempting to evict? Thinking of a scenario where we stop enforcing the policy, but leave an item in the cache, which would leak metrics proxy.

Also just for my understanding: when is `p.stop` signalled? On program termination? The channel is closed on `RemoteSnapshotterCache.Close` but I didn't see it written to anywhere

austinvazquez: 1. Leaving entries on the cache is okay in this case. An eviction policy is applied to a cache which each entry having a background eviction enforcer in the form of this goroutine. This goroutine is only stopped by `RemoteSnapshotterCache.Close` as you noted and then which it precedes to empty the cache.
2. So we just need to close the channel. All receives on a closed channel will result in zero values, in this case an empty struct, and we know it is safe to stop.

snapshotter/demux/cache/evict_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+	defer cancel()
+
+	for {
+		select {
+		case <-time.After(10 * time.Millisecond):
+			if len(cache.snapshotters) == 0 {
+				return
+			}
```
ginglis13, line 54: Saw some data race in this test. You might want to acquire the cache lock before reading here (and other tests in this file).

austinvazquez: Ty great catch.

snapshotter/demux/cache/cache.go
```diff
@@ -15,29 +15,173 @@ package cache
...
+}
+
+func (c *RemoteSnapshotterCache) startBackgroundReaper() {
+	reap := func() {
+		for {
+			s, ok := <-c.evict
+			if !ok {
+				break
+			}
+			c.Evict(s)
```
kzys, line 81: Evict is returning an error. If that is not expected, we may need to either log the error or stop the reaper and tell the caller.

snapshotter/demux/cache/cache.go
```diff
@@ -15,29 +15,173 @@ package cache
...
+	c.reaper.Do(func() {
+		go reap()
+	})
+}
+
+func (c *RemoteSnapshotterCache) length() int {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	return len(c.snapshotters)
+}
```
kzys, line 93: Moving to _test.go?

snapshotter/demux/cache/evict.go
```diff
@@ -0,0 +1,67 @@
...
+// On connection failure, the entry will be evicted from cache.
+func (p *EvictOnConnectionFailurePolicy) Enforce(key string) {
+	go func(dial func(context.Context, string) (net.Conn, error)) {
+		ticker := time.NewTicker(p.frequency)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ticker.C:
+				if _, err := dial(context.Background(), key); err != nil {
```
kzys, line 58: Is it okay to pass context.Background()? Can dial() take the context's cancellation signal?

austinvazquez: So if we had something like this where the ticker fired at a very frequent ratio, then cancel would stop all subsequent threads performing dials?

```
		ctx, cancel := context.WithCancel(context.Background())

		for {
			select {
			case <-ticker.C:
				conn, err := dial(ctx, key)
				if err != nil {
					p.evict <- key
					cancel()
					return
				}
				conn.Close()
			case <-p.stop:
				cancel()
				return
			}
		}
```

snapshotter/demux/cache/cache.go
austinvazquez: Will be added in next revision.

snapshotter/demux/cache/evict.go
austinvazquez: We can't send `p.stop` because dial is the same function used by the cache to dial. I feel like the `RemoteSnapshotter` should be reword to use some future-style connection so we don't block on dialing, but that is a separate PR.

kzys: Sorry, I meant, we know that this dial() would fail and this code is using the failure as a signal to remove dead VMs. If so, dial() should fail after a few seconds/minutes instead of waiting indefinitely.

Passing `context.Background()` *indicates* that this function doesn't have any deadlines and would't be cancelled.

austinvazquez: Hmm yes you are right. We have talked before about user configured timeouts. I think it makes sense to potentially try that here where the context has a timeout specified by customer. So the knobs would be

1. dial timeout (overall)
2. dial retry (should be much less than dial timeout)

Is it enough to infer individual dial attempt timeouts or would user specify that as well?

austinvazquez: Added a separate commit with dialer timeout threaded through to the eviction mechanism.

snapshotter/demux/cache/evict.go
```diff
@@ -0,0 +1,74 @@
...
+// from cache after a failed connection attempt occurs.
+type EvictOnConnectionFailurePolicy struct {
+	evictionPolicy
+
+	dialer proxy.Dialer
+
+	frequency time.Duration
+}
+
+// NewEvictOnConnectionFailurePolicy creates a new instance with
```
ginglis13, line 43: nit: godoc not complete

ginglis13 (approved):
LGTM, small nit

snapshotter/demux/cache/evict.go
austinvazquez: Great catch. Let's resolve this. I pushed a new revision with the doc reworked.

kzys (approved):

PR #669, opened 2022-06-09
volume: support in-VM snapshotters

kzys:
With demux snapshotter, firecracker-containerd can run snapshotters
inside VMs. This will allow firecracker-containerd run FUSE-backed
lazy-loading snapshotters such as stargz snapshotter.

This change supports in-VM snapshotters in volume package.

*Issue #, if available:*

*Description of changes:*


snapshotter/volume_integ_test.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+const mib = 1024 * 1024
+
+func TestGuestVolumeFrom_Isolated(t *testing.T) {
+	integtest.Prepare(t, integtest.WithDefaultNetwork())
+
+	runtimes := []string{
+		"aws.firecracker",
+		// "io.containerd.runc.v2"
```
aarongoulet, line 40: Is this commented line still needed?

kzys: No. Unlike volume.FromImage, GuestVolumeImageProvider is pretty tied to Firecracker.

snapshotter/volume_integ_test.go
```diff
@@ -0,0 +1,151 @@
...
+		t.Run(rt, func(t *testing.T) {
+			testGuestVolumeFrom(t, rt)
+		})
+	}
+}
+
+func testGuestVolumeFrom(t *testing.T, runtime string) {
+	const vmID = "default"
+	al2 := imageRef
+	postgres := "docker.io/library/postgres:14.3"
```
aarongoulet, line 53: `al2 := imageRef` doesn't seem necessary, as `imageRef` is a constant and `al2`'s value never changes.  I'd recommend either renaming the constant if clarity was the concern, or just using `imageRef` in place of `al2` directly.

For `postgres`, that can probably be a constant, like `vmID`.  There isn't much consistency with these assignments.

kzys: LeLet me rename imageRef to al2stargz.

snapshotter/volume_integ_test.go
```diff
@@ -0,0 +1,151 @@
...
+	)
+	require.NoError(t, err)
+
+	err = remoteImage.Copy(ctx, "image")
+	require.NoError(t, err)
+
+	err = vs.AddFrom(ctx, remoteImage)
+	require.NoError(t, err)
+
+	image, err := client.Pull(ctx, "docker.io/library/alpine:3.10.1",
```
aarongoulet, line 116: There are a lot of hard-coded image references in this test, but also some declared in the beginning of the function as constants.  We should probably aim for more consistency in this regard.

snapshotter/volume_integ_test.go
```diff
@@ -0,0 +1,145 @@
...
+	_, err = fcClient.SetVMMetadata(ctx, &proto.SetVMMetadataRequest{
+		VMID:     vmID,
+		Metadata: fmt.Sprintf(dockerMetadataTemplate, "ghcr.io", noAuth, noAuth),
+	})
+	require.NoError(t, err, "Failed to configure VM metadata for registry resolution")
+
+	err = remoteImage.Pull(ctx,
+		containerd.WithImageHandlerWrapper(source.AppendDefaultLabelsHandlerWrapper(al2stargz, 10*mib)),
+	)
+	require.NoError(t, err)
```
kzys, line 101: Since different snapshotters need different pull parameters, I have to expose this Pull method to take the pull parameters.

snapshotter/volume_integ_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// Add a stargz image.
+	// The volume directories must be specified since the host's containerd doesn't know about the image.
+	remoteImage := volume.FromGuestImage(
+		client, vmID, al2stargz, "al2-snapshot", []string{"/etc/yum"},
+		volume.WithSnapshotter("demux"),
+	)
+	mount, err := vs.PrepareDriveMount(ctx, 10*mib)
+	require.NoError(t, err)
```
kzys, line 66: I cannot do `err = vs.AddFrom(ctx, remoteImage)` here since the image cannot be pulled before starting the VM.

Alternatively we could make volume.Set "smarter" and treat GuestVolumeImageProvider somewhat differently.

volume/guest_image.go
```diff
@@ -0,0 +1,217 @@
...
+
+	result := make([]*Volume, 0, len(p.volumes))
+	for _, path := range p.volumes {
+		result = append(result, &Volume{
+			vmPath:        filepath.Join(p.vmDir, path),
+			containerPath: path,
+		})
+	}
+	return result, nil
+}
```
kzys, line 212: I probably should make this method specific to imageVolumeProvider. GuestVolumeImageProvider doesn't use tempDir.

volume/cmd/volume-init/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/containerd/continuity/fs"
+	"github.com/firecracker-microvm/firecracker-containerd/volume"
+)
+
+func copy(in volume.GuestVolumeImageInput) error {
+	err := os.MkdirAll(in.From, 0700)
```
austinvazquez, line 28: Why do we need to `MkdirAll` from here? Just to ensure read permissions?

kzys: Good catch. From should be there...

snapshotter/volume_integ_test.go
Kern--: Thinking through the way you actually setup volumes, we need one of:
`Set.PrepareDirectory`
or
`Set.PrepareDriveMount` and maybe `GuestVolumeImageProvider.Copy`

If we made `volume.Set` smarter so that we could user `vs.AddFrom` here, then it seems like we could make the setup a little less confusion with something like:

```
mount, err := set.PrepareDriveMount(ctx, 10*mib)
// launch vm
err = set.PrepareInGuestVolumes(ctx)
```

Maybe there's value in renaming the `Prepare` methods to match the function they provider rather than what they literally do? e.g.
`set.PrepareHostOnlyVoumes`, `set.PrepareVMPreLaunchVolumes`, `set.PrepareVMPostLaunchVolumes`

A couple things I'm not sure of with this proposal:
1. Does it push too much logic into `volume.Set`? 
2. Is it too obfuscated from what's actually going on (tbf, I'm not sure `set.AddFrom` is setting up temp mounts of the image or that `set.PrepareDriveMount` is copying data into a thin device)

snapshotter/volume_integ_test.go
Kern--: Could pull options be a functional parameter to `FromGuestImage`?

volume/cmd/volume-init/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+	err := os.MkdirAll(filepath.Dir(path), 0700)
+	if err != nil {
+		return err
+	}
+
+	return os.WriteFile(path, []byte(content), 0600)
+}
+
+func TestCopy(t *testing.T) {
+	from, err := os.MkdirTemp("", "From")
```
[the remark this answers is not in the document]
kzys: Will do. I was using MkdirTemp since it was easier to inspect during debugging.

volume/cmd/volume-init/main.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return err
+	}
+	for _, v := range in.Volumes {
+		err := os.MkdirAll(filepath.Join(in.To, v), 0700)
+		if err != nil {
+			return err
+		}
+
+		err = fs.CopyDir(filepath.Join(in.To, v), filepath.Join(in.From, v))
```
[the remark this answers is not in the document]
kzys: Sorry. Both should be.

volume/guest_image.go
```diff
@@ -0,0 +1,217 @@
...
+				Type:        "bind",
+				Options:     []string{"bind"},
+			},
+		)
+		return nil
+	}
+}
+
+// Copy files from the image by launching the container.
+func (p *GuestVolumeImageProvider) Copy(ctx context.Context, name string) error {
```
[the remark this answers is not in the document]
kzys: Let me update the doc comment.

snapshotter/volume_integ_test.go
kzys: No. Seems Go methods only have one `opts ...Type`.

snapshotter/volume_integ_test.go
kzys: Let me think about. volume.Set isn't aware about Providers, but I probably should prioritize usability over extensibility this time.

snapshotter/volume_integ_test.go
Kern--: Is this too hacky?
```
type imageProviderConfig struct {
	snapshotter string
        pullOpts []containerd.RemoteOpts
}

func WithPullOptions(opts ...containerd.RemoteOpts) ImageOpt {
    return func(c *imageProviderConfig) {
        c.pullOpts = opts
    }
}
```

kzys: That's good!

Kern-- (approved):

volume/guest_image.go
```diff
@@ -0,0 +1,223 @@
...
+	task, err := container.NewTask(ctx, cio.NewCreator(cio.WithStreams(stdin, &stdout, os.Stderr)))
+	if err != nil {
+		return err
+	}
+	defer task.Delete(ctx)
+
+	exitCh, err := task.Wait(ctx)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
ginglis13: this is correct, should begin to wait for task before starting it to avoid races. See https://github.com/containerd/containerd/blob/main/docs/getting-started.md#task-wait-and-start

> This makes sure that you do not encounter any races if the task has a simple program like /bin/true that exits promptly after calling start.

ginglis13 (approved):

austinvazquez (approved):

PR #671, opened 2022-06-12
add metrics integration tests

ginglis13:

*Issue #, if available:*

*Description of changes:*

This PR adds happy path integration testing for metrics. TODO included
for testing changes in service discovery response as VMs are started and
stopped and evicted from the demux snapshotter cache.

### DIFF
--- a/snapshotter/metrics_integ_test.go
+++ b/snapshotter/metrics_integ_test.go
@@ -1,0 +1,176 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+package main
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strconv"
+	"testing"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/namespaces"
+	"github.com/firecracker-microvm/firecracker-containerd/firecracker-control/client"
+	"github.com/firecracker-microvm/firecracker-containerd/internal/integtest"
+	"github.com/firecracker-microvm/firecracker-containerd/proto"
+	"github.com/firecracker-microvm/firecracker-containerd/snapshotter/internal/integtest/stargz/fs/source"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	serviceDiscoveryEndpoint = "http://localhost:8080"
+
+	stargzFsOperation     = "stargz_fs_operation_duration_milliseconds_bucket"
+	stargzFsOperationHelp = "# HELP stargz_fs_operation_duration_milliseconds Latency"
+	stargzFsOperationType = "# TYPE stargz_fs_operation_duration_milliseconds histogram"
+)
+
+func TestSnapshotterMetrics_Isolated(t *testing.T) {
+	integtest.Prepare(t)
+
+	vmID := 0
+
+	ctx := namespaces.WithNamespace(context.Background(), strconv.Itoa(vmID))
+
+	fcClient, err := integtest.NewFCControlClient(integtest.ContainerdSockPath)
+	defer fcClient.StopVM(ctx, &proto.StopVMRequest{VMID: strconv.Itoa(vmID)})
+	require.NoError(t, err, "Failed to create fccontrol client")
+
+	require.NoError(t, pullImageWithRemoteSnapshotterInVM(ctx, strconv.Itoa(vmID), fcClient))
+	verifyMetricsResponse(t, 1)
+}
+
+func TestSnapshotterMetricsMultipleVMs_Isolated(t *testing.T) {
+	integtest.Prepare(t)
+
+	numberOfVms := integtest.NumberOfVms
+	fcClient, err := integtest.NewFCControlClient(integtest.ContainerdSockPath)
+	require.NoError(t, err, "Failed to create fccontrol client")
+
+	group, ctx := errgroup.WithContext(context.Background())
+
+	for vmID := 0; vmID < numberOfVms; vmID++ {
+		id := vmID
+		ctxNamespace := namespaces.WithNamespace(ctx, strconv.Itoa(id))
+		defer fcClient.StopVM(ctxNamespace, &proto.StopVMRequest{VMID: strconv.Itoa(id)})
+
+		group.Go(
+			func() error {
+				return pullImageWithRemoteSnapshotterInVM(ctxNamespace, strconv.Itoa(id), fcClient)
+			},
+		)
+
+	}
+
+	err = group.Wait()
+	require.NoError(t, err)
+	verifyMetricsResponse(t, numberOfVms)
+}
+
+// TODO (ginglis13): ensure service discovery response changes as VMs are started and stopped.
+// pending https://github.com/firecracker-microvm/firecracker-containerd/issues/651
+// func TestSnapshotterMetricsMultipleVMsStopStart_Isolated(t *testing.T) {}
+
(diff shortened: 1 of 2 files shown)


snapshotter/metrics_integ_test.go
```diff
@@ -0,0 +1,165 @@
...
+			VcpuCount:  1,
+			MemSizeMib: 512,
+		},
+		ContainerCount: 1,
+	})
+	require.NoErrorf(t, err, "Failed to create microVM[%s]", vmID)
+
+	_, err = fcClient.SetVMMetadata(ctx, &proto.SetVMMetadataRequest{
+		VMID:     vmID,
+		Metadata: fmt.Sprintf(dockerMetadataTemplate, "ghcr.io", "", ""),
```
austinvazquez, line 119: Could use `noAuth` const in place of `""`

austinvazquez (approved):

snapshotter/metrics_integ_test.go
```diff
@@ -0,0 +1,165 @@
...
+	require.NoError(t, err, "Failed to create fccontrol client")
+
+	pullImageWithRemoteSnapshotterInVM(ctx, t, strconv.Itoa(vmID), fcClient)
+	verifyMetricsResponse(t, 1)
+}
+
+func TestSnapshotterMetricsMultipleVMs_Isolated(t *testing.T) {
+	integtest.Prepare(t, integtest.WithDefaultNetwork())
+
+	testTimeout := 300 * time.Second
```
kzys, line 61: You could use time.Minute instead

snapshotter/metrics_integ_test.go
```diff
@@ -0,0 +1,165 @@
...
+			fcClient.StopVM(ctxInner, &proto.StopVMRequest{VMID: strconv.Itoa(id)})
+		}(namespaces.WithNamespace(ctx, strconv.Itoa(vmID)), vmID)
+
+		wg.Add(1)
+		go func(ctxInner context.Context, id int) {
+			defer wg.Done()
+			pullImageWithRemoteSnapshotterInVM(ctxInner, t, strconv.Itoa(id), fcClient)
+		}(namespaces.WithNamespace(ctx, strconv.Itoa(vmID)), vmID)
+
+	}
```
kzys, line 81: It is better to not pass t to different goroutines.

https://pkg.go.dev/testing#T.FailNow

> FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test. Calling FailNow does not stop those other goroutines. 

And testify's [require](https://github.com/stretchr/testify#require-package) functions call FailNow

kzys (changes requested):

snapshotter/metrics_integ_test.go
ginglis13: Ok, I'll attempt using an error channel rather than passing testing.T to these goroutines. Looks like we have the same potential issue here as well: https://github.com/firecracker-microvm/firecracker-containerd/blob/47ceb625562fadcd7d615f519a24086a45cf73cc/snapshotter/service_integ_test.go#L74-L81

kzys: Returning an error from launchContainerWithRemoteSnapshotterInVM and using errgroup may be the way to go :)

snapshotter/metrics_integ_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+		group.Go(
+			func() error {
+				return pullImageWithRemoteSnapshotterInVM(ctxNamespace, strconv.Itoa(id), fcClient)
+			},
+		)
+
+	}
+
+	if err := group.Wait(); err != nil {
```
Kern--, line 84: nit: I think this syntax is preferred:
```
err = group.Wait()
require.NoError(t, err, "message")
```

snapshotter/metrics_integ_test.go
```diff
@@ -0,0 +1,175 @@
...
+	uniqueTargets := make(map[string]struct{})
+	for _, mt := range metricsTargets {
+		uniqueTargets[mt.Targets[0]] = struct{}{}
+		metricsResponse, err := http.Get("http://" + mt.Targets[0] + "/metrics")
+		require.NoError(t, err, "Failed to get metrics proxy")
+
+		mBytes, err := io.ReadAll(io.LimitReader(metricsResponse.Body, 300))
+		require.NoError(t, err, "Failed to read metrics response body")
+		require.NotEmpty(t, mBytes)
+		metricsResponse.Body.Close()
```
Kern--, line 172: nit: this should probably be deferred before the io.ReadAll.

ginglis13:
Most recent buildkite run is passing metrics integ tests, timeout on `TestLaunchMultipleContainersWithRemoteSnapshotter_Isolated`: https://buildkite.com/firecracker-microvm/firecracker-containerd/builds/3087#01816328-3a13-476f-9be9-cf7d69be8529 . related: #673

austinvazquez (approved):

snapshotter/metrics_integ_test.go
```diff
@@ -0,0 +1,178 @@
...
+	require.NoError(t, err, "Failed to unmarshall service discovery response")
+
+	require.Len(t, metricsTargets, numberOfVms)
+
+	// Test pulling individual metrics proxies
+	// uniqueTargets acts as a set to ensure each metrics target is a unique host:port
+	uniqueTargets := make(map[string]struct{})
+	for _, mt := range metricsTargets {
+		uniqueTargets[mt.Targets[0]] = struct{}{}
+		metricsResponse, err := http.Get("http://" + mt.Targets[0] + "/metrics")
```
Kern--, line 169: We should probably verify the response HTTP code is OK.

snapshotter/metrics_integ_test.go
```diff
@@ -0,0 +1,178 @@
...
+		require.NoError(t, err, "Failed to get metrics proxy")
+		defer metricsResponse.Body.Close()
+
+		mBytes, err := io.ReadAll(io.LimitReader(metricsResponse.Body, 300))
+		require.NoError(t, err, "Failed to read metrics response body")
+		require.NotEmpty(t, mBytes)
+	}
+	require.Len(t, uniqueTargets, numberOfVms)
```
Kern--, line 177: If all we're verifying is that the snapshotter metrics endpoint returns some bytes, do we actually need to pull an image?

snapshotter/metrics_integ_test.go
ginglis13: added check for HTTP status OK

snapshotter/metrics_integ_test.go
ginglis13: The demux snapshotter needs some snapshot API call in order to populate a remote snapshotter in its cache. If we only launch the VM, then yes, the we could pull from its metrics proxy server, but service discovery would not return anything as the demux snapshotter's cache would be empty.

Kern-- (approved):

snapshotter/metrics_integ_test.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+func TestSnapshotterMetrics_Isolated(t *testing.T) {
+	integtest.Prepare(t, integtest.WithDefaultNetwork())
+
+	vmID := 0
+
+	testTimeout := 2 * time.Minute
+	ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
+	defer cancel()
```
kzys, line 46: Like Austin's another PR, let's remove timeouts. It seems too arbitrary.

snapshotter/metrics_integ_test.go
```diff
@@ -0,0 +1,179 @@
...
+// TODO (ginglis13): ensure service discovery response changes as VMs are started and stopped.
+// pending https://github.com/firecracker-microvm/firecracker-containerd/issues/651
+// func TestSnapshotterMetricsMultipleVMsStopStart_Isolated(t *testing.T) {}
+
+func pullImageWithRemoteSnapshotterInVM(ctx context.Context, vmID string, fcClient *client.Client) error {
+	client, err := containerd.New(integtest.ContainerdSockPath, containerd.WithDefaultRuntime(integtest.FirecrackerRuntime))
+	if err != nil {
+		return fmt.Errorf("Unable to create client to containerd service at %s, is containerd running?: %v", integtest.ContainerdSockPath, err)
+	}
+	kernelArgs := integtest.DefaultRuntimeConfig.KernelArgs + " 8250.nr_uarts=0 quiet loglevel=1"
```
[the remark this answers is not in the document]
austinvazquez: @kzys it hasn't proven to reduce the flakiness but figured it would help increase firecracker vm stability even if we aren't stress testing the in-vm snapshotter.

kzys (approved):

Issue #670, opened 2022-06-10
Remove pkg/errors


kzys:
https://github.com/pkg/errors is being archived and containerd and Firecracker Go SDK are now using fmt.Errorf instead.

PR #679, opened 2022-06-17
chore: remove pkg/errors dependency

austinvazquez:

*Issue #, if available:*
#670 

*Description of changes:*
Remove all references to pkg/errors.


go.mod
```diff
@@ -23,7 +23,6 @@ require (
 	github.com/opencontainers/runc v1.1.2
 	github.com/opencontainers/runtime-spec v1.0.3-0.20210910115017-0d6cc581aeea
 	github.com/pelletier/go-toml v1.9.3
-	github.com/pkg/errors v0.9.1
```
austinvazquez, line 26: @kzys how are `go.sum` changes propagated?

runtime/service.go
```diff
@@ -482,10 +482,10 @@ func (s *service) CreateVM(requestCtx context.Context, request *proto.CreateVMRe
 	if err != nil {
 		s.shimCancel()
 		s.logger.WithError(err).Error("failed to create VM")
-		if errors.Cause(err) == context.DeadlineExceeded {
+		if strings.Contains(err.Error(), context.DeadlineExceeded.Error()) {
```
austinvazquez, line 485: @kzys everything else was pretty simple changes, but I wasn't sure the idiomatic change here. No longer wrapping errors so I assumed a substring check.

Kern--: `%w` in the format string wraps the error. The way to check is `errors.Is(err, context.DeadlineExceeded)`

go.mod
Kern--: `go mod tidy` but if any of our transitive dependencies include it, then it will still be there.

runtime/service.go
austinvazquez: Oh, +1. I had only checked errors.go should have checked wrap.go as well. :)

go.mod
austinvazquez: Ah that was my confusion.

PR #683, opened 2022-06-21
fix: extend vsock ack message timeout for remote snapshot dials

austinvazquez:

*Issue #, if available:*

*Description of changes:*


Kern-- (approved):

snapshotter/service_integ_test.go
```diff
@@ -54,7 +54,7 @@ func TestLaunchContainerWithRemoteSnapshotter_Isolated(t *testing.T) {
 
 	vmID := 0
 
-	testTimeout := 300 * time.Second
+	testTimeout := 1 * time.Minute
```
kzys, line 57: go test has its own timeout. Do we really need the timeouts in the tests?

austinvazquez: Guess not. The default is fine for both this test and the `MultipleContainers` test on local runs. So it's fine to just remove them.

kzys (approved):

Issue #673, opened 2022-06-14
Stabilize remote snapshotter integration tests.


austinvazquez:
The remote snapshotter integration tests are somewhat unstable when running on buildkite pipeline. Some datapoints observed so far.

1. Does not appear to be a OS issue. (Local runs on Debian and AL2 are successful)
2. Timeout of 10 minutes should be enough for this test.
3. Kernel args have been expanded to lower IO to prevent firecracker blocking issue.

austinvazquez:
Stopgap solution is to extend the timeout for dials.
https://github.com/firecracker-microvm/firecracker-containerd/pull/690

The issue is this works fine for local runs and PR builds, but mainline experiences different results.

austinvazquez:
No failures since June 27th. Resolving this ticket. Feel free to reopen if any flakiness is observed.

PR #684, opened 2022-06-22
add: remote snapshotter dialer timeouts to service config

austinvazquez:

*Issue #, if available:*
#673 

*Description of changes:*
Expose all the dialer timeout knobs at the snapshotter service level. This enables tuning by users based on their host stability.

Buildkite instances used by the pipeline tend to be somewhat slower compared to developer instances, so this helps enable stabilizing remote snapshotter integration tests.

Additionally, remove some unneeded code from the remote snapshotter integration tests and make the tests run in parallel again to speed up the build.

### DIFF
--- a/snapshotter/app/service.go
+++ b/snapshotter/app/service.go
@@ -156,11 +156,54 @@ func initResolver(config config.Config) (proxyaddress.Resolver, error) {
 	}
 }
 
+type dialerTimeouts struct {
+	DialTimeout       time.Duration
+	RetryTimeout      time.Duration
+	RetryInterval     time.Duration
+	ConnectMsgTimeout time.Duration
+	AckMsgTimeout     time.Duration
+}
+
+func parseDialerTimeouts(config config.Config) (dialerTimeouts, error) {
+	dialTimeout, err := time.ParseDuration(config.Snapshotter.Dialer.DialTimeout)
+	if err != nil {
+		return dialerTimeouts{}, fmt.Errorf("Failed parsing dialer dial timeout: %w", err)
+	}
+	retryTimeout, err := time.ParseDuration(config.Snapshotter.Dialer.RetryTimeout)
+	if err != nil {
+		return dialerTimeouts{}, fmt.Errorf("Failed parsing dialer retry timeout: %w", err)
+	}
+	retryInterval, err := time.ParseDuration(config.Snapshotter.Dialer.RetryInterval)
+	if err != nil {
+		return dialerTimeouts{}, fmt.Errorf("Failed parsing dialer retry interval: %w", err)
+	}
+	connectMsgTimeout, err := time.ParseDuration(config.Snapshotter.Dialer.ConnectMsgTimeout)
+	if err != nil {
+		return dialerTimeouts{}, fmt.Errorf("Failed parsing dialer connect msg timeout: %w", err)
+	}
+	ackMsgTimeout, err := time.ParseDuration(config.Snapshotter.Dialer.AckMsgTimeout)
+	if err != nil {
+		return dialerTimeouts{}, fmt.Errorf("Failed parsing dialer ack msg timeout: %w", err)
+	}
+	return dialerTimeouts{
+		DialTimeout:       dialTimeout,
+		RetryTimeout:      retryTimeout,
+		RetryInterval:     retryInterval,
+		ConnectMsgTimeout: connectMsgTimeout,
+		AckMsgTimeout:     ackMsgTimeout,
+	}, nil
+}
+
 const base10 = 10
 const bits32 = 32
 
 func initSnapshotter(ctx context.Context, config config.Config, cache cache.Cache, monitor *metrics.Monitor) (snapshots.Snapshotter, error) {
 	resolver, err := initResolver(config)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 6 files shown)


snapshotter/service_integ_test.go
```diff
@@ -49,15 +49,15 @@ const (
 )
 
 func TestLaunchContainerWithRemoteSnapshotter_Isolated(t *testing.T) {
-	integtest.Prepare(t, integtest.WithDefaultNetwork())
```
austinvazquez, line 52: Not needed. All `CreateVM` calls specify network interface.

snapshotter/service_integ_test.go
```diff
@@ -104,15 +100,14 @@ func launchContainerWithRemoteSnapshotterInVM(ctx context.Context, vmID string)
 				AllowMMDS: true,
 				CNIConfig: &proto.CNIConfiguration{
 					NetworkName:   "fcnet",
-					InterfaceName: "veth0",
+					InterfaceName: fmt.Sprintf("veth%s", vmID),
```
austinvazquez, line 103: While I don't believe having each microVM on the same interface causes any issue, there also isn't a reason to have them on the same interface so just place each VM on it's own interface.

snapshotter/config/config.go
```diff
@@ -43,7 +43,11 @@ type listener struct {
 }
 
 type dialer struct {
-	AckMsgTimeoutInSeconds int `toml:"ack_msg_timeout_in_seconds" default:"1"`
+	DialTimeout       string `toml:"dial_timeout" default:"100ms"`
+	RetryTimeout      string `toml:"retry_timeout" default:"10s"`
+	RetryInterval     string `toml:"retry_interval" default:"5s"`
```
austinvazquez, line 48: This might be too slow but we have the knobs now and can tune based on observations.

snapshotter/app/service.go
```diff
@@ -156,6 +156,44 @@ func initResolver(config config.Config) (proxyaddress.Resolver, error) {
 	}
 }
 
+type dialerTimeouts struct {
```
kzys, line 159: I don't like the direction it is heading to.

- firecracker-containerd itself is using vsock but we don't have them. What is the difference between firecracker-containerd and demux snapshotter?
- Exposing all knobs to end users basically moving the responsibility from ourselves to them, which isn't good.

snapshotter/app/service.go
```diff
@@ -177,9 +220,15 @@ func initSnapshotter(ctx context.Context, config config.Config, cache cache.Cach
...
 		}
 
-		ackMsgTimeout := time.Duration(config.Snapshotter.Dialer.AckMsgTimeoutInSeconds) * time.Second
 		snapshotterDialer := func(ctx context.Context, namespace string) (net.Conn, error) {
-			return vsock.DialContext(ctx, host, uint32(port), vsock.WithLogger(log.G(ctx)), vsock.WithAckMsgTimeout(ackMsgTimeout))
+			return vsock.DialContext(ctx, host, uint32(port),
+				vsock.WithLogger(log.G(ctx)),
+				vsock.WithDialTimeout(dialerTimeouts.DialTimeout),
+				vsock.WithRetryTimeout(dialerTimeouts.RetryTimeout),
+				vsock.WithRetryInterval(dialerTimeouts.RetryInterval),
```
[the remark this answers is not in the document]
kzys: 100ms -> 5s is 50x change. Why do you specifically pick 5s? Does it make sense to reduce a bit, like 1s?

snapshotter/config/config.go
kzys: Regarding the timeouts, if we don't have them, would it wait indefinitely without clear errors? I wonder whether we could have a single timeout, instead of asking end-users to budget all of them specifically.

snapshotter/service_integ_test.go
kzys: Let's have this one in either a different commit and/or PR.

kzys (changes requested):

snapshotter/app/service.go
austinvazquez: That's a fair comment. My reasoning for lift and shifting the responsibility was, and based purely on observations, the dial performance seemed to be greatly impacted by host load. I think we could consider mechanisms like exponential backoff to help reduce the number of knobs and loss of performance.

snapshotter/app/service.go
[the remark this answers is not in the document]
austinvazquez: So I noticed the ticker was firing every 100ms and executing code that had up to 1s timeout. Although I assume separate goroutines handle the execution, I wanted to see if lowering the frequency increased stability. 5s is somewhat arbitrary but based on doing runs with up to 4s acknowledge message timeouts which still lead to timeouts.

snapshotter/config/config.go
austinvazquez: I have not done any experiments here to prove without the timeouts. I agree reducing complexity would be better as long as the solution becomes stable.

snapshotter/service_integ_test.go
ginglis13: Question regarding this: have you tried just making this change and observed any different behavior on the BuildKite tests?

austinvazquez:
Superseded by #689, #690

PR #698, opened 2022-08-04
add gcc to quickstart yum packages and getting started prerequisites

ginglis13:

*Issue #, if available:*

n/a

*Description of changes:*

Noticed in going through the firecracker-containerd quickstart for Amazon Linux that by
default, gcc is not installed. add that as a yum package to be
installed. Just copy-pasting as is will run into an error of gcc not found.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -38,6 +38,7 @@ You need to have the following things in order to use firecracker-containerd:
 
   </details>
 * git
+* gcc, required by the Firecracker agent for building
 * A recent installation of [Docker CE](https://docker.com).
 * Go 1.16 or later, which you can download from [here](https://golang.org/dl/).
 
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -84,7 +84,8 @@ sudo yum -y install \
   e2fsprogs \
   util-linux \
   bc \
-  gnupg
+  gnupg \
+  gcc
 
 
 # Amazon Linux 2 packages can sometimes be dated, so let's install using


Kern--:
Can you update the getting started guide as well with an explanation of what requires gcc?

austinvazquez:
Does this need to be included on RPM based instances as well?

ginglis13:
> Can you update the getting started guide as well with an explanation of what requires gcc?

Sure thing, added a line to the gettting started guide - building the in-VM agent depends on gcc: https://github.com/firecracker-microvm/firecracker-containerd/blob/279a03c70541cd355f4f4a8adda0480fbc6fb95d/agent/Makefile#L24-L29

> Does this need to be included on RPM based instances as well?

This change does cover the RPM instance (AL2/yum). I tested on a fresh Debian instance, no issues there so no update required to the Debian scripts/instructions.

austinvazquez (approved):

kzys:
What was the error you got by not having gcc? I believe that firecracker-containerd doesn't have C dependencies.

ginglis13:
> What was the error you got by not having gcc? I believe that firecracker-containerd doesn't have C dependencies.

The error is `gcc` not in `PATH` in building the agent. I think there is a C dependency on using the Go linker. Here is a fresh AL2 ec2 instance where I've simply copy-pasted our quickstart guide up until `sg docker -c 'make all image firecracker'`:

```
$ sg docker -c 'make all image firecracker'
make -C agent EXTRAGOARGS=
make[1]: Entering directory `/home/ec2-user/firecracker-containerd/agent'
go build  -ldflags "-X main.revision=279a03c70541cd355f4f4a8adda0480fbc6fb95d" -o agent
...
/home/ec2-user/.go/pkg/tool/linux_amd64/link: running gcc failed: exec: "gcc": executable file not found in $PATH
```

kzys:
Can you force internal linking? Not really sure why it is doing external linking.

https://github.com/bazelbuild/rules_go/issues/2719#issuecomment-736871039

ginglis13:
> Can you force internal linking? Not really sure why it is doing external linking.
> 
> [bazelbuild/rules_go#2719 (comment)](https://github.com/bazelbuild/rules_go/issues/2719#issuecomment-736871039)

I can by passing `-linkmode=internal` to ldflags; I'm then seeing all other invocations of `go build` (e.g. for runtime, examples, volume, snapshotter, etc.) also attempt to use `gcc` as external linker, so would need to provide this linkmode flag to those rules as well... 

Better option if we don't want to include gcc installation in quickstart/getting started is to change `sg docker -c 'make all image firecracker'` to `CGO_ENABLED=0 sg docker -c 'make all image firecracker'`, since `CGO_ENABLED=0` will use Go's internal linker.

@kzys would you prefer that change to the quickstart over having users install gcc?

kzys:
Sorry for the very late reply. This is good to go.

PR #727, opened 2023-01-11
Set ARCH to CTRD and CRITEST

vvejell1:
Signed-off-by: Vaishnavi Vejella <vvejella@amazon.com>

*Issue:*
**Containerd** for ARM is not present to get runc shim and **Critest** binary is also not there for ARM architecture.

*Description of changes:*
We can install appropriate binaries by providing the Architecture to both Containerd and Critest in the Dockerfile.

### DIFF
--- a/tools/docker/Dockerfile.integ-test
+++ b/tools/docker/Dockerfile.integ-test
@@ -31,14 +31,16 @@ RUN mkdir -p \
   /etc/cni/net.d
 
 # Install containerd to have runc shim.
-ENV CTRD_VERSION="1.6.4"
-RUN wget --quiet -O- https://github.com/containerd/containerd/releases/download/v$CTRD_VERSION/containerd-$CTRD_VERSION-linux-amd64.tar.gz | tar zxf - -C /tmp/ && \
+ENV CTRD_VERSION="1.6.8"
+RUN ARCH=`go env GOARCH` && \
+	wget --quiet -O- https://github.com/containerd/containerd/releases/download/v$CTRD_VERSION/containerd-$CTRD_VERSION-linux-${ARCH}.tar.gz | tar zxf - -C /tmp/ && \
   install -D -o root -g root -m755 -t /usr/local/bin /tmp/bin/containerd-shim-runc-v2 && \
   rm -rf /tmp/bin
 
 # Install critest.
 ENV CRITEST_VERSION="1.23.0"
-RUN wget --quiet -O- https://github.com/kubernetes-sigs/cri-tools/releases/download/v$CRITEST_VERSION/critest-v$CRITEST_VERSION-linux-amd64.tar.gz | tar zxf - -C /tmp/ && \
+RUN ARCH=`go env GOARCH` && \
+  wget --quiet -O- https://github.com/kubernetes-sigs/cri-tools/releases/download/v$CRITEST_VERSION/critest-v$CRITEST_VERSION-linux-${ARCH}.tar.gz | tar zxf - -C /tmp/ && \
   install -D -o root -g root -m755 -t /usr/local/bin /tmp/critest && \
   rm -f /tmp/critest
 


vvejell1:
No issues for `make all-in-docker`, `make proto` on ARM instance.

tools/docker/Dockerfile.integ-test
```diff
@@ -30,18 +30,30 @@ RUN mkdir -p \
   ${FICD_LOG_DIR} \
   /etc/cni/net.d
 
-# Install containerd to have runc shim.
-ENV CTRD_VERSION="1.6.4"
+# Install containerd to have runc shim for amd.
```
ginglis13, line 33: The install for ARM will overwrite the x86 shim binary, which causes failures in tests requiring this binary run on x86 machines. Let's distinguish between the architecture this is running on so we download and install the proper binary.

```suggestion
# Install containerd to have runc shim.
ENV CTRD_VERSION="1.6.8"
RUN ARCH=`uname -m` && \
  if [ "$ARCH" == "x86_64" ]; then \
      ARCH="amd64"; \
  elif [ "$ARCH" == "aarch64" ]; then  \
      ARCH="arm64"; \
  fi  && \
  wget --quiet -O- https://github.com/containerd/containerd/releases/download/v$CTRD_VERSION/containerd-$CTRD_VERSION-linux-$ARCH.tar.gz | tar zxf - -C /tmp/ && \
  install -D -o root -g root -m755 -t /usr/local/bin /tmp/bin/containerd-shim-runc-v2 && \
  rm -rf /tmp/bin

```

tools/docker/Dockerfile.integ-test
```diff
@@ -30,18 +30,30 @@ RUN mkdir -p \
...
 RUN wget --quiet -O- https://github.com/kubernetes-sigs/cri-tools/releases/download/v$CRITEST_VERSION/critest-v$CRITEST_VERSION-linux-amd64.tar.gz | tar zxf - -C /tmp/ && \
   install -D -o root -g root -m755 -t /usr/local/bin /tmp/critest && \
   rm -f /tmp/critest
 
+# Install critest for ARM.
+ENV CRITEST_VERSION="1.23.0"
+RUN wget --quiet -O- https://github.com/kubernetes-sigs/cri-tools/releases/download/v$CRITEST_VERSION/critest-v$CRITEST_VERSION-linux-arm64.tar.gz | tar zxf - -C /tmp/ && \
+  install -D -o root -g root -m755 -t /usr/local/bin /tmp/critest && \
+  rm -f /tmp/critest
+
```
ginglis13, line 56: same suggestion as above - this will overwrite the critest binary installed for x86.

ginglis13 (changes requested):

tools/docker/Dockerfile.integ-test
vvejell1: Noted, thanks for the suggestion

ginglis13 (approved):
LGTM

tools/docker/Dockerfile.integ-test
```diff
@@ -30,15 +30,29 @@ RUN mkdir -p \
   ${FICD_LOG_DIR} \
   /etc/cni/net.d
 
+RUN echo `uname -m`
+
 # Install containerd to have runc shim.
-ENV CTRD_VERSION="1.6.4"
-RUN wget --quiet -O- https://github.com/containerd/containerd/releases/download/v$CTRD_VERSION/containerd-$CTRD_VERSION-linux-amd64.tar.gz | tar zxf - -C /tmp/ && \
+ENV CTRD_VERSION="1.6.8"
+RUN ARCH=`uname -m` && \
```
kzys, line 37: I believe that you can call `go` command in this Dockerfile. Can you try `go env GOARCH`? That should return amd64/arm64 instead.

runtime/service_integ_test.go
```diff
@@ -305,10 +305,11 @@ func testMultipleVMs(ctx context.Context, t *testing.T, count int) {
...
 			JailerConfig: &proto.JailerConfig{
-				UID:        300000,
-				GID:        300000,
-				NetNS:      netns.Path(),
-				CgroupPath: "/mycgroup",
+				UID:           300000,
+				GID:           300000,
+				NetNS:         netns.Path(),
+				CgroupPath:    "/mycgroup",
+				CgroupVersion: "2",
```
kzys, line 312: Does it really work? The newly-added CgroupVersion field won't be used automatically.

tools/docker/Dockerfile.integ-test
vvejell1: After trying `go env GOARCH`, it returned arm64

runtime/service_integ_test.go
vvejell1: yes, it helped me to resolve the test **TestMultipleVMs_Isolated** and I have added **CgroupVersion** field using `firecracker.proto` file then `make proto` helped in auto-generating the field in `firecracker.pb.go` file.

tools/docker/Dockerfile.integ-test
vvejell1: Added ` go env GOARCH` for CTRD and critest.

proto/firecracker.proto
```diff
@@ -136,6 +136,8 @@ message JailerConfig {
 
     // DriveExposePolicy is used to configure the method to expose drive files.
     DriveExposePolicy DriveExposePolicy = 7;
+    // CgroupVersion is the version of the cgroup filesystem to use.
+	string CgroupVersion = 8;
```
austinvazquez, line 140: Minor nit spacing here is off.

tools/docker/Dockerfile.integ-test
```diff
@@ -31,14 +31,16 @@ RUN mkdir -p \
   /etc/cni/net.d
 
 # Install containerd to have runc shim.
-ENV CTRD_VERSION="1.6.4"
-RUN wget --quiet -O- https://github.com/containerd/containerd/releases/download/v$CTRD_VERSION/containerd-$CTRD_VERSION-linux-amd64.tar.gz | tar zxf - -C /tmp/ && \
+ENV CTRD_VERSION="1.6.8"
```
austinvazquez, line 34: Should this be 1.6.9?

vvejell1: In the S3 bucket we have 1.6.8 version of containerd

runtime/service_integ_test.go
vvejell1: Pipeline works fine after removing Cgroupversion.

tools/docker/Dockerfile.integ-test
swagatbora90: Can you add this to the comment on why we have 1.6.8 and a link if possible.

vvejell1: Nothing specific. While adding ARM instance to the Firecracker ContainerCI, I have added the updated version 1.6.8 containerd and ctr binaries accordingly.

swagatbora90 (dismissed):

tools/docker/Dockerfile.integ-test
swagatbora90: On second thought, since we are only installing `/tmp/bin/containerd-shim-runc-v2`, I think it should be fine. Though we should look into upgrading CI to the same containerd version that fc-cd builds with.  Approving.

swagatbora90 (approved):

Issue #472, opened 2021-02-09
Error with quickstart and suggestion on docker dependency


alexellis:
There are two things this issue is about:

1) There is an error with the Quickstart that I cannot get past yet, when using Ubuntu
2) The dependency on Docker is confusing since Docker and containerd ship very different (incompatible) versions of containerd

I was chatting to @estesp about trying to get [faasd](https://github.com/openfaas/faasd) (based upon containerd) to work with firecracker, however I was surprised to see [that Docker is required to do a build of the containerd binary](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/docs/quickstart.md).

Quite often Docker and containerd ship with different binaries that aren't compatible, so is the idea that I should build on my workstation and copy over to a server which has a stand-alone version of the tools on it?

I'd much prefer being able to build then deploy to the same Linux host where I'll be testing out firecracker. This indirection is causing some friction and confusion for me.

Hope this is useful feedback on some level.

alexellis:
This is the last command that worked in the quickstart.

```
root@firecracker:/var/lib/firecracker-containerd/snapshotter/devmapper# if ! $(sudo dmsetup reload "${POOL}" --table "${THINP_TABLE}"); then
>     sudo dmsetup create "${POOL}" --table "${THINP_TABLE}"
> fi

device-mapper: reload ioctl on fc-dev-thinpool  failed: No such device or address
Command failed.
```

System details (on Packet/Equinix Metal)

```
root@firecracker:~# uname -a
Linux firecracker 5.8.0-26-generic #27-Ubuntu SMP Wed Oct 21 22:29:16 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
root@firecracker:~# cat /etc/os-release 
NAME="Ubuntu"
VERSION="20.10 (Groovy Gorilla)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.10"
VERSION_ID="20.10"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=groovy
UBUNTU_CODENAME=groovy
root@firecracker:~# 
```

alexellis:
[...]

PR #728, opened 2023-01-13
Make agent static build the default for host build as well

sgaist:
*Issue #472*

*Description of changes:*
Change the default agent build type when done on host to static as is done when using Docker. This will ensure that people have a working version of the agent when building the microVM image.

The build can still be turned back to dynamic by calling `make STATIC_AGENT=`.


vistaarjuneja:
thank you, I was facing the same error and this helped me fix it!

PR #744, opened 2023-04-27
regenerate all protobufs

henry118:
Regenerate all protobufs to address the random CI failures.

Normally our CI should skip protobuf generation because the target already exist. However in our recent upgrade to Bullseye we see random trigger of the workflow.

Manually deleting the existing protobuf files and regenerate locally would also result in diffs against the existing versions. So updating the files in the repo seems like a valid solution.


swagatbora90:
>Normally our CI should skip protobuf generation because the target already exist. However in our recent upgrade to Bullseye we see random trigger of the workflow.

Can you explain what you mean by target already exist. We explicitly run `make proto` in our [github actions](https://github.com/firecracker-microvm/firecracker-containerd/blob/main/.github/workflows/build.yaml#L53-L54). So why would CI skip this step?

Also, wondering if any protobuf gen packages upgrade recently that caused the proto definitions to change?

swagatbora90 (approved):
Changes looks good to me. But I have some questions related to the trigger of this failure and randomness.

henry118:
When I say skip the protobuf generation, I am referring to GNU Make, which will skip executing the build rules if it detects the target outputs already exist (and not outdated). In this case the outputs are the `*.pb.go` files, which already exist in the GH repo. Make only execute the rules in the following two cases:
1. The output target do not exist; or
2. The timestamp on the targets are older than the dependencies (`*.proto`) in our case.

It's not clear which of the two cases were causing the build rule execution in our CI, or if it's a bug in GNU Make. But the protobuf files were generated long time ago and regenerating the files always result in different versions. Most likely due to the fact that we always pull the head of the `protoc-gen-gogo` during the build.

fangn2:
LGTM, thanks for fixing it!

PR #752, opened 2023-05-14
Use fifo close on task exec cleanup to prevent directory leak

austinvazquez:
*Issue #, if available:*
N/A

*Description of changes:*
Close fifoSet for the task and exec when the task is deleted.

### DIFF
--- a/agent/service.go
+++ b/agent/service.go
@@ -253,6 +253,9 @@ func (ts *TaskService) Create(requestCtx context.Context, req *taskAPI.CreateTas
 			logger.WithError(err).Error()
 			return nil, err
 		}
+		ts.addCleanup(taskExecID, func() error {
+			return fifoSet.Close()
+		})
 
 		var stdinConnectorPair *vm.IOConnectorPair
 		if req.Stdin != "" {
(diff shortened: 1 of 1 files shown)


djdongjin (approved):

austinvazquez:
@djdongjin updated commit message and PR title to be more specific to the leak. Please re-review.

djdongjin (approved):

agent/service.go
```diff
@@ -253,6 +253,9 @@ func (ts *TaskService) Create(requestCtx context.Context, req *taskAPI.CreateTas
 			logger.WithError(err).Error()
 			return nil, err
 		}
+		ts.addCleanup(taskExecID, func() error {
+			return fifoSet.Close()
```
swagatbora90, line 257: These are fifos inside the VM, right? Once the task gets deleted, the VM should be shut down, so how is this contributing to the leak?

austinvazquez: > Once the task gets deleted, the VM should be shut down

The VM _can_ shutdown when all tasks are stopped, but that is not always the case. Default is for VM to stay up even if all tasks are deleted. See https://github.com/firecracker-microvm/firecracker-containerd/blob/main/proto/firecracker.proto#L34

However, you are correct `Create` cleanup removes the top level directory here and wouldn't contribute to the leak. Making this not _required_ but doesn't hurt anything to have either. Would you prefer to exclude this?

swagatbora90: Okay, just wanted to understand if this was contributing to the leak as well besides the exec path. I am fine keeping this in as an extra safety measure.

swagatbora90 (approved):
LGTM

PR #767, opened 2023-09-21
Send to copyDone when err == nil

Kern--:
*Issue #, if available:*

*Description of changes:*
The ioproxy helps copy container logs from the normal container fifos to the firecracker uvm's vsock. The copying happens in a go-routine and when complete, it send a notification over the `copyDone` channel to signal to the agent/runtime to cleanup the ioproxy.

This change fixes a bug where, if the iopoxy copy finished without an error (i.e. `io.CopyBuffer` got an `EOF` from the read end), the `copyDone` channel was not notified leaving the ioproxy connected to the vsock, but not the container. This change always notifies `copyDone` when the copy finishes.

### DIFF
--- a/internal/vm/ioproxy.go
+++ b/internal/vm/ioproxy.go
@@ -152,8 +152,8 @@ func (connectorPair *IOConnectorPair) proxy(
 			} else {
 				logger.WithError(err).Error("error copying io")
 			}
-			copyDone <- err
 		}
+		copyDone <- err
 		defer logClose(logger, reader, writer)
 	}()
 


austinvazquez (approved):

swagatbora90 (approved):

internal/vm/ioproxy.go
```diff
@@ -152,8 +152,8 @@ func (connectorPair *IOConnectorPair) proxy(
 			} else {
 				logger.WithError(err).Error("error copying io")
 			}
-			copyDone <- err
 		}
+		copyDone <- err
```
swagatbora90, line 156: qq: we close the `copyDone` channel regardless of whether the copy succeeds or fails. Shouldn't that indicate the runtime to clean up the ioproxy?

Kern--: Ah, yes. I missed that. Using the debug logs from https://github.com/firecracker-microvm/firecracker-containerd/pull/766 we actually don't see the "copied %d" message from https://github.com/firecracker-microvm/firecracker-containerd/blob/fc5e81ef584b54e91f7d7f8f6a1432dead7fa060/internal/vm/ioproxy.go#L147 which seems like we're actually still stuck in the `io.CopyBuffer`?

swagatbora90: Do we see the `copied` message with your change to explicitly send an error?

[this](https://github.com/firecracker-microvm/firecracker-containerd/blob/main/internal/vm/ioproxy.go#L209-L211) is where we wait for an error in the copy channel. I assumed this should return nil once we close the channel.

PR #782, opened 2024-02-17
Add unit test to not log stdout or stderr fields on create task request

austinvazquez:
*Issue #, if available:*
Validates #781 

*Description of changes:*
This change adds a unit test to validate when launching a container with the firecracker runtime the shim implementation will not log the stdout/stderr shim logger binary URI to disk.

### DIFF
--- a/runtime/service_test.go
+++ b/runtime/service_test.go
@@ -18,11 +18,15 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
+	"github.com/containerd/containerd/namespaces"
+	"github.com/containerd/containerd/runtime/v2/task"
 	"github.com/firecracker-microvm/firecracker-go-sdk"
 	models "github.com/firecracker-microvm/firecracker-go-sdk/client/models"
 	"github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus/hooks/test"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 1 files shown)


runtime/service_test.go
```diff
@@ -323,3 +327,42 @@ func TestDebugConfig(t *testing.T) {
...
 }
+
+// TestTaskCreateDoesNotLogStdoutOrStderrFields validates the firecracker-containerd runtime service implementation
+// does not log the Stdout/Stderr request fields because they may contain sensitive information for the shim logger binary use case.
+func TestTaskCreateDoesNotLogStdoutOrStderrFields(t *testing.T) {
+	testLogger, hook := test.NewNullLogger()
+	testLogger.Level = logrus.DebugLevel
+
+	vmIsReady := make(chan struct{})
+	close(vmIsReady)
```
sondavidb, line 338: `defer close(vmisReady)`?

sondavidb (commented):
Is there a mechanism for not redacting the info? (e.g. if a debug flag is set) If so, we might want to make sure that it gets tested as well.

runtime/service_test.go
austinvazquez: In this case, the channel is closed after the `CreateVM` API call is made. So mocking here the VM is ready.

austinvazquez:
> Is there a mechanism for not redacting the info? (e.g. if a debug flag is set) If so, we might want to make sure that it gets tested as well.

Not that I have found, but open to ideas there.

runtime/service_test.go
```diff
@@ -323,3 +327,42 @@ func TestDebugConfig(t *testing.T) {
...
+	createTaskRequest := &task.CreateTaskRequest{
+		ID:     t.Name(),
+		Stdout: "/sbin/shim-loggers-for-containerd --env USERNAME=admin --env PASSWORD=admin",
+		Stderr: "/sbin/shim-loggers-for-containerd --env TOKEN=tolkien",
+	}
+
+	// (*service).Create will fail on (Dir).CreateBundleLink after the log we want to validate.
+	_, _ = uut.Create(ctx, createTaskRequest)
+
+	for _, entry := range hook.AllEntries() {
```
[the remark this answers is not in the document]
austinvazquez: Good call.

swagatbora90 (approved):

sondavidb (approved):

PR #785, opened 2024-04-26
Upgrade containerd to v1.7

henry118:
Upgrade containerd to v1.7


agent/ioproxy_handler.go
```diff
@@ -16,11 +16,11 @@ package main
 import (
 	"context"
 
+	task "github.com/containerd/containerd/api/runtime/task/v2"
 	"github.com/containerd/containerd/log"
-	"github.com/containerd/containerd/runtime/v2/task"
 	"github.com/firecracker-microvm/firecracker-containerd/internal/vm"
 	ioproxy "github.com/firecracker-microvm/firecracker-containerd/proto/service/ioproxy/ttrpc"
-	"github.com/gogo/protobuf/types"
+	"google.golang.org/protobuf/types/known/emptypb"
```
swagatbora90, line 23: containerd has defined a [types](https://github.com/containerd/containerd/blob/main/protobuf/types/types.go) package to make this migration easier. This should also help with any future migration.

henry118: Updated. @swagatbora90 PTAL

.buildkite/pipeline.yml
```diff
@@ -78,21 +78,21 @@ steps:
...
-      DOCKER_IMAGE_TAG: "$BUILDKITE_BUILD_NUMBER"
-      NUMBER_OF_VMS: 10
-      EXTRAGOARGS: "-v -count=1 -race -timeout 3m"
-      FICD_DM_VOLUME_GROUP: fcci-vg
-    artifact_paths:
-      - "snapshotter/logs/*"
-    command:
-      - make -C snapshotter integ-test FICD_DM_POOL=build_${BUILDKITE_BUILD_NUMBER}_snapshotter
-    timeout_in_minutes: 10
+  # - label: ":running: snapshotter isolated tests"
```
swagatbora90, line 81: Do we intend to enable these tests in the future? Please cut a issue so that we can track it once this PR is merged.

Makefile
```diff
@@ -273,31 +273,31 @@ $(CNI_BIN_ROOT):
 
 BRIDGE_BIN?=$(BINPATH)/bridge
 $(BRIDGE_BIN):
-	GOBIN=$(dir $@) GO111MODULE=off go get -u github.com/containernetworking/plugins/plugins/main/bridge
+	GOBIN=$(dir $@) go install github.com/containernetworking/plugins/plugins/main/bridge@v1.1.0
```
swagatbora90, line 276: Why do we need to pin the version here. Can we not use `@latest`? This will need separate update mechanism otherwise

.buildkite/pipeline.yml
henry118: Probably not. The snapshotter is no longer needed. But i just commented this block in case we'd like to revoke this decision later.

Makefile
henry118: I had to pin the plugins to some older versions to pass integ-tests, which still have the CNI configured for older versions

swagatbora90 (approved):
LGTM.

Makefile
swagatbora90: Besides CNI, there are some other packages such as `git-validation` etc that have their versions pinned.

henry118: Yeah. For toolchain binaries, it's better to pin to a known good version otherwise upgrade may (surprisingly) break our build.

PR #789, opened 2024-05-22
regenerate protobuf with the right protoc compiler version

henry118:
*Issue #, if available:*

Fix proto issue

*Description of changes:*
regenerate protobuf with the right protoc compiler version


swagatbora90 (commented):
Why did the protoc version change? Do we need to pin the version in the [Dockerfile](https://github.com/firecracker-microvm/firecracker-containerd/blob/main/tools/docker/Dockerfile.proto-builder) to prevent future regression?

henry118:
> Why did the protoc version change? Do we need to pin the version in the [Dockerfile](https://github.com/firecracker-microvm/firecracker-containerd/blob/main/tools/docker/Dockerfile.proto-builder) to prevent future regression?

Good point. I pinned the compiler versions to prevent future failures

swagatbora90 (approved):

austinvazquez (approved):

PR #795, opened 2024-09-25
add logs to debug flaky test

swagatbora90:
*Issue #, if available:*
There are some tests that use `startAndWaitTask` call to wait for the container exit and also capture the logs. Sometimes, containers will exit successfully but will return empty stdout. Adding logs to see if the containers logged anything to the stderr instead. 
*Description of changes:*


pendo324 (dismissed):

henry118 (commented):
The PR desc said adding logs, but patch only contains sleeps. Can you either update the commit message or perhaps adding logs as mentioned?

Issue #819, opened 2025-04-12
docker-credential-mmds Glibc Version Mismatch Issue (Static Linking)


andre-j3sus:
I was experimenting with remote snapshotters, and I got the following error when setting the credentials:

```
/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by docker-credential-mmds)
```

This error occurred because the binary was being linked against the version of glibc present on the host system (Ubuntu 24), but the environment where the binary was expected to run (Debian-based microVM image) had a different version of glibc, leading to compatibility issues.

The `docker-credential-mmds` binary should be statically linked  (similar to the agent's build process), ensuring compatibility across different systems regardless of the host’s glibc version.

---

### Suggested Solution

- Add a flag to the `docker-credential-mmds` build process (similar to the agent's build process) to **enable static linking** ( set `CGO_ENABLED=0`) and avoid dependency on the host's glibc version.

I'll open a PR with this fix.

sondavidb:
Closed by #820, thanks for the contribution! Feel free to reopen if there's any further issues with this.

PR #820, opened 2025-04-12
#819: fix: enable static linking for docker-credential-mmds to resolve glib

andre-j3sus:
*Issue #819, if available:*

*Description of changes:*

Add a `STATIC_DOCKER_CREDENTIAL_MMDS` flag to the `docker-credential-mmds` build process (similar to the agent's build process) to enable static linking ( set `CGO_ENABLED=0`) and avoid dependency on the host's glibc version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ INTEG_TEST_SUBDIRS:=$(addprefix integ-test-,$(SUBDIRS))
 
 export INSTALLROOT?=/usr/local
 export STATIC_AGENT
+export STATIC_DOCKER_CREDENTIAL_MMDS
 
 export DOCKER_IMAGE_TAG?=latest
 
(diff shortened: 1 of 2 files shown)


sondavidb (approved):
CI passes, Buildkite link: https://buildkite.com/firecracker-microvm/firecracker-containerd/builds/4384

Cancelled jobs are because Ubuntu 20.04 runners are now obsolete, will add a PR separately to remove them.

LGTM, I'll get one more reviewer on our end to look at it. Thanks!

sondavidb:
Worth calling out, this is logic we use in our agent as well ([code](https://github.com/firecracker-microvm/firecracker-containerd/blob/439db8d5bf7b5431d482a6f3f9b8080a9cb2b5c5/agent/Makefile#L24-L29), so it should be fine to reuse here IMO

Kern-- (approved):

PR #841, opened 2025-11-20
chore: bump dependencies

pendo324:
*Issue #, if available:*

*Description of changes:*
- bump selinux, crypto, tc-redirect-tap deps (closes #839, #840)
- minor potential CI reliability improvement


runtime/service_integ_test.go
```diff
@@ -2469,6 +2469,8 @@ func TestPauseResume_Isolated(t *testing.T) {
 		_, err := fcClient.ResumeVM(ctx, &proto.ResumeVMRequest{VMID: vmID})
 		require.NoError(t, err)
 
+		time.Sleep(3 * time.Second)
```
sondavidb: leftover artifact?

sondavidb: Makes sense, for commit history reasons this feels weird to include in this PR since we do squash and merge but I guess if it's just one line it's not that bad 🤷

PR #843, opened 2025-12-10
fix: add HTTP server timeouts for local network test helper (G114)

### LINKED ISSUE #807: Resolve G114 use of net/http serve function that has no support for setting timeouts
Resolve the following errors:
```
Error: internal/network_test_utils.go:135:10: G114: Use of net/http serve function that has no support for setting timeouts (gosec)
		return http.ListenAndServe(l.ipAddr+":80", nil)
```

arianbeh:
Fixes #807.

This patch replaces a raw `http.ListenAndServe()` call in
`internal/network_test_utils.go` with an explicit `http.Server`
configured with ReadTimeout, WriteTimeout, and IdleTimeout to satisfy
gosec rule G114 ("Use of net/http serve function without timeouts").

This code is used only in test helpers and does not affect runtime
behavior.

Validated locally with:
- make lint
- go test ./internal/...

Signed-off-by: Aryan Behmardi

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -25,7 +25,6 @@ issues:
   exclude:
     - G104 # Errors unhandled
     - G103 # Use of unsafe calls should be audited
-    # - G114 # Use of net/http serve function without timeouts.
     - G115 # Integer overflow conversion should be audited
     - G204 # Subprocess launched with variable
     - G301 # Expect directory permissions to be 0750 or less


.golangci.yml
```diff
@@ -25,7 +25,7 @@ issues:
   exclude:
     - G104 # Errors unhandled
     - G103 # Use of unsafe calls should be audited
-    - G114 # Use of net/http serve function without timeouts.
+    # - G114 # Use of net/http serve function without timeouts.
```
sondavidb: Can just remove it, no need to comment out.

